PR #28, opened 2023-11-16
chore: auto sync addons cr  to block-index

JashBook:


hack/gen-addon-crs.py
```diff
@@ -35,13 +35,19 @@
     with open(chart_file, 'r') as f:
         print("Generating addon CR for " + name)
         chart = yaml.safe_load(f)
+        if 'annotations' not in chart:
+            print("Skip generating addon CR for " + name)
+            continue
```
ldming, line 40: Need this?

JashBook: If the key does not exist, an error will occur.
Skip can easily scan the entire directory.

ldming (approved):

Issue #55, opened 2023-11-30
[BUG]mysql cannot be started in some Storage Classes due to the presence of the "lost+found" directory


wuxxfree:
mysql cannot be started in some Storage Classes due to the presence of the "lost+found" directory.
![image](https://github.com/apecloud/kubeblocks/assets/92258528/422b5a52-2d68-4b40-b575-a84545c71c35)
![6aabb7bed7ba3861a7996bfc2b64735](https://github.com/apecloud/kubeblocks/assets/92258528/d8ece68f-fe56-4990-8877-7b1280f7c112)
version:
[root@k8s-cpu ~]# kbcli version
Kubernetes: v1.24.6
KubeBlocks: 0.7.0
kbcli: 0.7.0

wuxxfree:
[https://github.com/apecloud/kubeblocks-addons/pull/56](url)

PR #56, opened 2023-11-30
fix: mysql cannot be started in some Storage Classes 

wuxxfree:
fix: mysql cannot be started in some Storage Classes due to the presence of the "lost+found" directory


wuxxfree:
https://github.com/apecloud/kubeblocks-addons/issues/55

wangyelei (approved):

PR #61, opened 2023-12-06
feature: support kafka component definition and opsDefinition

wangyelei:


addons/kafka/scripts/kafka-exporter-setup.tpl
```diff
@@ -0,0 +1,34 @@
...
+    {{- if contains "kafka-broker" $e.componentDef }}
+    {{- $component = $e }}
+    {{- end }}
+  {{- end }}
+{{- end }}
+{{- /* build --kafka.server= string */}}
+{{- $replicas := $component.replicas | int }}
+{{- $servers := "" }}
+{{- range $i, $e := until $replicas }}
+  {{- $podFQDN := printf "%s-%s-%d.%s-%s-headless.%s.svc.%s" $clusterName $component.name $i $clusterName $component.name $namespace $.clusterDomain }}
```
Y-Rookie, line 23: If it is implemented based on KubeBlocks version 0.8, for such references, try to use Vars on the ComponentDefinition as much as possible. There are some built-in ENVs and other types of references available. Try to avoid referencing via the .component method.

addons/kafka/templates/componentdef-combine.yaml
```diff
@@ -0,0 +1,203 @@
...
+  provider: kubeblocks
+  description: |-
+    Kafka servers that act as both brokers and controllers are referred to as "combined" servers. Combined servers
+    are simpler to operate for small use cases like a development environment. Combined mode is not recommended in critical
+    deployment environments.
+  serviceKind: kafka
+  serviceVersion: {{ default .Chart.AppVersion .Values.clusterVersionOverride }}
+  services:
+  - name: broker
+    serviceName: broker
```
Y-Rookie, line 23: combine server’s service name also called "broker"?

wangyelei: yes, the combine contains the broker and controller process

addons/kafka/scripts/kafka-exporter-setup.tpl
wangyelei: here will ref the broker component headlessService and the broker Replicas to build a param, no found these ENV can be used. only can ref the broker headless service is not enough

Y-Rookie (approved):

ldming (approved):

PR #66, opened 2023-12-07
fix: update oceanbase cluster

### LINKED ISSUE #67: [Features]Oceanbase creates primary/secondary tenant on creating

shanshanying:
- fix #67 
support creating primary/standby tenants on creating Oceanbase Cluster, by
- specify 'obClusters=2' (will create two ob clusters, one for primary, one for standby)
- specify 'tenant.name'.

```bash
helm install oceanbase addons/oceanbase
```
To create a primary-secondy OB Cluster (which is actually a primary cluster, AND a secondary cluster)
we set `obCluster=2`.
e.g.  create a primary-secondy OB Cluster, each with 1 replica, and the tenant name is `tenant1. we use:
```bash
helm install mycluster addons/oceanbase-cluster --set obClusters=2 --set zoneCount=1 --set replicas=1  --set tenant.name='tenat1'
```


shanshanying:
Will wait for one more day.
Have invited member from OceanBase Operator team for review.

PowerfooI:
Removing `statefulSpec.minReadySeconds` may cause failed cluster restart, which should be concerned.

Other parts LGTM.

shanshanying:
> Removing `statefulSpec.minReadySeconds` may cause failed cluster restart, which should be concerned.
> 
> Other parts LGTM.

`statefulSpec.minReadySeconds` is no longer available in kubeblocks 0.8. Will checkout why and hwo to add the api if necessary.

PR #70, opened 2023-12-08
chore: support redis sentinel

1aal:
- to support `kbcli cluster create redis` with sentinel


Y-Rookie:
n version 0.8, there is a new approach to implement NodePort, so it is recommended to abandon this Pull Request (PR).

PR #130, opened 2023-12-22
support: new api redis cluster chart

1aal:
- kbcli will adapt the `cluster create` command to the new API using this cluster chart as an example.


addons/0.8-redis-cluster/values.yaml
```diff
@@ -0,0 +1,49 @@
...
+## @param sentinel.cpu
+## @param sentinel.memory
+## @param sentinel.storage
+##
+
+nodePort:
+  enable: true
+
+sentinel:
+  enable: true
```
ldming, line 45: enable to  enabled

Y-Rookie (approved):

ldming (approved):

PR #193, opened 2024-01-08
feat: support dynamic config for wescale

earayu:
1. upgrade wescale image from 0.2.0 to 0.2.2
2. make some config params to be dynamic for wescale


heng4fun (approved):
LGTM

PR #196, opened 2024-01-08
fix: weaviate node discovery failed

iziang:


addons/weaviate/Chart.yaml
ldming, line 1: better to update chart version, maybe 0.8.0?

iziang: fixed

ldming (approved):

PR #208, opened 2024-01-10
chore: Use 'unixSignalTrigger' to support parameter hot reload for wescale

earayu:
1. set `reloadOptions=unixSignalTrigger` for vttablet and vtgate
2. make pool size parameters in vttablet to be dynamic


heng4fun (approved):
LGTM

PR #220, opened 2024-01-11
chore: add svc for ob-repl to select pod by role

shanshanying:


wangyelei (approved):

addons/oceanbase-cluster/templates/cluster.yaml
```diff
@@ -11,6 +11,18 @@ spec:
...
+  services:
+    - name: rw-svc
+      spec:
+        ports:
+          - name: sql
+            port: 2881
+            targetPort: sql
+            protocol: TCP
+        selector:
+          kubeblocks.io/role: primary
```
Y-Rookie, line 24: It would be better to use the outer RoleSelector. kubeblocks.io/role: is a key for internal implementation.

PR #237, opened 2024-01-16
chore: remove metrics component in ES cluster

kubeJocker:


addons/elasticsearch-cluster/templates/cluster.yaml
```diff
@@ -39,16 +39,3 @@ spec:
               requests:
                 storage: {{ .Values.persistence.data.size }}
       {{- end }}
-
-    - name: metrics # user-defined
-      componentDefRef: metrics # ref clusterdefinition componentDefs.name
-      replicas: {{ .Values.metrics.replicaCount | default 1 }}
```
leon-inf, line 45: the values of metrics should also be cleared

kubeJocker: done

leon-inf (approved):

PR #275, opened 2024-01-29
feat: support kafka nodeport in componentDefinition apis

caiq1nyu:
1. Use ComponentDefinition Apis to build Kafka add-on resources
2. Kafka services support NodePort mode
3. The broker addresses are exposed through the service, and they can be provided with the expose capability later to support public/vpc network access requirements.

### DIFF
--- a/addons/kafka-cluster/templates/cluster.yaml
+++ b/addons/kafka-cluster/templates/cluster.yaml
@@ -4,7 +4,8 @@ metadata:
   name: {{ include "kblib.clusterName" . }}
   labels: {{ include "kblib.clusterLabels" . | nindent 4 }}
   annotations:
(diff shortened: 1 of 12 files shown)


wangyelei (approved):

addons/kafka-cluster/templates/cluster.yaml
```diff
@@ -46,6 +48,7 @@ spec:
     {{- else }}
     - name: broker
       componentDefRef: kafka-broker
```
Y-Rookie, line 50: componentDefRef can be removed

Y-Rookie (approved):

addons/kafka-cluster/templates/cluster.yaml
caiq1nyu: considering the compatibility issues with the ClusterDefinition API in some special cases. this logic will be temporarily retained in one or two versions and will be removed in subsequent iterations.

PR #292, opened 2024-02-02
chore: update ob resource name

shanshanying:


wangyelei (approved):

addons/oceanbase-cluster/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-name: oceanbase-cluster
+name: ob-ce-cluster
```
JashBook, line 2: change the addon oceanbase-cluster dir name too

addons/oceanbase/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-name: oceanbase
+name: ob-ce
```
JashBook, line 2: change the addon oceanbase-cluster dir name too 
and update kubeblcoks addon name

PR #324, opened 2024-02-22
fix: set proxy default auth

gerayking:
https://github.com/apecloud/kubeblocks/issues/6661


earayu:
@gerayking I think vttablet also need to use `mysqlbased`

earayu (approved):

earayu (approved):

heng4fun (approved):
LGTM

PR #358, opened 2024-03-07
feat: pulsar support nodeport

caiq1nyu:
Pulsar Broker & Proxy component support NodePort service with KB0.8 Component api.

Because the KB component api does not support version management so far, this PR uses Component api for version 3.0.2, while 2.11.2 still uses the ClusterDefinition api.

users can enable the nodeport function by setting enableNodePort=true in the pulsar-cluster helm values.


addons/pulsar/templates/componentdef-zookeeper.yaml
```diff
@@ -0,0 +1,138 @@
...
+metadata:
+  name: pulsar-zookeeper
+  labels:
+      {{- include "pulsar.labels" . | nindent 4 }}
+spec:
+  provider: kubeblocks
+  description: Pulsar zookeeper.
+  serviceKind: pulsar
+  serviceVersion: {{ default .Chart.AppVersion .Values.clusterVersionOverride }}
+  updateStrategy: BestEffortParallel
```
Y-Rookie, line 12: Serial would be better

Y-Rookie (approved):

addons/pulsar/templates/componentdef-zookeeper.yaml
caiq1nyu: for components without kb role detection logic, RSM currently has bugs under the SerialUpdateStrategy strategy(issue: https://github.com/apecloud/kubeblocks/issues/6785). so try the BestEffortParallel strategy to ensure normal function.

wangyelei (approved):

PR #366, opened 2024-03-11
fix: polarx role probe

xuriwuyun:


ahjing99:
should fix https://github.com/apecloud/kubeblocks/issues/6786

shanshanying (approved):

PR #396, opened 2024-03-20
chore: set skip slave/replica start if exist signal file

ldming:


addons/mysql/templates/actionset-volumesnapshot.yaml
```diff
@@ -24,5 +24,12 @@ spec:
            if [ -f ${DATA_DIR}/relay-log.info ]; then
               rm -rf ${DATA_DIR}/relay-log.info
            fi
+           touch ${DATA_DIR}/.xtrabackup_restore_new_cluster
```
wangyelei, line 27: better to use `.restore_new_cluster`

ldming: done, rename to `.restore_new_cluster`

wangyelei (approved):

PR #414, opened 2024-04-01
chore: support redis cluster sharding scaling

Y-Rookie:


addons/redis/config/redis-cluster7-config.tpl
```diff
@@ -1,8 +1,88 @@
+bind 0.0.0.0
```
ldming, line 1: Use `* -::*` to compatible with IPv6

ldming (commented):
Better to change the chart version

addons/redis/config/redis-cluster7-config.tpl
Y-Rookie: > Use `* -::*` to compatible with IPv6

done

ldming (approved):

PR #425, opened 2024-04-07
temp: add pitr logic

yabinji:


addons/mogdb/templates/secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
nayutah, line 1: pls remove the default secret, it is not a safe way

PR #482, opened 2024-04-30
feat: tidb addon revamp

cjc7373:
- [x] use new KB api
  - [x] replace `$KB_REPLICA_COUNT`
- monitor
  - [x] metrics config in compdef
  - [x] grafana dashboards
  - ~~logs config~~ doesn't take effect for now
- [ ] account management?
- [x] backup/restore
  - only minio is supported, others are not tested
- [x] configuration management
- hscale
  - tidb is stateless
  - [x] pd
  - [x] tikv
    - [x] scale in may need memberLeave hook


free6om (approved):

addons/tidb/templates/clusterdefinition.yaml
```diff
@@ -5,138 +5,27 @@ metadata:
...
+      orders:
+        provision:
+          - tidb-pd
+          - tikv
+          - tidb
+        terminate:
+          - tidb
+          - tikv
+          - tidb-pd
+        # FIXME: due to kb bugs, if update order is specified, update will not succeed
```
Y-Rookie, line 27: What is the bug? Are there any corresponding issues?

wangyelei (approved):

addons/tidb/scripts/tikv_member_leave.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -exo pipefail
+
+DOMAIN=$KB_NAMESPACE".svc.cluster.local"
+PD_ADDRESS="http://${KB_CLUSTER_NAME}-tidb-pd.${DOMAIN}:2379"
```
Y-Rookie, line 6: It's better not to hardcode `-tidb-pd`

addons/tidb/scripts/tikv_member_leave.sh
```diff
@@ -0,0 +1,19 @@
...
+echo "$PD_ADDRESS"
+TIKV_ADDRESS="${KB_LEAVE_MEMBER_POD_NAME}.${KB_CLUSTER_NAME}-tikv-headless.${KB_NAMESPACE}.svc:20160"
```
Y-Rookie, line 8: ditto

addons/tidb/templates/componentversion.yaml
```diff
@@ -0,0 +1,56 @@
...
+metadata:
+  name: tidb-pd
+  labels:
+    {{- include "tidb.labels" . | nindent 4 }}
+spec:
+  releases:
+    - name: "7.1.5"
+      serviceVersion: "7.1.5"
+      images:
+        pd: {{ .Values.image.registry | default "docker.io" }}/{{ .Values.image.repository }}/pd:{{ default .Chart.AppVersion .Values.image.tag }}
```
Y-Rookie, line 12: The image tag should better be static

addons/tidb/templates/clusterdefinition.yaml
cjc7373: The observation is the first created component's (pd) annotation `kubeblocks.io/generation` does not equal to cluster's generation. So that component's status is stuck in `Creating`. Though I haven't found the root cause.

Issue #620, opened 2024-06-03
[Improvement] Enable Node Label Assignment in Helm Charts for Clusters


dingshun-cmss:
## Is your improvement request related to a problem? Please describe.
When using the Helm charts for xxxxxx-clusters in the official addon, it's not possible to assign a value to `cluster.spec.affinity.nodeLabels`. This limitation arises because the settings within `kblib.affinity` do not accommodate this configuration.  As a result, we cannot create `Clusters` on specific Kubernetes nodes witn specific labels

## If this is a new function, please describe the motivation and goals.
We want to create `Clusters` on specific Kubernetes nodes witn specific labels, using the official helm charts for clusters

## Describe the solution you'd like
Maybe we can modify the `kblib.affinity` to make it possible to set all the attributes in `cluster.spec.affinity`

## Describe alternatives you've considered
N/A

## Additional context
N/A

PR #621, opened 2024-06-03
feat: support setting affinity node labels

dingshun-cmss:
- Support setting affinity node labels in kblib.affinity

Fix: #620

### DIFF
--- a/addons/kblib/values.yaml
+++ b/addons/kblib/values.yaml
@@ -11,6 +11,10 @@ terminationPolicy: Delete
 ##
 availabilityPolicy: node
 
+## @param affinityNodeLabels
+##
+affinityNodeLabels: {}
+
 ## @param podAntiAffinity
 ##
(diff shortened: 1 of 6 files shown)


addons/kblib/values.yaml
```diff
@@ -11,6 +11,10 @@ terminationPolicy: Delete
 ##
 availabilityPolicy: node
 
+## @param affinityNodeLabels
+##
+affinityNodeLabels: {}
```
shanshanying, line 16: affinityNodeLabels is expected to be a map. could you pls give some more exmaples to show how to set the values.

ldming (approved):

addons/kblib/values.yaml
dingshun-cmss: Fixed in f8e85c0cb0314da89fc9739599eb412d00145aef

Issue #619, opened 2024-06-03
[Features]  Move wescale related sidecars out out ApeCloud MySQL Cmpd


shanshanying:
WeScale works for variants of MySQL, covering MySQL, MariaDB, ApeCloud-MySQL, etcd.
This issued must be completed before we move on to:
-  https://github.com/apecloud/kubeblocks-addons/issues/163

PR #632, opened 2024-06-05
feat #619 :apecloud mysql remove wescale

skyrise-l:
[#619](https://github.com/apecloud/kubeblocks-addons/issues/619)
This pull request separates the Wescale proxy component from MySQL, facilitating the subsequent complete independence of Wescale to serve multiple databases.

The default configuration of the apecloud-mysql addon cluster is in standalone mode, without enabling the proxy. The startup method is similar to the old version and does not include any wesql-scale content. You can start it as a regular MySQL database according to your configuration needs.

Here, we focus on the startup method of the apecloud-mysql addon with the wesql-scale proxy enabled.

1 Firstly, compared to the old version, the biggest difference in using the apecloud-mysql cluster is that you must start an external etcd service before starting wesql-scale. You can use the kubeblocks etcd addon to start an etcd cluster, or use other methods to start it.
Here, we use the kubeblocks etcd addon source code installation as an example (other methods such as kbcli are similar):
```
git clone git@github.com:apecloud/kubeblocks-addons.git
cd kubeblocks
helm install etcd addons/etcd
helm install etcd-cluster addons/etcd-cluster 
```
2 Confirm the status of the etcd service and check the etcd service address (you can also check the etcd pod address).

kubectl get service

image
3 You can obtain the endpoint by concatenating the service address, or you can use the IP of the service address as the access address.
Using the service address mentioned above, you can concatenate to get the etcd service endpoint:
**etcd-cluster-etcd.default.svc.cluster.local:2379**

4  Start the apecloud-mysql cluster (addon is enabled by default).
Please note that in the creation parameters of apecloud-mysql cluster, mode represents the startup mode of mysql, including standalone and raftGroup. Only when in raftGroup mode, the additional parameter proxyEnabled is true to enable wsql-scale proxy is recommended and meaningful. The following is an example command, make sure to set the mode and proxyEnabled to enable the wesql-scale proxy for MySQL, and set the etcd service address/POD address list.

`helm install mysql-cluster addons/apecloud-mysql-cluster --set mode=raftGroup,proxyEnabled=true,etcd.serviceReference.endpoint="etcd-cluster-etcd.default.svc.cluster.local:2379"`

5 Check the cluster status.
```
kubectl get cluster 
kubectl get pods
# Enter the etcd container/wesql-scale container to check the wesql-scale configuration. This can also be used to probe the availability of the etcd service
etcdctl --endpoints=http://etcd-cluster-etcd.default.svc.cluster.local:2379 get /vitess --prefix --keys-only
```


nayutah (approved):

shanshanying (approved):

shanshanying:
image
And I guess we shall remove the servicerRef in ac-mysql component ?

PR #810, opened 2024-07-15
fix: delete some dep api in kafka addon

skyrise-l:
https://github.com/apecloud/kbcli/issues/397
fix some dep api in kafka addon


addons/kafka/templates/componentdef-broker.yaml
```diff
@@ -14,7 +14,6 @@ spec:
   services:
     - name: broker
       serviceName: broker
-      generatePodOrdinalService: true
```
weicao, line 17: should be replaced by podService?

Issue #786, opened 2024-07-09
[Improvement]Add ShellScript Check to CICD


shanshanying:

dingshun-cmss:
We can use [ShellCheck](https://www.shellcheck.net/) as the shell script linter

PR #816, opened 2024-07-16
chore: add shell check workflow

JashBook:
resolve https://github.com/apecloud/kubeblocks-addons/issues/786
1. add shell check for push
2. only check increment files
3. add `make shellcheck` on all shell scripts locally

### DIFF
--- a/.github/workflows/shellcheck-push.yml
+++ b/.github/workflows/shellcheck-push.yml
@@ -2,8 +2,6 @@ name: Shell Check
 
 on:
   push:
-#    paths:
-#      - '**.sh'
     branches:
       - '*'
       - '*/*'
--- a/.github/workflows/shellcheck-push.yml
+++ b/.github/workflows/shellcheck-push.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Get base commit id
         id: get_base_commit_id
         env:
-          REF_NAME: ${{ github.base_ref }}
+          REF_NAME: ${{ github.ref_name }}
         run: |
           BASE_COMMITID=`bash .github/utils/utils.sh --type 1 \
               --branch-name "${{ env.REF_NAME }}" \
--- a/.github/workflows/shellcheck-push.yml
+++ b/.github/workflows/shellcheck-push.yml
@@ -1,0 +1,63 @@
+name: Shell Check
+
+on:
+  push:
+#    paths:
+#      - '**.sh'
+    branches:
+      - '*'
+      - '*/*'
+    tags-ignore:
+      - '*'
+
+env:
+  BASE_BRANCH: origin/main
+
+jobs:
+  shell-check:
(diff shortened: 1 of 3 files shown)


weicao (approved):

.github/workflows/shellcheck-push.yml
```diff
@@ -0,0 +1,62 @@
...
+          fetch-depth: 0
+
+      - name: Get base commit id
+        id: get_base_commit_id
+        env:
+          REF_NAME: ${{ github.ref_name }}
+        run: |
+          BASE_COMMITID=`bash .github/utils/utils.sh --type 1 \
+              --branch-name "${{ env.REF_NAME }}" \
+              --base-branch "${{ env.BASE_BRANCH }}"`
```
weicao, line 38: have a try,
```
$ ./utils.sh --type 1  -bn origin/release-0.9 -br origin/main
HEAD^
```

```
$ ./utils.sh --type 1  -bn origin/release-0.8 -br origin/main
HEAD^
```

weicao: Why does the script attempt to find the latest common commit between origin/main and the current branch (e.g., origin/release-0.8) to use as the base point for generating the diff?

JashBook: Because the checks of GitHub actions only judge the result of the last check, the purpose of getting the commitid here is to check all commits starting from the base commitid of the main checkout branch. After pr is merged into main, it is also squashed into one commitid.

JashBook: > origin/release-0.9

Because the commitid span of the release-* branch after checking out the base commitid of the main branch is too large, the default search range of main branch commitid (e.g. 100) and release-* branch commitid (e.g. 50) cannot match the commitid.We will not submit PRs from release-* branches to main.

weicao: I see.

There is one more question. Since BASE_BRANCH is set to origin/main, if a developper commit a pr that base on origin/release-0.8, could it work?

JashBook: > origin/release-0.8

The push workflow cannot automatically obtain the base branch, and it cannot obtain the PR information.
If the base branch is origin/release-0.8, the workflow will still get the commitid list from origin/main branch by default for comparison. If they do not match, only the latest commitid can be checked.
But there is a solution. In each release-* branch workflow,BASE_BRANCH can be changed to origin/release-*, and it will work normally.

shanshanying (approved):

.github/workflows/shellcheck-push.yml
JashBook: Yes, after cherry-picking to release-* branch, I will modify BASE_BRANCH

PR #820, opened 2024-07-16
Added support for internal etcd to apecloud-mysql

skyrise-l:
Added support for internal etcd to apecloud-mysql. Two new modes have been added, controlled by the localEtcd variable in the cluster to use internal/external etcd. Internal etcd is temporarily not configured for high availability. We will first see if this change is feasible.


addons/apecloud-mysql/values.yaml
```diff
@@ -108,3 +108,7 @@ wesqlscale:
 ## if you install multiple releases of this chart.
 ## If specified, the cluster definition will use it as name.
 resourceNamePrefix: ""
+
+etcd:
+  apeiVersion: "2"
```
shanshanying, line 113: what is this filed for

addons/apecloud-mysql/templates/cmpd-proxy.yaml
```diff
@@ -181,4 +206,73 @@ spec:
...
+      valueFrom:
+        serviceVarRef:
+          compDef: {{ include "apecloud-etcd.componentDefName" . }}
+          name: etcd
+          optional: true
+          port:
+            name: etcd-client
+            option: Optional
+---
+apiVersion: apps.kubeblocks.io/v1alpha1
```
shanshanying, line 234: add replicas llimits to this cmpd, it is a simplified version with only ONE replica

addons/apecloud-mysql/templates/cmpd-proxy.yaml
```diff
@@ -181,4 +206,73 @@ spec:
       valueFrom:
         serviceRefVarRef:
           name: etcd
-          endpoint: Required
\ No newline at end of file
+          endpoint: Required
+          optional: true
+    - name: ETCD_LOCAL_POD_LIST
```
shanshanying, line 211: what is the difference btw ETCD_LOCAL_POD_LIST and KB_POD_LIST

addons-cluster/apecloud-mysql-cluster/templates/_helpers.tpl
```diff
@@ -80,4 +88,25 @@ raftGroup mode: max(replicas, 3)
...
+- name: etcd
+  volumeClaimTemplates:
+    - name: data
+      spec:
+        storageClassName: {{ .Values.proxy.storageClassName | quote }}
+        accessModes:
+          - ReadWriteOnce
+        resources:
+          requests:
+            storage: 20Gi
```
shanshanying, line 103: better to set etcd storage size using an input value.

addons/apecloud-mysql/scripts/vtctld.sh
```diff
@@ -1,16 +1,25 @@
...
 echo "starting vtctl"
+
+if [ -n "$ETCD_LOCAL_POD_LIST" ]; then
+  IFS=',' read -ra ETCD_POD_ARRAY <<< "$ETCD_LOCAL_POD_LIST"
+  endpoints=""
+  for pod in "${ETCD_POD_ARRAY[@]}"; do
+    endpoints+="${pod}.${ETCD_LOCAL_HEADLESS}.${KB_NAMESPACE}.svc.cluster.local:${ETCD_LOCAL_PORT},"
+  done
+  endpoints="${endpoints%,}"
+fi
```
shanshanying, line 11: what if the condition does not hold? endpoints values is?

shanshanying (commented):
the service descriptor CR will be created only when we use external ETCD. please add a check.

addons/apecloud-mysql/templates/cmpd-proxy.yaml
```diff
@@ -181,4 +206,73 @@ spec:
...
+          name: etcd
+          optional: true
+          port:
+            name: etcd-client
+            option: Optional
+---
+apiVersion: apps.kubeblocks.io/v1alpha1
+kind: ComponentDefinition
+metadata:
+  name: {{ include "apecloud-etcd.componentDefName" . }}
```
weicao, line 237: shall we import the etcd cmpd inside etcd-addon?
if can't, why?

addons/apecloud-mysql/values.yaml
skyrise-l: Control the API version of the local etcd, supporting either API v2 or API v3.

addons-cluster/apecloud-mysql-cluster/templates/_helpers.tpl
```diff
@@ -62,10 +62,18 @@ raftGroup mode: max(replicas, 3)
...
+    {{- else }}
+      {{- "apecloud-mysql-proxy-etcd" }}
+    {{- end }}
   {{- else }}
-    {{- "apecloud-mysql-with-proxy" }}
+    {{- if .Values.auditLogEnabled }}
+      {{- "apecloud-mysql-audit-with-proxy" }}
+    {{- else }}
+      {{- "apecloud-mysql-with-proxy" }}
+    {{- end }}
```
weicao, line 76: It seems that the composition of options has resulted in an excessive number of cmpd

addons-cluster/apecloud-mysql-cluster/templates/cluster.yaml
```diff
@@ -16,15 +16,20 @@ spec:
...
         - error
       {{- include "kblib.componentResources" . | indent 6 }}
       {{- include "kblib.componentStorages" . | indent 6 }}
       {{- include "kblib.componentServices" . | indent 6 }}
-    {{- if and (eq .Values.mode "raftGroup") .Values.proxyEnabled }}
-    {{- include "apecloud-mysql-cluster.proxyComponents" . | indent 4 }}
-    {{- end }}
+      {{- if and (eq .Values.mode "raftGroup") .Values.proxyEnabled }}
+      {{- include "apecloud-mysql-cluster.proxyComponents" . | indent 4 }}
+        {{- if .Values.localEtcd }}
```
weicao, line 32: localEtcd -> enableLocalEtcd

addons-cluster/apecloud-mysql-cluster/templates/etcd-service-descriptor.yaml
```diff
@@ -7,4 +8,9 @@ spec:
   serviceKind: etcd
   serviceVersion: {{ .Values.etcd.serviceVersion }}
   endpoint:
-    value: {{ .Values.etcd.serviceReference.endpoint }}
\ No newline at end of file
+    {{- if .Values.localEtcd }}
+    value: {{ include "kblib.clusterName" . }}-etcd-headless.default.svc.cluster.local:2379
+    {{- else }}
```
weicao, line 13: The ServiceDescriptor represents an external etcd. If it's an etcd component within the cluster, use vars.componentVars.

weicao: btw, here it seems you should not use headless svc.

addons-cluster/apecloud-mysql-cluster/values.yaml
```diff
@@ -56,5 +61,8 @@ sidecars:
 
 etcd:
   serviceReference:
-    endpoint: "etcd-cluster-etcd.default.svc.cluster.local:2379" # node1:2379,node2:2379,node3:2379
-  serviceVersion: 3.5.6
\ No newline at end of file
+    endpoint: "etcd-cluster-etcd-headless.default.svc.cluster.local:2379" # node1:2379,node2:2379,node3:2379
```
weicao, line 64: seems should not configure etcd endpoint address in values.yaml?

addons/apecloud-mysql/scripts/vtgate.sh
```diff
@@ -8,7 +8,13 @@ grpc_port=${VTGATE_GRPC_PORT:-'15991'}
 mysql_server_port=${VTGATE_MYSQL_PORT:-'15306'}
 mysql_server_socket_path="/tmp/mysql.sock"
 
-endpoints=${ETCD_SERVER:-'127.0.0.1:2379'}
+if [ -n "$SERVICE_ETCD_ENDPOINT" ]; then
```
weicao, line 11: better rename SERVICE_ETCD_ENDPOINT -> EXTERNAL_ETCD_SERVICE_ENDPOINT

addons-cluster/apecloud-mysql-cluster/templates/etcd-service-descriptor.yaml
skyrise-l: This part is saying that in order to standardize the use of serviceRef, I simply created a branch here. The value of the internal etcd should be fixed. 
Initially, I used Vars to reference componentVars. But it seemed a bit complicated.Should I change it back?

weicao: > This part is saying that in order to standardize the use of serviceRef, I simply created a branch here. The value of the internal etcd should be fixed. Initially, I used Vars to reference componentVars. But it seemed a bit complicated.Should I change it back?

Yes, always use ServiceDescriptor to configure the address of external services. I'm discussing with @leon-inf to add support for dynamically choosing and representing both external services and component services based on an environment variable.

addons/apecloud-mysql/scripts/etcd-post-start.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
-endpoints=${ETCD_SERVER:-'127.0.0.1:2379'}
+if [ -n "$SERVICE_ETCD_ENDPOINT" ]; then
+  endpoints="$SERVICE_ETCD_ENDPOINT"
+else
+  echo "SERVICE_ETCD_ENDPOINT is empty. Cannot proceed."
+  exit 1
+fi
+
 cell=${CELL:-'zone1'}
```
weicao, line 9: if multiple Clusters share the same etcd, $CELL will conflict.

weicao: if you want wescale shares etcd, its zone should be a value that contains cluster name.

skyrise-l: vtctl --topo_implementation etcd2 \
  --topo_global_server_address "${servers}" \
  --topo_global_root "/vitess/${KB_NAMESPACE}/${KB_CLUSTER_NAME}/global" VtctldCommand AddCellInfo \
  --root "/vitess/${KB_NAMESPACE}/${KB_CLUSTER_NAME}/$cell" \
  --server-address "${servers}" \
  $cell
set -e

As with the parameter settings of vtctl, all etcd-related configurations include the cluster name and namespace.

addons-cluster/apecloud-mysql-cluster/templates/etcd-service-descriptor.yaml
weicao: for now, maybe you should define both EXTERNAL_ETCD_SERVICE_ENDPOINT and ETCD_COMP_SERVICE_ENDPOINT vars, and write a sh function getEndpoint() to resolve.

addons/apecloud-mysql/values.yaml
```diff
@@ -108,3 +108,6 @@ wesqlscale:
 ## if you install multiple releases of this chart.
 ## If specified, the cluster definition will use it as name.
 resourceNamePrefix: ""
+
+etcd:
+  etcdCrdName: etcd-0.9.0
```
weicao, line 113: etcdCmpdName

addons/apecloud-mysql/templates/cr-check.yaml
```diff
@@ -0,0 +1,4 @@
+{{- $namespace := .Release.Namespace }}
+{{- if not (lookup "apps.kubeblocks.io/v1alpha1" "ComponentDefinition" $namespace .Values.etcd.etcdCrdName) }}
+  {{- fail (printf "The ComponentDefinition %s is required but does not exist in namespace %s." .Values.etcd.etcdCrdName $namespace) }}
+{{- end }}
```
weicao, line 4: how will this check executed

addons/apecloud-mysql/templates/_helpers.tpl
```diff
@@ -195,6 +195,7 @@ serviceRefDeclarations:
     serviceRefDeclarationSpecs:
       - serviceKind: etcd
         serviceVersion: "^*"
+    optional: true
```
weicao, line 198: It seems etcd related info should not appear in apecloud-mysql.spec.common, because if proxy is not enabled, vttablet won't be launched.

addons/apecloud-mysql/scripts/etcd-post-start.sh
weicao: I see

addons/apecloud-mysql/scripts/etcd-post-start.sh
```diff
@@ -13,13 +19,15 @@ done
 
 servers=${servers:1}
 
+echo $servers
+
 if [[ ${ETCDCTL_API} -eq "3" ]]; then
-  output=$(etcdctl --endpoints="${servers}" get "/vitess/${KB_NAMESPACE}/${KB_CLUSTER_NAME}/global" --prefix --keys-only)
+  output=$(etcdctl --endpoints="${servers}" get "/vitess/${KB_NAMESPACE}/${KB_CLUSTER_NAME}/global/$cell" --prefix --keys-only)
```
weicao, line 25: "/vitess/${KB_NAMESPACE}/${KB_CLUSTER_NAME}/global/$cell"

or 

"/vitess/${KB_NAMESPACE}/${KB_CLUSTER_NAME}/$cell"

?

addons/apecloud-mysql/templates/cr-check.yaml
skyrise-l: In execute helm install

addons/apecloud-mysql/templates/_helpers.tpl
skyrise-l: However, cmpd doesn't seem to be able to determine whether etcd is needed at the moment. When install addon, it doesn't know the actual state of the cluster. The current design is that if etcd is not needed, the variables rendered for etcd will be empty. To accommodate this, I set optional: true.

addons/apecloud-mysql/templates/cr-check.yaml
skyrise-l: In fact, this is indeed a stopgap measure, because this design requires the etcd addon to be set up regardless of whether proxy and etcd are needed. Ideally, this determination should be made during cluster installation, but it cannot be done.

weicao: > In fact, this is indeed a stopgap measure, because this design requires the etcd addon to be set up regardless of whether proxy and etcd are needed. Ideally, this determination should be made during cluster installation, but it cannot be done.

I see, could be left as a future work.

addons-cluster/apecloud-mysql-cluster/values.yaml
```diff
@@ -36,9 +36,14 @@ storage: 20
...
 ## if mode is standalone, proxyEnabled is always false
 ## if mode is raftGroup, proxyEnabled can be true or false
-##
+## 
 proxyEnabled: false
 
+## @param enableLocalEtcd if true, use the local etcd
+## if mode is standalone, enableLocalEtcd is always false
+## if mode is raftGroup, enableLocalEtcd can be true
+enableLocalEtcd: false
```
weicao, line 45: seems localEtcdEnabled is more consistent with other variables

addons/apecloud-mysql/scripts/etcd-post-start.sh
```diff
@@ -1,5 +1,18 @@
 #!/bin/bash
-endpoints=${ETCD_SERVER:-'127.0.0.1:2379'}
+if [ -n "$LOCAL_ETCD_POD_LIST" ]; then
+  IFS=',' read -ra ETCD_POD_ARRAY <<< "$LOCAL_ETCD_POD_LIST"
+  endpoints=""
+  for pod in "${ETCD_POD_ARRAY[@]}"; do
+    endpoints+="${pod}.${LOCAL_ETCD_HEADLESS}.${KB_NAMESPACE}.svc.cluster.local:${LOCAL_ETCD_PORT},"
```
weicao, line 6: why not use LOCAL_ETCD_HEADLESS?

weicao: and you may have a try with vars expression

skyrise-l: Isn't it true that I am currently using LOCAL_ETCD_HEADLESS, meaning directly using the HEADLESS service?

weicao: > Isn't it true that I am currently using LOCAL_ETCD_HEADLESS, meaning directly using the HEADLESS service?

yeah, you're correct, the pod's headless svc should be pod name ++ Local_ETC_HEADLESSS,

weicao, line 6: ".cluster.local" should be replaced by {{ .Values.clusterDomain }}

addons/apecloud-mysql/scripts/etcd-clean.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 
-endpoints=${ETCD_SERVER:-'127.0.0.1:2379'}
+if [ -n "$SERVICE_ETCD_ENDPOINT" ]; then
+  endpoints=$SERVICE_ETCD_ENDPOINT
+else
+# local etcd no need to clean
+  exit 0 
+fi
```
weicao, line 8: doesn't set endpoints properly

skyrise-l: I don't understand what the problem is. etcd-clean.sh only needs to be executed when using an external etcd. This can be determined through SERVICE_ETCD_ENDPOINT. If it is empty, it means that an external etcd is not being used, and the process can be terminated immediately.

addons/apecloud-mysql/scripts/etcd-post-start.sh
skyrise-l: 截屏2024-07-23 10 21 23
Do you mean adding an environment variable and then referencing it in the script to avoid hardcoding?

weicao: yeah, this is a solution

addons/apecloud-mysql/scripts/etcd-clean.sh
weicao: I see, you're correct.

addons/apecloud-mysql/templates/_helpers.tpl
```diff
@@ -169,8 +169,9 @@ lifecycleActions:
       exec:
         command:
           - /scripts/switchover-without-candidate.sh
-  {{/*
   accountProvision:
+   # builtinHandler: wesql
```
weicao, line 173: why comment this line?

addons/apecloud-mysql/scripts/etcd-clean.sh
```diff
@@ -18,11 +23,17 @@ servers=${servers:1}
 echo $servers
 
 echo "Deleting all keys with prefix /vitess/${KB_NAMESPACE}/${KB_CLUSTER_NAME} from Etcd server at ${endpoints}..."
-etcdctl --endpoints $servers del /vitess/${KB_NAMESPACE}/${KB_CLUSTER_NAME} --prefix
+
+# Set different deletion methods according to different etcdctl versions.
+if [[ ${ETCDCTL_API} == "3" ]]; then
```
weicao, line 28: it's better changed to :

```
if [[ ${ETCDCTL_API} == "2" ]]; then
    ...
else
   ...
fi
```

so that if ETCDCTL_API is not set in extra_env, "3" is the default value.

skyrise-l: yes I will modify it

addons/apecloud-mysql/templates/cmpd-proxy.yaml
```diff
@@ -52,8 +53,8 @@ spec:
             value: "15999"
           - name: ETCD_SERVER
             value: $(SERVICE_ETCD_ENDPOINT)
```
weicao, line 55: seems useless now, can be deleted

addons/apecloud-mysql/templates/cmpd-proxy.yaml
```diff
@@ -151,6 +176,8 @@ spec:
...
```
weicao, line 176: as the above

addons/apecloud-mysql/templates/_helpers.tpl
```diff
@@ -223,9 +225,31 @@ vars:
...
+        compDef: {{ .Values.etcd.etcdCmpdName }}
+        optional: true
+        instanceNames: Optional
+  - name: LOCAL_ETCD_HEADLESS
+    valueFrom:
+      serviceVarRef:
+        compDef: {{ .Values.etcd.etcdCmpdName }}
+        name: headless
+        optional: true
+        host: Optional
```
weicao, line 241:
```
- name: LOCAL_ETCD_POD_FQDN
   valueFrom:
        componentVarRef:
          compDef: {{ .Values.etcd.etcdCmpdName }}
          optional: false
          podFQDNs: Required
```
can use above env to get pod FQDN address, (lacks clusterDomain, however)

weicao (approved):

PR #828, opened 2024-07-18
fix: kafka image hardcode

yipeng1030:
Move three opsdefinition YAMLs to templates directory, now support template.
imageimage
Refer to kubeblocks/deploy/opsdefinition/switchover-preconfitions and kubeblocks/deploy/templates/opsdefinition.yaml
image

### DIFF
--- a/addons/kafka/opsdefinitions/quota.yaml
+++ b/addons/kafka/opsdefinitions/quota.yaml
@@ -56,7 +56,7 @@ spec:
         podSpec:
           containers:
             - name: kafka-topic
-              image: docker.io/bitnami/kafka:3.3.2-debian-11-r54
+              image: apecloud-registry.cn-zhangjiakou.cr.aliyuncs.com/apecloud/kafka:3.3.2-debian-11-r54
               imagePullPolicy: IfNotPresent
               command:
                 - bash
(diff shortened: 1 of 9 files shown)


addons/kafka/opsdefinitions/quota.yaml
```diff
@@ -56,7 +56,7 @@ spec:
         podSpec:
           containers:
             - name: kafka-topic
-              image: docker.io/bitnami/kafka:3.3.2-debian-11-r54
+              image: apecloud-registry.cn-zhangjiakou.cr.aliyuncs.com/apecloud/kafka:3.3.2-debian-11-r54
```
weicao, line 59: why not use template?

{{ .Values.images.registry | default "docker.io" }}/{{ .Values.images.kafka.repository }}:{{ default .Chart.AppVersion .Values.images.kafka.tag }}

yipeng1030: These three YAML documents are not defined within the templates directory, and therefore do not support template. Should I move them there and use template?

ldming: Should use template, otherwise the image registry can not be modified, and it will still be hardcoded.

ldming (approved):

PR #841, opened 2024-07-22
feat[starrocks]: support KB v9 API

iziang:


nayutah (approved):

addons/starrocks-ce/templates/componentdefinition-be.yaml
```diff
@@ -0,0 +1,137 @@
...
+  - name: data
+    needSnapshot: true
+  vars:
+  - name: FE_DISCOVERY_SERVICE_NAME
+    valueFrom:
+      serviceVarRef:
+        compDef: {{ include "fe.componentDefName" . }}
+        name: fe
+        host: Required
+  - name: MYSQL_PWD
```
weicao, line 42: PASSWD is better, PWD is the abbreviation for 'Print Working Directory'.

addons/starrocks-ce/templates/componentdefinition-be.yaml
```diff
@@ -0,0 +1,137 @@
...
+      command:
+      - bash
+      - -c
+      - |
+        /opt/starrocks/be_entrypoint.sh ${FE_DISCOVERY_SERVICE_NAME}
+      env:
+      - name: HOST_TYPE
+        value: FQDN
+      - name: TZ
+        value: Asia/Shanghai
```
weicao, line 63: put it in values.yaml ?

addons/starrocks-ce/templates/_helpers.tpl
```diff
@@ -74,3 +74,11 @@ be.conf: |
 {{ .Values.be.config | indent 2 }}
 {{- end }}
 {{- end }}
+
+{{- define "fe.componentDefName" -}}
+{{ include "starrocks.name" . }}-fe
```
weicao, line 79: should we add a version for cmpd?

addons/starrocks-ce/templates/componentdefinition-be.yaml
iziang: MYSQL_PWD is a native environment variable supported by the MySQL.

addons/starrocks-ce/templates/_helpers.tpl
iziang: StarRocks currently supports versions 3.2.2 and 3.3.0, both of which are compatible with the same cmpd API. We can consider adding version information to the cmpd name in the future when incompatible versions arise.

addons/starrocks-ce/templates/componentdefinition-be.yaml
iziang: done

iziang: https://github.com/apecloud/kubeblocks-addons/pull/844/files

addons/starrocks-ce/templates/_helpers.tpl
weicao: Yeah, I mean add a version for the cmpd cr, not for the starrocks docker image version. So that in the future, if the cmpd CR upgrades, may be two versions of cmpd CR can co-exist.

addons/starrocks-ce/templates/componentdefinition-fe.yaml
```diff
@@ -0,0 +1,170 @@
...
+  lifecycleActions:
+    accountProvision:
+      customHandler:
+        image: {{ .Values.image.registry | default "docker.io" }}/{{ .Values.image.fe.repository }}:{{ default .Values.image.fe.tag }}
+        exec:
+          command:
+          - /bin/bash
+          - -c
+          - |
+            mysql -P{{ .Values.fe.queryPort }} -h127.0.0.1 -u${STARROCKS_USER} -e "SET PASSWORD = PASSWORD('${STARROCKS_PASSWORD}')"
```
weicao, line 54: what's the difference with fe-post-start.sh

PR #846, opened 2024-07-22
chore: optimize settings for polardbx

Chen-speculation:
make polardbx component image source accessable  in apecloud

### DIFF
--- a/addons/polardbx/values.yaml
+++ b/addons/polardbx/values.yaml
@@ -62,7 +62,7 @@ images:
 
   # Images for DN exporter
   prom:
-    repository: prom
+    repository:
     pullPolicy: IfNotPresent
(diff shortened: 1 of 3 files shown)


addons/polardbx/values.yaml
```diff
@@ -21,7 +21,7 @@ images:
   polardbx:
     pullPolicy: IfNotPresent
     # Repo of polardbx default images. Default is polardbx.
-    repository: polardbx
+    repository:
```
nayutah, line 24: refine the annotation

Chen-speculation: it's correct .default value is still polardbx
image

nayutah (approved):

weicao (approved):

PR #864, opened 2024-07-29
chore: remove all cluster version CRs

leon-inf:


addons/xinference/templates/clusterdefinition.yaml
```diff
@@ -29,6 +29,8 @@ spec:
       podSpec:
         containers:
           - name: xinference
+            image:  {{ .Values.image.registry | default "docker.io" }}/{{ .Values.image.repository}}:{{ default .Chart.AppVersion .Values.image.tag }}
+            # image:  {{ .Values.image.registry | default "docker.io" }}/{{ .Values.image.repository}}:{{ default .Chart.AppVersion .Values.image.tag }}-cpu
```
weicao, line 33: seems should introduce ComponentVersions to support both images?

weicao (approved):

PR #873, opened 2024-07-31
chore: cmpd for risingwave

leon-inf:


addons/risingwave/templates/clusterdefinition.yaml
```diff
@@ -1,441 +0,0 @@
-apiVersion: apps.kubeblocks.io/v1alpha1
```
nayutah, line 1: Add ClusterDefinition & ComponentVersion chart for cloud integration

weicao (approved):

PR #874, opened 2024-07-31
feat: add a new addon gbase8c 

skyrise-l:
This PR adds an addon for the domestic database GBase 8c. The current primary-secondary (HA) mode is available, but the distributed mode is not yet supported. The current limitations are as follows:
**1 Operating System Dependency:** GBase operation scripts are designed for several operating systems, but for Linux, there is a strong dependency on systemd. Since running and deploying the GBase cluster requires the image's cgroup version to match the local machine, the current image cgroup version is v1 (For example, if you are using an Ubuntu Linux system, the version should be less than 21.04. You can refer to the link [here](https://github.com/opencontainers/runc/blob/main/docs/cgroup-v2.md).). 
Moreover, the GBase container's startup command must be /sbin/init (as pid 1), so the database installation and startup commands are executed in postStart.
**2 Installation Speed:** GBase 8c's installation mode involves the primary node distributing data to each node, resulting in very slow installation speed. Deploying the entire cluster takes approximately 4-10 minutes, depending on the number of nodes.
**3 User Creation Limitations:** GBase 8c does not allow remote connections using the initial user. Therefore, new management users must be created on the primary node using the initial user. Currently, the creation of the kbadmin user is implemented, but the accountProvision has no practical significance since remote user creation is not possible.
**4 Scaling Limitations:** Due to GBase 8c's own operational restrictions, scaling must be done on the initial primary node (performing a primary-secondary switch does not genuinely change the initial primary node). Therefore, it is necessary to ensure the initial primary node is available.
**5 Primary-Secondary Switch Support:** Non-designated candidate primary-secondary switching is currently not supported. If needed, the cm server component can be added later to further support this feature.


addons/gbase/templates/cmpd-coordinator.yaml
```diff
@@ -0,0 +1,53 @@
...
+            name: data
+          - name: scripts
+            mountPath: /scripts
+          - name: ssh-key
+            mountPath: /ssh-key
+            readOnly: true
+    volumes:
+    - name: ssh-key
+      secret:
+        secretName: gbase-ssh-rsa-secret
```
shanshanying, line 53: gbase-ssh-rsa-secret, the secret name is a constant.
we can define `volumeMounts` in cmpd, and define volumes in 'cluster.spec.componentSpecs.volumes'.  and the secret name could be any name given by user.

shanshanying:
1. remove gbasetmp files
2. better to use go template to reduce redundant fragments.
3. define 'volumesMounts' in cmpd.spec, and define 'volumes' in cluster.spec.componentSpecs.volumes', when mount a secret.  secret name should not be a constant.

PR #876, opened 2024-08-01
chore: cmpd for flink

leon-inf:


nayutah (approved):

addons/flink/templates/clusterdefinition.yaml
```diff
@@ -1,111 +0,0 @@
-apiVersion: apps.kubeblocks.io/v1alpha1
```
nayutah, line 1: Add ClusterDefinition & ComponentVersion chart for cloud integration

PR #883, opened 2024-08-01
feat: add a gbase8c addon with HA mode 

skyrise-l:
This PR adds an addon for the domestic database GBase 8c. The current primary-secondary (HA) mode is available.
The full For the complete GBase PR, please refer to https://github.com/apecloud/apecloud-addons/pull/188. It includes both the primary-secondary HA mode and the distributed mode.
 The current limitations are as follows:
1 Operating System Dependency: GBase operation scripts are designed for several operating systems, but for Linux, there is a strong dependency on systemd. Since running and deploying the GBase cluster requires the image's cgroup version to match the local machine, the current image cgroup version is v1 (For example, if you are using an Ubuntu Linux system, the version should be less than 21.04. You can refer to the link [here](https://github.com/opencontainers/runc/blob/main/docs/cgroup-v2.md).).
Moreover, the GBase container's startup command must be /sbin/init (as pid 1), so the database installation and startup commands are executed in postStart.
2 Installation Speed: GBase 8c's installation mode involves the primary node distributing data to each node, resulting in very slow installation speed. Deploying the entire cluster takes approximately 4-10 minutes, depending on the number of nodes.
3 User Creation Limitations: GBase 8c does not allow remote connections using the initial user. Therefore, new management users must be created on the primary node using the initial user. Currently, the creation of the kbadmin user is implemented, but the accountProvision has no practical significance since remote user creation is not possible.
4 Scaling Limitations: Due to GBase 8c's own operational restrictions, scaling must be done on the initial primary node (performing a primary-secondary switch does not genuinely change the initial primary node). Therefore, it is necessary to ensure the initial primary node is available.
5 Primary-Secondary Switch Support: Non-designated candidate primary-secondary switching is currently not supported. If needed, the cm server component can be added later to further support this feature.


addons-cluster/gbase-cluster/templates/ssh_secret_generate.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: batch/v1
```
shanshanying, line 1: maybe we can render ssh keys in a go tpl fun , to simplify the procedure. (such that all pods of the same cmp will share the same keys, and other pods for backup/restore, can mount ssh keys, will discuss with @sophon-zt

PR #884, opened 2024-08-01
chore: cmpd for polardbx

leon-inf:


nayutah (approved):

addons/polardbx/templates/clusterDefinition.yaml
```diff
@@ -1,707 +0,0 @@
-apiVersion: apps.kubeblocks.io/v1alpha1
```
nayutah, line 1: Add ClusterDefinition & ComponentVersion chart for cloud integration

leon-inf: It's weird to introduce the cd & cmpv CR but without necessary content to define within them.

PR #897, opened 2024-08-06
feat: rabbitmq expand the range of supported versions from 3.8 to 3.13

xuriwuyun:


addons/rabbitmq/templates/componentversion.yaml
```diff
@@ -10,9 +10,39 @@ spec:
...
+        - 3.11.28
+        - 3.10.25
+        - 3.9.29
+        - 3.8.14
   releases:
     - name: 3.13.2
       changes:
       serviceVersion: 3.13.2
       images:
         rabbitmq: {{ .Values.image.registry | default "docker.io" }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}
```
weicao, line 23: replace {{ .Values.image.tag }} to 3.13.2-management ?

addons/rabbitmq/templates/componentversion.yaml
```diff
@@ -10,9 +10,39 @@ spec:
...
+    - name: 3.12.14
+      changes:
+      serviceVersion: 3.12.14
```
weicao, line 26: I think serviceVersion should be 3.12,3.11,3.10,3.9

nayutah (approved):

PR #899, opened 2024-08-07
fix postgresql etcd refer problem

skyrise-l:
The previous PR had a slight issue as it didn't use the latest solution. This PR has fixed the etcd-clean script and the omission of the etcd optional settings for the PG12 and PG14 versions.

### DIFF
--- a/addons-cluster/postgresql/templates/rbac.yaml
+++ b/addons-cluster/postgresql/templates/rbac.yaml
@@ -1,1 +1,37 @@
 {{- include "kblib.rbac" . }}
+---
+{{- if .Values.etcd.proxyEnabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: {{ printf "%s-etcd-clean-role" (include "kblib.clusterName" .) }}
+  labels: 
+    {{ include "kblib.clusterLabels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": "post-install, pre-delete"
+    "helm.sh/hook-weight": "0"
+rules:
+- apiGroups: [""]
+  resources: ["pods/exec"]
+  verbs: ["create"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ printf "%s-etcd-clean" (include "kblib.clusterName" .) }}
+  namespace: {{ .Release.Namespace }} 
+  labels: 
+    {{ include "kblib.clusterLabels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": "post-install, pre-delete"
+    "helm.sh/hook-weight": "0"
+subjects:
+- kind: ServiceAccount
+  name: kb-{{ include "kblib.clusterName" . }}
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: Role
+  name: {{ printf "%s-etcd-clean-role" (include "kblib.clusterName" .) }}
+  apiGroup: rbac.authorization.k8s.io
+{{ end }}
--- a/addons/postgresql/scripts/etcd-clean.sh
+++ b/addons/postgresql/scripts/etcd-clean.sh
@@ -4,29 +4,11 @@ if [ -z "$ETCD_SERVER" ]; then
   exit 0
 fi
 
-IFS=',' read -ra ADDR <<< "$KB_CLUSTER_COMPONENT_POD_IP_LIST"
-FIRST_IP=${ADDR[0]}
+echo "Executing pkill -f patroni in pod"
 
-PATRONI_API_URL="http://$FIRST_IP:8008/patroni"
-PATRONI_SHUTDOWN_URL="http://$FIRST_IP:8008/shutdown"
+pkill -f patroni
 
-status_code=$(curl -s -o /dev/null -w "%{http_code}" $PATRONI_API_URL)
-
-if [ "$status_code" -ne 200 ]; then
-  echo "Failed to access Patroni API at $PATRONI_API_URL. HTTP response code: $status_code."
-  exit 0
-fi
-
-response=$(curl -s -o /dev/null -w "%{http_code}" -XPOST $PATRONI_SHUTDOWN_URL)
-
-if [ "$response" -eq 200 ]; then
-  echo "Successfully terminated Patroni service at $FIRST_IP."
-else
-  echo "Failed to terminate Patroni service at $FIRST_IP. HTTP response code: $response."
-  sleep 10
-  exit 0
-fi
-sleep 10
+sleep 3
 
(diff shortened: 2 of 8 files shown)


addons/postgresql/templates/componentdefinition-15.yaml
```diff
@@ -201,6 +196,10 @@ spec:
           runAsUser: 0
         command:
           - /kb-scripts/setup.sh
+        lifecycle:
+          preStop:
+            exec: 
+              command: ["/kb-scripts/etcd-clean.sh"]
```
weicao, line 202: It seems not appropriate for this to be executed when each Pod is terminating

weicao: The first Pod in $KB_CLUSTER_COMPONENT_POD_IP_LIST is selected and its terminate API is called

weicao: And all keys 're deleted

addons/postgresql/scripts/etcd-clean.sh
```diff
@@ -4,29 +4,11 @@ if [ -z "$ETCD_SERVER" ]; then
...
-
-if [ "$response" -eq 200 ]; then
-  echo "Successfully terminated Patroni service at $FIRST_IP."
-else
-  echo "Failed to terminate Patroni service at $FIRST_IP. HTTP response code: $response."
-  sleep 10
-  exit 0
-fi
-sleep 10
+sleep 3
```
weicao, line 11: sleep 3 doesn't gaurantee patroni is killed?

addons/postgresql/templates/componentdefinition-12.yaml
```diff
@@ -147,6 +149,11 @@ spec:
             - -c
           args:
             - curl -s http://$(KB_REPLICATION_PRIMARY_POD_FQDN):8008/switchover -XPOST -d '{"leader":"$(KB_REPLICATION_PRIMARY_POD_NAME)"}'
+    preTerminate:
+      customHandler:
+        exec:
+          image: {{ .Values.etcdCleanToolsImage.registry | default "docker.io" }}/{{ .Values.etcdCleanToolsImage.repository }}:{{ .Values.etcdCleanToolsImage.tags }}
```
weicao, line 155: Does it work? I remember the controller enforces all Actions must use the same image in the code, since they're executed by the kbagent (previous, lorry) container. And the image of kbagent is determined by `exec.image`.

addons-cluster/postgresql/templates/rbac.yaml
```diff
@@ -1 +1,37 @@
-{{- include "kblib.rbac" . }}
\ No newline at end of file
+{{- include "kblib.rbac" . }}
+---
+{{- if .Values.etcd.proxyEnabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: {{ printf "%s-etcd-clean-role" (include "kblib.clusterName" .) }}
```
Y-Rookie, line 8: Introducing a new role in Helm to implement this functionality is inappropriate because the lifecycle of this role is tied to Helm rather than the cluster's lifecycle. This poses significant challenges for subsequent maintenance and management.

A correct approach is that if a component requires additional permissions, it should be explicitly declared before it can be applied in scenarios such as lifecycle actions. The management of these permissions should be part of the controller's functionality. You can consider extending the implementation of the current PolicyRules (not yet implemented) in cmpd to meet such requirements.

addons/postgresql/templates/clusterdefinition.yaml
```diff
@@ -14,4 +14,4 @@ spec:
     - name: replication
       components:
         - name: postgresql
-          compDef: {{ include "postgresql.componentDefNamePrefix" . }}
\ No newline at end of file
+          compDef:  postgresql-14 #{{ include "postgresql.componentDefNamePrefix" . }}
```
weicao, line 17: why bind to postgresql-14?

addons/postgresql/templates/componentdefinition-12.yaml
skyrise-l: yes it works, The preterminate action involves launching another pod without using Lorry.

addons/postgresql/scripts/etcd-clean.sh
```diff
@@ -1,34 +1,50 @@
...
+    echo "Attempting to execute pkill -f /usr/local/bin/patroni in pod: $POD_NAME"
 
-status_code=$(curl -s -o /dev/null -w "%{http_code}" $PATRONI_API_URL)
+    MAX_ATTEMPTS=3
+    ATTEMPT=1
+    SUCCESS=0
+    
+    while [ "$ATTEMPT" -le "$MAX_ATTEMPTS" ]; do
+      echo "Attempt $ATTEMPT: Executing pkill -f /usr/local/bin/patroni in pod: $POD_NAME"
+      kubectl exec $POD_NAME -- pkill -f /usr/local/bin/patroni
```
weicao, line 19: Do not use kubectl exec in etcd-clean.sh.
why not shutdown all patroni processes one by one using PATRONI_SHUTDOWN_URL?

addons-cluster/postgresql/templates/rbac.yaml
```diff
@@ -1 +1,37 @@
-{{- include "kblib.rbac" . }}
\ No newline at end of file
+{{- include "kblib.rbac" . }}
+---
+{{- if .Values.etcd.proxyEnabled }}
```
weicao, line 3: why the name is etcd.proxyEnabled?

there is no proxy

weicao: .Values.useExternalEtcd ?

addons/postgresql/scripts/etcd-clean.sh
weicao, line 19: A particularly challenging corner case arises when a Patroni instance becomes unresponsive due to pod failure, process timeout, or network partition (which is even more difficult to handle), causing its shutdown API to fail. In such situations, how should the etcd-clean process proceed? 

Under normal circumstances, etcd-clean should report an error and exit, waiting for the next retry of the preTerminate task. 

This introduces two key requirements:

- etcd-clean needs to be idempotent. This means that Patroni processes that were shut down in the previous round of etcd-clean should be detected and skipped in subsequent rounds.
- The stop cluster/component API should include a force flag. This flag would allow the operation to bypass the preTerminate step and release resources in cases where preTerminate execution fails. In the context of etcd-clean, this failure might occur due to Patroni processes not being completely terminated, potentially leaving uncleaned garbage objects in the shared etcd.

addons/postgresql/templates/componentdefinition-12.yaml
weicao: The most recent KB version has replaced lorry to kbagent, which checks exec.image of all actions.

PR #904, opened 2024-08-08
fix: set zookeeper volume size via kblib and change default storage size to 10Gi

kubeJocker:
When creating a Zookeeper cluster, the minimum storage value is set to 10Gi, but it's actually created with 4Gi.

fix cloud issue #5999

### DIFF
--- a/addons-cluster/zookeeper/templates/cluster.yaml
+++ b/addons-cluster/zookeeper/templates/cluster.yaml
@@ -35,4 +35,4 @@ spec:
               - ReadWriteOnce
             resources:
               requests:
-                storage: {{ print .Values.log "Gi" }}
+                storage: {{ print .Values.logStorage "Gi" }}
--- a/addons-cluster/zookeeper/templates/cluster.yaml
+++ b/addons-cluster/zookeeper/templates/cluster.yaml
@@ -14,20 +14,24 @@ spec:
       serviceAccountName: {{ include "kblib.serviceAccountName" . }}
       {{- include "kblib.componentMonitor" . | indent 6 }}
       {{- include "kblib.componentResources" . | indent 6 }}
+      {{- include "kblib.componentStorages" . | indent 6 }}
+      {{- include "kblib.componentServices" . | indent 6 }}
       volumeClaimTemplates:
-        - name: data
+        - name: data # ref clusterDefinition components.containers.volumeMounts.name
           spec:
-            storageClassName: {{ .Values.persistence.data.storageClassName }}
             accessModes:
               - ReadWriteOnce
             resources:
               requests:
-                storage: {{ .Values.persistence.data.size }}
+                storage: {{ print .Values.storage "Gi" }}
+            {{- if .Values.storageClassName }}
+            storageClassName: {{ .Values.storageClassName | quote }}
+            {{- end }}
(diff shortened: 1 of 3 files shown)


nayutah (approved):

addons-cluster/zookeeper/values.yaml
```diff
@@ -23,23 +23,7 @@ memory: 2
...
-    ## @param shard[*].persistence.size Size of data volume
-    ##
-    size: 4Gi
-  log:
-    storageClassName:
-    ## @param shard[*].persistence.size Size of data volume
-    ##
-    size: 4Gi
+
+storage: 10
```
weicao, line 27: This naming is unclear; it appears to refer to defaultDataVolumeStorageSize.

addons-cluster/zookeeper/values.yaml
```diff
@@ -23,23 +23,7 @@ memory: 2
...
+
+log: 2
```
weicao, line 29: defaultLogVolumeStorageSize

addons-cluster/zookeeper/templates/cluster.yaml
```diff
@@ -14,20 +14,26 @@ spec:
       serviceAccountName: {{ include "kblib.serviceAccountName" . }}
       {{- include "kblib.componentMonitor" . | indent 6 }}
       {{- include "kblib.componentResources" . | indent 6 }}
+      {{- include "kblib.componentStorages" . | indent 6 }}
+      {{- include "kblib.componentServices" . | indent 6 }}
       volumeClaimTemplates:
-        - name: data
+        - name: data # ref clusterDefinition components.containers.volumeMounts.name
```
weicao, line 20: "kblib.componentStorages" includes such an entry: 
```
volumeClaimTemplates
  - name: data
```
Why there is a duplicate one?

kubeJocker: there is a missing commit, so I pushed it

addons-cluster/zookeeper/values.yaml
kubeJocker: maybe logStorage is better

addons-cluster/zookeeper/templates/cluster.yaml
```diff
@@ -14,20 +14,25 @@ spec:
       serviceAccountName: {{ include "kblib.serviceAccountName" . }}
       {{- include "kblib.componentMonitor" . | indent 6 }}
       {{- include "kblib.componentResources" . | indent 6 }}
+      {{- include "kblib.componentServices" . | indent 6 }}
```
weicao, line 17: "kblib.componentServices" works with the legacy API (override service defined in `clusterDef.componentDefs[*].service`), the `clusterDef.componentDefs` fields have already been deleted, so it should not be included in main branch.

However, addons in 0.9 branch still have `clusterDef.componentDefs`.

weicao: do not need to do any modification, @Y-Rookie will handle this situation uniformly.

weicao (approved):

kubeJocker:
cherry-pick release-0.9

Issue #906, opened 2024-08-09
[BUG] Redis cluster nodes fail to join cluster due to FQDN DNS resolving lag


leonliao:
**Describe the bug**
For Redis Custer, current `addons/redis/redis-cluster-scripts/redis-cluster-server-start.sh` uses FQDN to add a node to cluster. But after a redis node pod rebuild or creation, due the DNS cache refresh or new DNS resolving may take effect before the cluster join command, it is possible that the cluster joining could fail.

**To Reproduce**

Simulate a pod leaving the cluster and rejoin.
1. Login a slave pod in Redis Cluster, execute the `redis-cli --cluster del-node $current_node_ip_and_port $current_node_cluster_id`, simulating [addons/redis/redis-cluster-scripts/redis-cluster-replica-member-leave.sh](https://github.com/apecloud/kubeblocks-addons/blob/28a0e2c7c94f7b249211786ad4ca6b8182a49d29/addons/redis/redis-cluster-scripts/redis-cluster-replica-member-leave.sh#L15)
2. Delete the slave pod
3. There could be chances that seeing errors like below:
```

# DNS staled cache points the FQDN to old POD
+ current_node_with_port=redis-cluster-shard-kqr-0.redis-cluster-shard-kqr-headless.default.svc:6379
+ set +x
scale out replica replicated command: redis-cli --cluster add-node redis-cluster-shard-kqr-0.redis-cluster-shard-kqr-headless.default.svc:6379 redis-cluster-shard-kqr-1.redis-cluster-shard-kqr-headless.default.svc:6379 --cluster-slave --cluster-master-id 587b870ce5809309004f785d0837ed11f82a33e5 -a ********
Warning: Using a password with '-a' or '-u' option on the command line interface may not be safe.
Could not connect to Redis at redis-cluster-shard-kqr-0.redis-cluster-shard-kqr-headless.default.svc:6379: Connection timed out

# DNS taking effect after the add-node command
+ current_node_with_port=redis-cluster-shard-btn-0.redis-cluster-shard-btn-headless.default.svc:6379
+ set +x
scale out replica replicated command: redis-cli --cluster add-node redis-cluster-shard-btn-0.redis-cluster-shard-btn-headless.default.svc:6379 redis-cluster-shard-btn-1.redis-cluster-shard-btn-headless.default.svc:6379 --cluster-slave --cluster-master-id 75203464beaf7403fc82eeeb24c98f9a9590a054 -a ********
[...]

PR #908, opened 2024-08-09
fix: Retry joining cluster to make sure DNS refreshed (#906)

leonliao:
This PR fixes the issue https://github.com/apecloud/kubeblocks-addons/issues/906.

### DIFF
--- a/addons/redis/redis-cluster-scripts/redis-cluster-server-start.sh
+++ b/addons/redis/redis-cluster-scripts/redis-cluster-server-start.sh
@@ -377,19 +377,37 @@ scale_redis_cluster_replica() {
     replicated_command="redis-cli --cluster add-node $current_node_with_port $primary_node_endpoint_with_port --cluster-slave --cluster-master-id $primary_node_cluster_id -a $REDIS_DEFAULT_PASSWORD"
   fi
   echo "scale out replica replicated command: $replicated_command" | sed "s/$REDIS_DEFAULT_PASSWORD/********/g"
-  replicated_output=$($replicated_command)
-  replicated_exit_code=$?
-  set -x
-  echo "Scale out replica replicated command result: $replicated_output"
-  if [ $replicated_exit_code -ne 0 ]; then
+  # Keep trying until add-node is done.
+  # And the current_pod_fqdn DNS cache time is normally 30 seconds, thus make 30 attempts
+  local max_attempts=30
+  local attempt=1
+
+  while [ $attempt -le $max_attempts ]
+  do
+    # Avoid exiting with non-zero code and avoid printing password 
+    set +ex
+    replicated_output=$($replicated_command)
+    replicated_exit_code=$?
+    set -ex
+    echo "Attempt $attempt: Scale out replica replicated command result: $replicated_output"
+    if [ $replicated_exit_code -eq 0 ]; then
+      break
+    fi
     if [[ $replicated_output == *"is not empty"* ]]; then
       echo "Replica is not empty, Either the node already knows other nodes (check with CLUSTER NODES) or contains some key in database 0"
+      break
     else
-      echo "Failed to add the node $current_pod_fqdn to the cluster in scale_redis_cluster_replica, shutdown redis server"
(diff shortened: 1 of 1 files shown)


shanshanying:
Hi @leonliao, Thanks for your contribution. 

 @Y-Rookie PTAL.

addons/redis/redis-cluster-scripts/redis-cluster-server-start.sh
```diff
@@ -377,19 +377,37 @@ scale_redis_cluster_replica() {
...
-  replicated_exit_code=$?
-  set -x
-  echo "Scale out replica replicated command result: $replicated_output"
-  if [ $replicated_exit_code -ne 0 ]; then
+  # Keep trying until add-node is done.
+  # And the current_pod_fqdn DNS cache time is normally 30 seconds, thus make 30 attempts
+  local max_attempts=30
+  local attempt=1
+
+  while [ $attempt -le $max_attempts ]
```
nayutah, line 385: 有 while 循环的 shell 脚本最好加一下对 SIGTERM 的信号响应：

// 捕获信号
trap "exit_handler" SIGTERM SIGINT

exit_handler() {
    echo "Received stop signal, cleaning up..."
    do some clean job
    exit 0
}

leonliao: In this file, there is a retry() function, which has an until loop in it. Also, a while true loop is just behind the code I submitted. So a global `exit_handler` is needed for this script file?

BTW: Because the retry function just validates the return code, but not checking the error msgs like the "is not empty", thus the retry() function was not used by my code.

nayutah: add a global one to make the scripts stop and exit ASAP

addons/redis/redis-cluster-scripts/redis-cluster-server-start.sh
```diff
@@ -377,19 +377,37 @@ scale_redis_cluster_replica() {
     replicated_command="redis-cli --cluster add-node $current_node_with_port $primary_node_endpoint_with_port --cluster-slave --cluster-master-id $primary_node_cluster_id -a $REDIS_DEFAULT_PASSWORD"
   fi
   echo "scale out replica replicated command: $replicated_command" | sed "s/$REDIS_DEFAULT_PASSWORD/********/g"
-  replicated_output=$($replicated_command)
-  replicated_exit_code=$?
-  set -x
```
weicao, line 382: Remove the matching set +x along with this set -x

addons/redis/redis-cluster-scripts/redis-cluster-server-start.sh
weicao: > In this file, there is a retry() function, which has an until loop in it. Also, a while true loop is just behind the code I submitted. So a global `exit_handler` is needed for this script file?
> 
> BTW: Because the retry function just validates the return code, but not checking the error msgs like the "is not empty", thus the retry() function was not used by my code.

I think it could be considered in later pr.

weicao (approved):

weicao:
Thanks for your contribution. 

BTW, it should also be fixed in main branch.

nayutah (approved):

Y-Rookie:
> Thanks for your contribution.
> 
> BTW, it should also be fixed in main branch.

I am refactoring the redis-related scripts and adding corresponding unit tests. I will fix this case in the main branch.

addons/redis/redis-cluster-scripts/redis-cluster-server-start.sh
leonliao: OK. Let's fix issues one by one.

leonliao:
> Thanks for your contribution.
> 
> BTW, it should also be fixed in main branch.

I found this issue on 0.9.0, and I tried in main branch, everything is so different, including CRD's schema has been changed. Due to time limit, I could not afford to apply to main branch and test everything again. So please cherry-pick the commit back to main branch.

PR #910, opened 2024-08-09
chore: support tidb 6.5

cjc7373:
fixes apecloud/kubeblocks#7942, fixes apecloud/kubeblocks-addons#895


addons/tidb/templates/componentdefinition-tidb.yaml
```diff
@@ -23,11 +37,10 @@ spec:
           - -c
           - |
             echo "start tidb..."
-            DOMAIN=$KB_NAMESPACE".svc{{ .Values.clusterDomain }}"
             exec /tidb-server --store=tikv \
             --advertise-address=${KB_POD_FQDN}{{ .Values.clusterDomain }} \
             --host=0.0.0.0 \
-            --path=${KB_CLUSTER_NAME}-tidb-pd.${DOMAIN}:2379 \
+            --path=${PD_ADDRESS}:2379 \
```
weicao, line 43: it seems PD_ADDRESS already contains the port ?

addons/tidb/scripts/pd_start.sh
```diff
@@ -29,6 +29,12 @@ if [[ -f $DATA_DIR/join ]]; then
 elif [[ ! -d $DATA_DIR/member/wal ]]; then
     echo "first started pod"
     replicas=$(echo "${KB_POD_LIST}" | tr ',' '\n')
+    # FIXME: Relying on leader status to determine whether to join or initialize a cluster
```
nayutah, line 32: KB can render a list of envs KB_CLUSTER_OPERATION/KB_COMPONENT_OPERATION/KB_POD_OPERATION to Pod to tell the current operation. It helps to identify the operation types and phase of a cluster/component/pod.

nayutah: When the Pod is doing another operation, KB rerender it to another operation type.

weicao: There is no such env.

weicao: And environment variables set through envFrom do not automatically update when the ConfigMap is modified. Therefore, unless the Pod is restarted, the env variables will not change dynamically.

addons/tidb/templates/componentdefinition-tidb.yaml
cjc7373: indeed

addons/tidb/scripts/pd_start.sh
cjc7373: What I need here is to distinguish whether the pod is started in cluster initialization or in cluster hscale process. Because these two circumstances will map to different start args.

nayutah: > There is no such env.

KB can import such envs to tell if pod is in creating or in hscale/restart/... process, the operator will know it, so it can render it, so the scripts will get the operation context.

nayutah (dismissed):

zjx20 (approved):

weicao:
It looks th expression in vars reads concisely !

weicao (approved):

PR #920, opened 2024-08-12
feat: update minio addon

fengluodb:
```yaml
apiVersion: apps.kubeblocks.io/v1alpha1
kind: Cluster
metadata:
  name: test-minio
  namespace: default
spec:
  componentSpecs:
  - componentDef: minio
    env:
    - name: MINIO_BUCKETS
      value: "kb-log,kb-backup"
    name: minio
    replicas: 2
    resources:
      limits:
        cpu: "1"
        memory: 1Gi
      requests:
        cpu: 100m
        memory: 102Mi
    serviceVersion: 1.0.0-alpha.0
    volumeClaimTemplates:
    - name: data
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 20Gi
  resources:
    cpu: "0"
    memory: "0"
  storage:
    size: "0"
  terminationPolicy: Delete
```


ldming (approved):

addons/minio/templates/cpmd.yaml
```diff
@@ -0,0 +1,166 @@
...
+          IFS=',' set -- $cur
+          array="$@"
+          last=$(echo $array | awk '{print $NF}')
+          if [[ "$last" == "$KB_COMP_REPLICAS" ]]; then
+            new="[$cur]"
+          else
+            new="[$cur,$KB_COMP_REPLICAS]"
+          fi
+        fi
+        kubectl patch configmap "$name" -n "$namespace" --type strategic -p "{\"data\":{\"$key\":\"$new\"}}"
```
nayutah, line 90: It is very hacky to manipulate a configmap in init container, if the Pod is restricted by RBAC to write against a configmap, this command will fail. Try to use vars and templates to render the configmap.

fengluodb: For MinIO expansion, I need to concatenate commands using the history of replicas. Therefore, I've recorded this information in ConfigMaps. And now, the service account (SA) created by default for the cluster already has the necessary permissions. However, it would be better if there were other methods to record this information.

PR #927, opened 2024-08-14
Chore/kbcli support etcd

loomts:
add values.schema.json to add description to kbcli


addons-cluster/etcd/values.schema.json
```diff
@@ -0,0 +1,160 @@
...
+      "description": "Override the name of the chart",
+      "type": "string",
+      "default": ""
+    },
+    "fullnameOverride": {
+      "title": "Full Name Override",
+      "description": "Override the full name of the chart",
+      "type": "string",
+      "default": ""
+    },
```
ldming, line 16: I don't think these name fields should be set by user, it can be set by the helm release name that is enough.

addons-cluster/etcd/values.schema.json
```diff
@@ -0,0 +1,160 @@
...
+        },
+        "annotations": {
+          "title": "Client Service Annotations",
+          "description": "Optional annotations for Cloud Provider if clientService type is LoadBalancer",
+          "type": "object",
+          "default": {}
+        }
+      }
+    },
+    "resources": {
```
ldming, line 115: Better to refer other addon cluster chart, set its cpu/memory and storage.

loomts: Since resources.limits[requests].cpu[mem, storage] = 0 means that no limit, so I use 0 as default.

addons-cluster/etcd/values.schema.json
```diff
@@ -0,0 +1,216 @@
...
+        "type": "string"
+      },
+      "default": ["kubernetes.io/hostname"]
+    },
+    "affinity": {
+      "title": "Affinity",
+      "description": "Affinity setting for etcd cluster pods assignment",
+      "type": "object",
+      "default": {}
+    },
```
weicao, line 28: too much, if users want to customize these properties, they may write a Cluster yaml.

addons-cluster/etcd/values.schema.json
weicao: I think @ldming means the format of etcd/values.shcema.json should follow other addons' values.shcema.json

weicao: so that the parameters of `kbcli cluster create etcd` can be similar to other addons.
e.g. 
kbcli cluster create etcd --cpu=1 --memory=2 --storage=30

loomts: fixed

addons-cluster/etcd/templates/cluster.yaml
```diff
@@ -15,10 +15,10 @@ spec:
   {{- end }}
   componentSpecs:
     - name: etcd
-      {{- if .Values.peerService.name }}
+      {{- if .Values.peerService.enabled }}
       services:
-        - name: {{ .Values.peerService.name }}
-          serviceType: {{ .Values.peerService.type }}
+        - name: peer
+          serviceType: LoadBalancer
```
shanshanying, line 21: why LB is the only choice ?

loomts: cause default peerService in etcd is headless svc, and we can specify LoadBalancer svc to introduce ip addrs to enable multi-cluster communication.
so this field is only used for multi-cluster etcd, and I haven't test for NodePort svc for multi-cluster, not sure if it can run well.

loomts: sounds reasonable, but NodePort service is not supported for multi-cluster yet, here is [related issue](https://github.com/apecloud/kubeblocks/issues/7916).

loomts: anyway, I make it configurable for future implementation

loomts:
see #941

PR #955, opened 2024-08-19
chore: zookeeper support version 3.4

kubeJocker:


addons-cluster/zookeeper/templates/_helpers.tpl
```diff
@@ -0,0 +1,12 @@
+{{- define "zookeeper-cluster.extra-envs" }}
```
shanshanying, line 1: use `cluster.spec.componentSpecs.env`  instead, to inject envs for Cluster.

kubeJocker: done

addons-cluster/zookeeper/values.yaml
```diff
@@ -2,6 +2,10 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+## @param version cluster version
+##
+version: 3.9.2
```
Y-Rookie, line 7: better to rename to `serviceVersion`

addons/zookeeper/scripts/startup.sh
```diff
@@ -0,0 +1,29 @@
...
+# Execute entrypoint as usual after obtaining ZOO_SERVER_ID
+# check ZOO_SERVER_ID in persistent volume via myid
+# if not present, set based on POD hostname
+# Execute entrypoint as usual after obtaining ZOO_SERVER_ID
+# check ZOO_SERVER_ID in persistent volume via myid
+# if not present, set based on POD hostname
+if [[ -f "/bitnami/zookeeper/data/myid" ]]; then
+  export ZOO_SERVER_ID="$(cat /bitnami/zookeeper/data/myid)"
+else
+  SERVICE_ID=$((${KB_POD_NAME##*-}))
```
Y-Rookie, line 12: Does SERVICE_ID necessarily need to be of type int? Obtaining SERVICE_ID in this way may cause conflicts in scenarios involving instanceTemplate.

Y-Rookie: Furthermore, try to avoid using any built-in env with the prefix KB_. All env should be explicitly declared in vars or env. The built-in env with the prefix KB_ will be directly removed in subsequent 1.0 versions. If you need POD_NAME, you should use the native way to obtain it. For example:

```
          - name: CURRENT_POD_NAME
            valueFrom:
              fieldRef:
                apiVersion: v1
                fieldPath: metadata.name
```

addons/zookeeper/templates/cmpd.yaml
```diff
@@ -63,15 +52,8 @@ spec:
...
-
-            isSSL="false"
-            clientPortAddress="localhost"
-            clientPort=2181
-            GREP=grep
-            STAT=`"$JAVA" "-Dzookeeper.log.dir=${ZOO_LOG_DIR}" "-Dzookeeper.log.file=${ZOO_LOG_FILE}" \
-            -cp "$CLASSPATH" $CLIENT_JVMFLAGS $JVMFLAGS org.apache.zookeeper.client.FourLetterWordMain \
-            $clientPortAddress $clientPort srvr $isSSL 2> /dev/null    \
-            | $GREP Mode`
+            STAT=$(java -cp "$CLASSPATH" $CLIENT_JVMFLAGS $JVMFLAGS org.apache.zookeeper.client.FourLetterWordMain \
```
Y-Rookie, line 55: Please synchronize with the latest version and API from the main branch. In the new API, `builtinHandler: custom` is no longer available. Please use the kbagent approach for integration.

addons/zookeeper/scripts/startup.sh
```diff
@@ -0,0 +1,29 @@
...
+  echo $ZOO_SERVER_ID > /bitnami/zookeeper/data/myid
+fi
+
+function version_gt() { test "$(echo "$@" | tr " " "\n" | sort -V | head -n 1)" != "$1"; }
+function version_le() { test "$(echo "$@" | tr " " "\n" | sort -V | head -n 1)" == "$1"; }
+function version_lt() { test "$(echo "$@" | tr " " "\n" | sort -rV | head -n 1)" != "$1"; }
+function version_ge() { test "$(echo "$@" | tr " " "\n" | sort -rV | head -n 1)" == "$1"; }
+
+
+if  version_gt "3.6.0" "${ZOOKEEPER_IMAGE_VERSION%%-*}"  ; then
```
Y-Rookie, line 23: Consider the scenario where ZOOKEEPER_IMAGE_VERSION is empty. This env variable doesn't exist in legacy cluster.

addons/zookeeper/scripts/startup.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
Y-Rookie, line 1: better to add UT for script

addons/zookeeper/templates/cmpv.yaml
```diff
@@ -7,13 +7,18 @@ metadata:
...
         - 3.7.2
         - 3.8.4
         - 3.9.2
       compDefs:
         - zookeeper
   releases:
+    - name: 3.4.14
+      serviceVersion: 3.4.14
+      images:
+        zookeeper: {{ .Values.images.registry | default "docker.io" }}/{{ .Values.images.repository }}:3.4.14
```
Y-Rookie, line 21: Synchronize with the main branch and keep the style consistent.

kubeJocker: just rebase main

addons/zookeeper/scripts/startup.sh
kubeJocker: done

addons-cluster/zookeeper/values.yaml
kubeJocker: renamed

kubeJocker: apecloud use .Values.version  to decide which version to use. Better to keep this name

Y-Rookie (approved):

ldming (approved):

PR #961, opened 2024-08-19
feat: support redis(standalone replica) pitr

Chiwency:
We leverage the AOF backup feature from Redis version 7.0 onwards to implement Point-In-Time Recovery (PITR). By enabling the `aof-timestamp-enabled yes` and `aof-disable-auto-gc no` parameters, we activate AOF timestamp annotations while disabling the automatic cleanup of historical AOF files. A StatefulSet, which is continuously responsible for the ongoing backup process, manages and tracks the AOF files.

The continuous backup process works as follows:
1. Historical AOF files in the data directory are compressed, packaged, and pushed to the backup repository.
2. The latest AOF file is backed up to the repository directory and continuously updated. When Redis triggers AOF rewriting, the file is compressed and archived as described in step 1.
3. The backup files in the repository are named using the format `${base_file_ctime}.${seq}.${suffix}`, where `base_file_ctime` is the creation time of the base aof file, marking the start of this particular backup package. This timestamp creates a continuous timeline for all backups. The `seq` is a sequence number used for recovery in case of a backup pod failure.

The recovery process works as follows:
1. Based on `DP_RESTORE_TIME`, the backup package with the `base_file_ctime` closest to `DP_RESTORE_TIME` is selected. This base file's data is used as the starting point. Then, AOF files are processed according to their timestamp annotations , up until `DP_RESTORE_TIME` is reached.

### DIFF
--- a/addons/redis/config/redis7-config.tpl
+++ b/addons/redis/config/redis7-config.tpl
@@ -43,10 +43,9 @@ appendfsync everysec
 no-appendfsync-on-rewrite no
 auto-aof-rewrite-percentage 100
 auto-aof-rewrite-min-size 64mb
-aof-disable-auto-gc no
 aof-load-truncated yes
 aof-use-rdb-preamble yes
-aof-timestamp-enabled yes
+aof-timestamp-enabled no
 slowlog-log-slower-than 10000
 slowlog-max-len 128
 latency-monitor-threshold 0
--- a/addons/redis/config/redis7-config.tpl
+++ b/addons/redis/config/redis7-config.tpl
@@ -43,9 +43,10 @@ appendfsync everysec
 no-appendfsync-on-rewrite no
 auto-aof-rewrite-percentage 100
 auto-aof-rewrite-min-size 64mb
+aof-disable-auto-gc no
 aof-load-truncated yes
 aof-use-rdb-preamble yes
-aof-timestamp-enabled no
+aof-timestamp-enabled yes
 slowlog-log-slower-than 10000
 slowlog-max-len 128
 latency-monitor-threshold 0
--- a/addons/redis/dataprotection/common-scripts.sh
+++ b/addons/redis/dataprotection/common-scripts.sh
@@ -1,0 +1,71 @@
+# log info file
+function DP_log() {
+  msg=$1
+  local curr_date=$(date -u '+%Y-%m-%d %H:%M:%S')
+  echo "${curr_date} INFO: $msg"
+}
+
+# log error info
+function DP_error_log() {
+  msg=$1
+  local curr_date=$(date -u '+%Y-%m-%d %H:%M:%S')
+  echo "${curr_date} ERROR: $msg"
+}
+
+# Get file names without extensions based on the incoming file path
+function DP_get_file_name_without_ext() {
+  local fileName=$1
+  local file_without_ext=${fileName%.*}
+  echo $(basename ${file_without_ext})
+}
+
+# Save backup status info file for syncing progress.
+# timeFormat: %Y-%m-%dT%H:%M:%SZ
+# receive timestamp
+function DP_save_backup_status_info() {
+  local totalSize=$1
+  local startTime=$(date -u -d @$2 +%Y-%m-%dT%H:%M:%SZ)
+  local stopTime=$(date -u -d @$3 +%Y-%m-%dT%H:%M:%SZ)
+  local timeZone=$4
+  local extras=$5
+  local timeZoneStr=""
+  if [ ! -z ${timeZone} ]; then
+    timeZoneStr=",\"timeZone\":\"${timeZone}\""
+  fi
+  if [ -z "${stopTime}" ]; then
+    echo "{\"totalSize\":\"${totalSize}\"}" >${DP_BACKUP_INFO_FILE}
+  elif [ -z "${startTime}" ]; then
+    echo "{\"totalSize\":\"${totalSize}\",\"extras\":[${extras}],\"timeRange\":{\"end\":\"${stopTime}\"${timeZoneStr}}}" >${DP_BACKUP_INFO_FILE}
+  else
+    echo "{\"totalSize\":\"${totalSize}\",\"extras\":[${extras}],\"timeRange\":{\"start\":\"${startTime}\",\"end\":\"${stopTime}\"${timeZoneStr}}}" >${DP_BACKUP_INFO_FILE}
+  fi
+}
+
+function DP_pull_directory() {
+  local dir_path="$1"
+  local local_path="$2"
+
+  while IFS= read -r filename; do
+      datasafed pull "$filename" "${local_path}/${filename#*/}"
+  done < <(datasafed list -r -f "$dir_path")
+}
+
+# Clean up expired logfiles.
+# Default interval is 60s
+# Default rootPath is /
+function DP_purge_expired_files() {
+  local currentUnix="${1:?missing current unix}"
+  local last_purge_time="${2:?missing last_purge_time}"
+  local root_path=${3:-"/"}
+  local interval_seconds=${4:-60}
+  local diff_time=$((${currentUnix} - ${last_purge_time}))
+  if [[ -z ${DP_TTL_SECONDS} || ${diff_time} -lt ${interval_seconds} ]]; then
+    return
+  fi
+  expiredUnix=$((${currentUnix} - ${DP_TTL_SECONDS}))
+  files=$(datasafed list -f --recursive --older-than ${expiredUnix} ${root_path})
+  for file in ${files[@]}; do
+    datasafed rm ${file}
+    echo ${file}
+  done
+}
--- a/addons/redis/dataprotection/pitr-backup.sh
+++ b/addons/redis/dataprotection/pitr-backup.sh
@@ -39,6 +39,7 @@ function aof_base_file() {
 
 function get_backup_files_prefix() {
   local base_file=${1}
+  # use the creation time of the base file as the start time
   echo "$(stat -c %Y "$base_file")".${global_backup_seq}
 }
 
@@ -127,7 +128,7 @@ function save_backup_status() {
   fi
   global_old_size=${total_size}
   local start_time=$(datasafed list / | awk -F '.' '{print $1}' | sort | head -n 1)
-  DP_save_backup_status_info "${total_size}" "${start_time}" "${global_aof_last_modify_time}"
+  DP_save_backup_status_info "${total_size}" "${start_time}" "$(date +%s)"
 }
 
 trap "echo 'Terminating...' && sync && ${connect_url} CONFIG SET aof-disable-auto-gc no && exit 0" TERM
--- a/addons/redis/dataprotection/pitr-backup.sh
+++ b/addons/redis/dataprotection/pitr-backup.sh
@@ -1,0 +1,150 @@
+export PATH="$PATH:$DP_DATASAFED_BIN_PATH"
+export DATASAFED_BACKEND_BASE_PATH="$DP_BACKUP_BASE_PATH"
+
+connect_url="redis-cli -h ${DP_DB_HOST} -p ${DP_DB_PORT} -a ${DP_DB_PASSWORD}"
+global_last_purge_time=$(date +%s)
+global_aof_last_modify_time=0
+global_old_size=0
+
+AOF_DIR=$(${connect_url} CONFIG GET appenddirname | awk 'NR==2')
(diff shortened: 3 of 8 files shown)


addons/redis/dataprotection/common-scripts.sh
```diff
@@ -0,0 +1,71 @@
...
+  local interval_seconds=${4:-60}
+  local diff_time=$((${currentUnix} - ${last_purge_time}))
+  if [[ -z ${DP_TTL_SECONDS} || ${diff_time} -lt ${interval_seconds} ]]; then
+    return
+  fi
+  expiredUnix=$((${currentUnix} - ${DP_TTL_SECONDS}))
+  files=$(datasafed list -f --recursive --older-than ${expiredUnix} ${root_path})
+  for file in ${files[@]}; do
+    datasafed rm ${file}
+    echo ${file}
```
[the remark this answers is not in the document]
wangyelei: it's a return value. and already logs in the out function

addons/redis/dataprotection/pitr-backup.sh
```diff
@@ -0,0 +1,151 @@
...
+  DP_save_backup_status_info "${total_size}" "${start_time}" "$(date +%s)"
+}
+
+trap "echo 'Terminating...' && sync && ${connect_url} CONFIG SET aof-disable-auto-gc no && exit 0" TERM
+echo "INFO: start to backup"
+set_conf
+while true; do
+  aof_seq=$(awk '/type i/ { print $4 }' "${AOF_MANIFEST_FILE}")
+  while [ "${global_backup_seq}" -lt "${aof_seq}" ]; do
+    archive_pair_files
```
wangyelei, line 140: There are multiple instances in the  replication mode, and the AOF files of each instance are inconsistent. It is necessary to archive and backup each pod

addons/redis/dataprotection/pitr-backup.sh
```diff
@@ -0,0 +1,151 @@
...
+  fi
+  global_old_size=${total_size}
+  local start_time=$(datasafed list / | awk -F '.' '{print $1}' | sort | head -n 1)
+  DP_save_backup_status_info "${total_size}" "${start_time}" "$(date +%s)"
+}
+
+trap "echo 'Terminating...' && sync && ${connect_url} CONFIG SET aof-disable-auto-gc no && exit 0" TERM
+echo "INFO: start to backup"
+set_conf
+while true; do
```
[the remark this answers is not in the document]
Chiwency: ```trap "echo 'Terminating...' && sync && ${connect_url} CONFIG SET aof-disable-auto-gc no && exit 0" TERM```
sorry, a little doubt,  doesn't this line take effect?

addons/redis/templates/backupactionset.yaml
```diff
@@ -76,4 +76,49 @@ spec:
...
+      runOnTargetPodNode: true
+      syncProgress:
+        enabled: true
+        intervalSeconds: 5
+      command:
+      - bash
+      - -c
+      - |
+        #!/bin/bash
+        {{- .Files.Get "dataprotection/common-scripts.sh" | nindent 8 }}
```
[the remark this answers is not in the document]
Chiwency: image
In my PC, 8 is ok and the generated content is as expected

image

nayutah (approved):

wangyelei (approved):

ldming:
@Y-Rookie could you take a look?

addons/redis/config/redis7-config.tpl
```diff
@@ -43,9 +43,10 @@ appendfsync everysec
 no-appendfsync-on-rewrite no
 auto-aof-rewrite-percentage 100
 auto-aof-rewrite-min-size 64mb
+aof-disable-auto-gc no
 aof-load-truncated yes
 aof-use-rdb-preamble yes
-aof-timestamp-enabled no
+aof-timestamp-enabled yes
```
ldming, line 49: If upgrade, will these parameter change affect the existing cluster?

Chiwency: Only `aof-timestamp-enabled yes` will take effect. In the later aof log, each command will be attached with a timestamp annotation, is transparent to the user.

addons/redis/config/redis7-config.tpl
```diff
@@ -43,9 +43,10 @@ appendfsync everysec
 no-appendfsync-on-rewrite no
 auto-aof-rewrite-percentage 100
 auto-aof-rewrite-min-size 64mb
+aof-disable-auto-gc no
```
Y-Rookie, line 46: What is the default value of this parameter? And what is its purpose? I couldn't find it in the official redis.conf file. This issue: https://github.com/redis/redis/issues/10561 mentions that it is just a testing parameter that exists in the internal code.

shanshanying:
Hi @Chiwency ,

after some discussion with my colleges. we suggested:
1. add new parameters to configConstraint (are they static/dynamic/immutble parameters) to make sure they are updated in the expected way.
2. do not modify existing parameters' default values. otherwise, other users,  who are using redis now, will be affected.
3. instead, write a brief description on hwo to trigger a PITR. e.g  create cluster -> reconfiguration (to update parametes) -> pitr.

addons/redis/config/redis7-config.tpl
Chiwency: the default value is no, I mentioned this parameter just to clarify that I used this parameter. I actually set the parameter to no in the backup target pod only while the users trigger the PITR.  As I see, using the testing parameter doesn't affect the existing cluster and the users who don't need PITR.

Chiwency:
> Hi @Chiwency ,
> 
> after some discussion with my colleges. we suggested:
> 
> 1. add new parameters to configConstraint (are they static/dynamic/immutble parameters) to make sure they are updated in the expected way.
> 2. do not modify existing parameters' default values. otherwise, other users,  who are using redis now, will be affected.
> 3. instead, write a brief description on hwo to trigger a PITR. e.g  create cluster -> reconfiguration (to update parametes) -> pitr.

@shanshanying I get it. 
1. And I will revert the `aof-timestamp-enabled`  to default value `false` 
2. I have a question about point  2, to implement PIRT which is unsupported in Redis, for users who open PITR's continuous backup, the backup script will change the two parameters `aof-disable-auto-gc` and `aof-timestamp-enabled` to `yes` for target redis instance, is OK?  It won't affect the users who don's use PITR and existing cluster.
3. For point 3, the two parameters just take effect in the target one instance, users don't know which instance is selected to backup, so I prefer to set them by script.

shanshanying:
> > Hi @Chiwency ,
> > after some discussion with my colleges. we suggested:
> > 
> > 1. add new parameters to configConstraint (are they static/dynamic/immutble parameters) to make sure they are updated in the expected way.
> > 2. do not modify existing parameters' default values. otherwise, other users,  who are using redis now, will be affected.
> > 3. instead, write a brief description on hwo to trigger a PITR. e.g  create cluster -> reconfiguration (to update parametes) -> pitr.
> 
> @shanshanying I get it.
> 
> 1. And I will revert the `aof-timestamp-enabled`  to default value `no`
> 2. I have a question about point  2, to implement PIRT which is unsupported in Redis, for users who open PITR's continuous backup, the backup script will change the two parameters `aof-disable-auto-gc` and `aof-timestamp-enabled` to `yes` for target redis instance, is OK?  It won't affect the users who don's use PITR and existing cluster.
> 3. For point 3, the two parameters just take effect in the target one instance, users don't know which instance is selected to backup, so I prefer to set them by script.

Hi @Chiwency,
> 2. I have a question about point  2...
The point is: if the change of parameters will affect exsitng RUNNING redis clusters, we'd better change is in a more EXPLICIT way: tell users how to modify the values mannuly, either by OpsRequest, or edit some configmap, and write a doc to explain the side-effect. 
I think we can update parameters through our "Reconfig" OpsRequest. Isn't it?

Chiwency:
> The point is: if the change of parameters will affect exsitng RUNNING redis clusters, we'd better change is in a more EXPLICIT way: tell users how to modify the values mannuly, either by OpsRequest, or edit some configmap, and write a doc to explain the side-effect. I think we can update parameters through our "Reconfig" OpsRequest. Isn't it?

Hi @shanshanying 
How can I understand the meaning of "affect exsitng RUNNING redis clusters"? As I revert the config file changes, users who don't trigger PITR backup will not be affected, and there will be no any change. But when users apply PITR to a exsiting RUNNING redis clusters, there are two scenarios at this point:
1. **IMPLICT parameter modification**
User trigger PITR by only one command, then the script will change the value of `aof-timestamp-enabled` and `aof-disable-auto-gc`. These two parameters are only applied to the AOF logs. From the user's perspective, the parameter changes will not be noticeable.
```bash
kbcli cluster update <redis-cluster-name> --backup-enabled=true --backup-method=aof
```
2. **EXPLICIT parameter modification**
User trigger PITR by "Reconfig" the  `aof-timestamp-enabled yes` first, then trigger PITR backup.
```bash
kbcli cluster edit-config <redis-cluster-name>
kbcli cluster update <redis-cluster-name> --backup-enabled=true --backup-method=aof
```
  Whether it's Option 1 or Option 2, documentation will be provided later to explain the changes. My key question here is whether these parameters should be managed by the users,  do these parameter modifications count as affecting the existing cluster? 
  As I see, "update parameters through our "Reconfig" OpsRequest" may be an additional burden for the users, including understanding and operating. And PITR feature relies on these two parameters, so the two command's in Option 2 actually an atomic command.
What is your opinion on this?

nayutah:
> > The point is: if the change of parameters will affect exsitng RUNNING redis clusters, we'd better change is in a more EXPLICIT way: tell users how to modify the values mannuly, either by OpsRequest, or edit some configmap, and write a doc to explain the side-effect. I think we can update parameters through our "Reconfig" OpsRequest. Isn't it?
> 
> Hi @shanshanying How can I understand the meaning of "affect exsitng RUNNING redis clusters"? As I revert the config file changes, users who don't trigger PITR backup will not be affected, and there will be no any change. But when users apply PITR to a exsiting RUNNING redis clusters, there are two scenarios at this point:
> 
> 1. **IMPLICT parameter modification**
>    User trigger PITR by only one command, then the script will change the value of `aof-timestamp-enabled` and `aof-disable-auto-gc`. These two parameters are only applied to the AOF logs. From the user's perspective, the parameter changes will not be noticeable.
> 
> ```shell
> kbcli cluster update <redis-cluster-name> --backup-enabled=true --backup-method=aof
> ```
> 
> 2. **EXPLICIT parameter modification**
>    User trigger PITR by "Reconfig" the  `aof-timestamp-enabled yes` first, then trigger PITR backup.
> 
> ```shell
> kbcli cluster edit-config <redis-cluster-name>
> kbcli cluster update <redis-cluster-name> --backup-enabled=true --backup-method=aof
> ```
> 
> Whether it's Option 1 or Option 2, documentation will be provided later to explain the changes. My key question here is whether these parameters should be managed by the users, do these parameter modifications count as affecting the existing cluster? As I see, "update parameters through our "Reconfig" OpsRequest" may be an additional burden for the users, including understanding and operating. And PITR feature relies on these two parameters, so the two command's in Option 2 actually an atomic command. What is your opinion on this?

two independent operations is better for open-source users, first to set the flag, and then use the PITR, if one use PITR with aof timestamp disabled, please return an error code and exits immediately.

addons/redis/dataprotection/pitr-backup.sh
```diff
@@ -0,0 +1,200 @@
...
+  global_old_size=${total_size}
+  local start_time=$(datasafed list / | awk -F '.' '{print $1}' | sort | head -n 1)
+  if [ -z "$start_time" ]; then
+    start_time=$(date +%s)
+    DP_log "save_backup_status: empty start_time from backup repo, use current time"
+  fi
+  DP_save_backup_status_info "${total_size}" "${start_time}" "$(date +%s)"
+}
+
+function check_conf() {
```
nayutah, line 168: Also check the aof-timestamp-enabled here, it can fail and return an error when both either of them is disabled. Prompt the user to manually edit the config items with 'kbcli edit-config' or  'kubectl + reconfigure ops'

Chiwency:
> two independent operations is better for open-source users, first to set the flag, and then use the PITR, if one use PITR with aof timestamp disabled, please return an error code and exits immediately.

Hi @nayutah  
As kbcli doesn't offer complex logic validation releate querying the API server, it relies on validation by the manager after generating the opsReq. Therefore, we cannot return error codes immediately in kbcli. Instead, errors are detected through cluster status queries. The same applies to the restore-in-time parameter. The time validation process is placed manager controller. I think enhancing the validation capabilities of kbcli is a future optimization direction.

Issue #967, opened 2024-08-21
[BUG] PITR Not Work for PostgreSQL AddOn


dingshun-cmss:
**Describe the bug**
The PITR feature for PG doesn't work

**To Reproduce**
1. Prepare for the full backup and WAL archive for the source DB
2. Create an `OpsRequest` , with some contents like this:
```yaml
apiVersion: apps.kubeblocks.io/v1alpha1
kind: OpsRequest
metadata:
  generateName: dstestrestore-
spec:
  clusterRef: dstestpg-restore2
  restoreSpec:
    backupName: backup-dbaas-dstestpg-20240821073913
    volumeRestorePolicy: Parallel
    restoreTimeStr: "2024-08-21T07:44:20Z"
  ttlSecondsAfterSucceed: 30
  ttlSecondsBeforeAbort: 0
  type: Restore
```

The restored DB cannot start successfully.  

When we check the logs for the failed DB contaier, we can find the following line:
```
2024-08-21 07:49:37 GMT [93]: [1-1] 66c59c11.5d 0     LOG:  invalid value for parameter "recovery_target_time": "2024-08-21T07:44:20Z"
```

Here's the restore part in the generated patroni config:
```
  ...
  kb_restore_from_time:
    command: bash /home/postgres/pgdata/kb_restore/kb_restore.sh
    keep_existing_recovery_conf: false
    recovery_conf:
      recovery_target_action: promote
      recovery_target_time: '2024-08-21T07:44:20Z'
      recovery_target_timeline: latest
      restore_command: envdir /home/postgres/pgdata/wal-g/restore-env /home/postgres/pgdata/wal-g/wal-g
        wal-fetch %f %p >> /home/postgres/pgdata/kb_restore/wal-g.log 2>&1
  ...
```

It looks like that PostgreSQL doesn't recognize the RFC-3339 format for restore timestamp.

Then I tried to change the `restoreTimeStr` to a format that is recognized by PostgreSQL: `2024-08-21 07:44:20+00` ,  and recreate the `OpsRequest` .  However, the Pods cannot be created.  There're some errors in the KB operator log: 
```
[...]

PR #970, opened 2024-08-21
fix: postgresql pitr not work

dingshun-cmss:
- The problem is that in OpsRequest, the timestamp should be in RFC-3339 format. But PostgreSQL doesn't recognize RFC-3339 format. So we need to do a format conversion
- Fixed both in pg PITR restore & wal-g PITR restore
- Optimize the code style when saving multi-line to a file

Fix: #967

### DIFF
--- a/addons/postgresql/dataprotection/postgresql-pitr-restore.sh
+++ b/addons/postgresql/dataprotection/postgresql-pitr-restore.sh
@@ -26,10 +26,12 @@ echo "#!/bin/bash" > ${RESTORE_SCRIPT_DIR}/kb_restore.sh;
 echo "[[ -d '${DATA_DIR}.old' ]] && mv -f ${DATA_DIR}.old/* ${DATA_DIR}/;" >> ${RESTORE_SCRIPT_DIR}/kb_restore.sh;
 echo "sync;" >> ${RESTORE_SCRIPT_DIR}/kb_restore.sh;
 chmod +x ${RESTORE_SCRIPT_DIR}/kb_restore.sh;
-echo "restore_command='case "%f" in *history) cp ${PITR_DIR}/%f %p ;; *) mv ${PITR_DIR}/%f %p ;; esac'" > ${CONF_DIR}/recovery.conf;
-echo "recovery_target_time='${DP_RESTORE_TIME}'" >> ${CONF_DIR}/recovery.conf;
-echo "recovery_target_action='promote'" >> ${CONF_DIR}/recovery.conf;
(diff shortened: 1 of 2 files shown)


wangyelei (approved):

wangyelei:
what's your PGSQL version?

wangyelei (approved):

addons/postgresql/dataprotection/postgresql-pitr-restore.sh
```diff
@@ -26,10 +26,12 @@ echo "#!/bin/bash" > ${RESTORE_SCRIPT_DIR}/kb_restore.sh;
 echo "[[ -d '${DATA_DIR}.old' ]] && mv -f ${DATA_DIR}.old/* ${DATA_DIR}/;" >> ${RESTORE_SCRIPT_DIR}/kb_restore.sh;
 echo "sync;" >> ${RESTORE_SCRIPT_DIR}/kb_restore.sh;
 chmod +x ${RESTORE_SCRIPT_DIR}/kb_restore.sh;
-echo "restore_command='case "%f" in *history) cp ${PITR_DIR}/%f %p ;; *) mv ${PITR_DIR}/%f %p ;; esac'" > ${CONF_DIR}/recovery.conf;
-echo "recovery_target_time='${DP_RESTORE_TIME}'" >> ${CONF_DIR}/recovery.conf;
-echo "recovery_target_action='promote'" >> ${CONF_DIR}/recovery.conf;
-echo "recovery_target_timeline='latest'" >> ${CONF_DIR}/recovery.conf;
+cat << EOF > "${CONF_DIR}/recovery.conf"
+restore_command='case "%f" in *history) cp ${PITR_DIR}/%f %p ;; *) mv ${PITR_DIR}/%f %p ;; esac'
+recovery_target_time='$( date -d "@${DP_RESTORE_TIMESTAMP}" '+%F %T%::z' )'
```
wangyelei, line 31: PGSQL uses ISO-8601 international standard, but seems it is not a ISO-8601.  replace with` date -u -d @"${DP_RESTORE_TIMESTAMP}" +"%Y-%m-%dT%H:%M:%SZ"`?

dingshun-cmss: From the [PG documentation](https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-RECOVERY-TARGET-TIME): 
image

And the timestamp format is described [here](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-INPUT-TIME-STAMPS)
image

So I just follow the PG documentation to generate the timestamp string

nayutah (approved):

PR #990, opened 2024-08-29
fix: zookeeper reconfigure donot take effect

kubeJocker:


kubeJocker:
There is a mistake that all configurations can be dynamically changed. In fact, only the configurations related to the cluster members can be dynamically changed.

shanshanying (dismissed):

shanshanying (approved):

sophon-zt (approved):

PR #1005, opened 2024-09-04
fix: check the validity of configuration parameters

sophon-zt:


shanshanying:
by restricting the validation rules, it seems only parameters explicitly defined can be updated. (or an optional work-around is to edit corresponding configmap). my question is:
- what if user performs a minor version upgrading on a cluster. There might be new parameters in the new version. All updates on these new parameters will be rejected.
- with `reRenderResourceTypes` api , parameters will be rendered after HScale or VScale. And users' modifictions on this configmap might be lost

nayutah (approved):

shanshanying (approved):

shanshanying (approved):

zjx20 (approved):

PR #1006, opened 2024-09-04
fix: orchestrator roleProbe can not work in raft mode

PhantomMaa:
It can not work using the kubeblocks 0.9 version.


addons/orchestrator/templates/cmpd-raft.yaml
```diff
@@ -18,19 +18,21 @@ spec:
...
-            role=$(curl -s http://127.0.0.1:3000/api/leader-check)
-            if [[ $role == "\"OK\"" ]]; then
-              echo -n "primary"
-            elif [[ $role == "\"Not leader\"" ]]; then
-              echo -n "secondary"
-            else
-              echo -n ""
-            fi
+      builtinHandler: custom
+      customHandler:
```
nayutah, line 22: Add time settings for this action, like:
      periodSeconds: 1
      initialDelaySeconds: 60
      timeoutSeconds: 10

kubeJocker:
@PhantomMaa  thank you for your commit
You can check the version of Kubeblocks you are using by running 
`kbcli version`
 If you are using Kubeblocks version 0.9.X, you will need to use the release-0.9 branch of kubeblocks-addons.

addons/orchestrator/templates/cmpd-raft.yaml
nayutah: it is good to set them as they are best recommendations for orchestrator.

kubeJocker: @nayutah talk is cheap, show me the data

kubeJocker:
By the way, the MySQL in the current release-0.9 does not yet support registration with Orchestrator. If you wish to experience this process, you can install the latest alpha version of KubeBlocks, by
`kbcli kubeblocks list-versions --devel=true`

PR #1023, opened 2024-09-11
feat: redis-restore-by-keys

Chiwency:
# Support redis restore by multiply key patterns.

We can apply the key restoration by
```bash
kbcli cluster restore --backup "backup_name" --restore-keys  "a*,b*"
```
"a*" and "b*" represent the patterns of keys we want to restore, and split by comma.
for the details of the key pattern, refer to [Redis/KEYS](https://redis.io/docs/latest/commands/keys/)

## Implementation
1. We inject the key patterns into the restore pod as the environment variable `DP_RESTORE_KEY_PATTERNS`.
2. Let the prepareData job pull full data to a different directory.
3. Then in the postReady job, we start a local redis instance based on the full data,  and [SCAN](https://redis.io/docs/latest/commands/scan/) keys
by pattern then [MIGRATE](https://redis.io/docs/latest/commands/migrate/) the selected keys to the target redis instance.

### DIFF
--- a/addons/redis/dataprotection/pitr-restore.sh
+++ b/addons/redis/dataprotection/pitr-restore.sh
@@ -50,10 +50,11 @@ function get_files_to_restore() {
   esac
 }
 
+mkdir -p ${DATA_DIR}
 res=$(find ${DATA_DIR} -type f)
 data_protection_file=${DATA_DIR}/.kb-data-protection
-if [ ! -z "${res}" ] && [ ! -f ${data_protection_file} ]; then
-  echo "${DATA_DIR} is not empty! Please make sure that the directory is empty before restoring the backup."
+if [ -n "${res}" ] && [ ! -f ${data_protection_file} ]; then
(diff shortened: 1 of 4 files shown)


zjx20:
When a key migration fails, this script appears to ignore the error and continue running. This behavior seems unreliable. It might be better to add an environment variable to control whether errors are ignored (e.g., 
`DP_RESTORE_KEY_IGNORE_ERRORS`), with the default behavior being not to ignore errors.

Chiwency:
> When a key migration fails, this script appears to ignore the error and continue running. This behavior seems unreliable. It might be better to add an environment variable to control whether errors are ignored (e.g., `DP_RESTORE_KEY_IGNORE_ERRORS`), with the default behavior being not to ignore errors.

Get, plz check the new commit

wangyelei (approved):

zjx20 (approved):

PR #1096, opened 2024-10-15
support update clickhouse from 08->09

loomts:
resolve [#8271](https://github.com/apecloud/kubeblocks/issues/8271)


addons/clickhouse/values.yaml
```diff
@@ -19,3 +19,26 @@ image:
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
   tag: 24.3.10-debian-12-r1
+
+zookeeper:
```
shanshanying, line 23: set default to docker.io

PR #1097, opened 2024-10-16
chore: upgrade zk to v1.0 api, refactor script and add ut

Y-Rookie:


kubeJocker (dismissed):
LGTM

addons/zookeeper/templates/cmpd.yaml
```diff
@@ -43,27 +44,63 @@ spec:
...
+      value: {{ .Values.metrics.enabled | quote }}
+    - name: ZOOKEEPER_METRICS_PORT
+      value: {{ .Values.metrics.port | quote }}
+    - name: ZOOKEEPER_DATA_DIR
+      value: {{ .Values.zookeeper.dataDir }}
+    - name: ZOOKEEPER_DATA_LOG_DIR
+      value: {{ .Values.zookeeper.dataLogDir }}
+    - name: ZOOBINDIR
+      value: "/opt/bitnami/zookeeper/bin"
+    - name: ZOOBIN
```
zjx20, line 64: Is this var useful? Its value is the same as the upper one.

addons/zookeeper/templates/cmpd.yaml
```diff
@@ -43,27 +44,63 @@ spec:
...
+      value: "/opt/bitnami/zookeeper/bin"
+    - name: ZOO_ENABLE_AUTH
+      value: "yes"
+    - name: SERVICE_PORT
```
zjx20, line 68: This var seems useless.

kubeJocker: service_port needs to be preserved as Gemini will utilize this variable.

zjx20: please add a comment to the code.

addons/zookeeper/templates/cmpd.yaml
Y-Rookie: > Is this var useful? Its value is the same as the upper one.

It should be necessary. This addon uses the Bitnami image, and the Bitnami image internally has some dependent environment variables (env).

Y-Rookie: > > Is this var useful? Its value is the same as the upper one.
> 
> It should be necessary. This addon uses the Bitnami image, and the Bitnami image internally has some dependent environment variables (env).

@kubeJocker please take a look

kubeJocker: This var is used in the ZkServer.sh of version 3.4, so it is necessary.

zjx20: Got it, please add a comment as well.

addons/zookeeper/templates/cmpd.yaml
Y-Rookie: > please add a comment to the code.

done

addons/zookeeper/templates/cmpd.yaml
Y-Rookie: > Got it, please add a comment as well.

done

zjx20 (approved):

wangyelei (approved):

PR #1107, opened 2024-10-18
chore: adapt rabbitmq addon for API v1.0

xuriwuyun:


addons/rabbitmq/templates/_helpers.tpl
```diff
@@ -51,41 +51,27 @@ app.kubernetes.io/name: {{ include "rabbitmq.name" . }}
...
+{{- define "rabbitmq.cmpdName" -}}
+{{- if eq (len .Values.resourceNamePrefix) 0 -}}
+{{ include "rabbitmq.cmpdNamePrefix" . }}{{ .Chart.Version }}
 {{- else -}}
-    {{- range $enabledClusterVersions -}}
-        {{- if eq $cvName . -}}
-            {{- true -}}
-        {{- end -}}
-    {{- end -}}
+{{ .Values.resourceNamePrefix}}-{{ .Chart.Version }}
```
zjx20, line 75: How about `{{ .Values.resourceNamePrefix}}-rabbitmq-{{ .Chart.Version }}`?

addons/rabbitmq/templates/componentversion.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
 spec:
   compatibilityRules:
     - compDefs:
-        - {{ include "rabbitmq.componentDefName" . }}
+        - {{ include "rabbitmq.cmpdNamePrefix" . }}
```
zjx20, line 10: This pattern can't match the comp if `.Values.resourceNamePrefix` is not empty.

zjx20 (approved):

wangyelei (approved):

PR #1110, opened 2024-10-21
chore: upgrade victoria-metrics to v1.0 API

Y-Rookie:


sophon-zt (dismissed):

sophon-zt (approved):

addons/victoria-metrics/templates/cmpv.yaml
zjx20, line 1: Does this kind of resource need to set `helm.sh/resource-policy: keep`?

zjx20 (approved):

PR #1112, opened 2024-10-21
chore: support to define the vttablet sidecar in wescale

leon-inf:


addons/wescale/templates/cmpd-wescale.yaml
```diff
@@ -213,3 +116,43 @@ spec:
           port: 
             name: client
             option: Optional
+  sidecars:
+    - name: vtablet
+      selectors:
+        - apecloud-mysq
```
xuriwuyun, line 122: How to use selectors to specify which component should inject the sidecar?

PR #1147, opened 2024-11-01
chore: upgrade influxdb to v1.0 API

Y-Rookie:


addons/influxdb/values.yaml
```diff
@@ -9,7 +14,4 @@ image:
     tag: 2.7.4-debian-11-r0
```
zjx20, line 14: The tag is unused.

zjx20 (approved):

wangyelei (approved):

PR #1197, opened 2024-11-14
chore: remove rbac related definition

cjc7373:
also add patroni's policy rules

todo: rabbitmq seems to need some rbac policies


zjx20 (approved):
LGTM. Just curious, in which case should the rbacEnabled be set to true?

zjx20:
I suggest deleting it if it's useless.

ldming:
> It seems like rbacEnabled is used to "simulate" user defined rbac resources. @Y-Rookie What's this field designed for?

In earlier versions of KB, KB did not support the automatic creation of service accounts (SA) with specific roles for clusters. Therefore, they would be created in kbcli or helm charts. After KB started supporting this feature, in most cases, this parameter has become unnecessary.

But, to my knowledge, currently, Elasticsearch sets this parameter to true, referring to this PR https://github.com/apecloud/kbcli/pull/460.

ldming (dismissed):

zjx20 (dismissed):

leon-inf (approved):

PR #1204, opened 2024-11-15
feat: mongodb support host network with new api

xuriwuyun:


addons/mongodb/templates/componentdefinition.yaml
```diff
@@ -195,3 +193,28 @@ spec:
           name: root
           optional: false
           password: Required
+    - name: KB_SERVICE_PORT  
+      value: "27017"
```
zjx20, line 197: remove the value?

xuriwuyun: it is the default port for mongodb, if hostnetwork is not enabled, the env will use this value.

zjx20: got it

zjx20 (approved):

kizuna-lek (approved):

PR #1291, opened 2024-12-06
fix: TLS_ENABLED might not be defined

leon-inf:
resolve https://github.com/apecloud/kubeblocks/issues/8598

### DIFF
--- a/addons/clickhouse/configs/ch_keeper_00_default_overrides.xml.tpl
+++ b/addons/clickhouse/configs/ch_keeper_00_default_overrides.xml.tpl
@@ -2,7 +2,7 @@
 {{- $namespace := $.cluster.metadata.namespace }}
 <clickhouse>
   <listen_host>0.0.0.0</listen_host>
-  {{- if eq $.TLS_ENABLED "true" }}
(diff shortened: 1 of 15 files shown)


sophon-zt (dismissed):

wangyelei (dismissed):

addons/clickhouse/configs/ch_keeper_00_default_overrides.xml.tpl
```diff
@@ -30,14 +30,14 @@
           <raft_logs_level>warning</raft_logs_level>
       </coordination_settings>
       <raft_configuration>
-      {{- if eq $.TLS_ENABLED "true" }}
+      {{- if and (index . "TLS_ENABLED") (eq .TLS_ENABLED "true") }}
```
wangyelei, line 33: render failed with nest scope, define a var to resolve it

zjx20 (approved):

addons/clickhouse/configs/ch_keeper_00_default_overrides.xml.tpl
leon-inf: Fixed

PR #1337, opened 2024-12-18
feat: custom security context for kafka addon

### LINKED ISSUE #1336: [Improvement]support custom securityContext for kafka-addon
Restrict securityContext is good practice for container runtime, but in some scenario of user-defined images, user may need to switch user or even sudo. Maybe making securityContext can be configed by helm values will be more flexible.

lancelot1989:
resolves #1336


addons/kafka/templates/cmpd-broker.yaml
```diff
@@ -228,9 +230,10 @@ spec:
       - name: jmx-exporter
         image: {{ .Values.images.registry | default "docker.io" }}/{{ .Values.images.jmxExporter.repository }}:{{ .Values.images.jmxExporter.tag }}
         imagePullPolicy: {{ default "IfNotPresent" .Values.images.pullPolicy }}
+        {{- if .Values.container.securityContext }}
         securityContext:
-          runAsNonRoot: true
```
Y-Rookie, line 232: The modification logic here differs from the previous implementation. 

the default value of allowPrivilegeEscalation is true in k8s？

```
	// defaultAllowPrivilegeEscalation controls the default setting for whether a
	// process can gain more privileges than its parent process.
	// +optional
	DefaultAllowPrivilegeEscalation *bool `json:"defaultAllowPrivilegeEscalation,omitempty" protobuf:"varint,15,opt,name=defaultAllowPrivilegeEscalation"`
	// allowPrivilegeEscalation determines if a pod can request to allow
	// privilege escalation. If unspecified, defaults to true.
	// +optional
	AllowPrivilegeEscalation *bool `json:"allowPrivilegeEscalation,omitempty" protobuf:"varint,16,opt,name=allowPrivilegeEscalation"`
```

lancelot1989: Yes, so in the previous implementation, the jmx-exporter will be able to `sudo`. Now it can't.

If we jmx-exporter needs `AllowPrivilegeEscalation`, i can define another value for it.

zjx20:
@caiq1nyu please take a look.

caiq1nyu (approved):

addons/kafka/templates/cmpd-broker.yaml
lancelot1989: using separate value for exporter's securityContext

Y-Rookie (approved):

zjx20 (approved):

PR #1338, opened 2024-12-18
chore: support new switchover action

cjc7373:
ref: https://github.com/apecloud/kubeblocks/pull/8645


addons/apecloud-mysql/templates/_helpers.tpl
```diff
@@ -167,7 +167,16 @@ lifecycleActions:
         - /bin/sh
         - -c
         - |
-          /tools/syncerctl switchover --primary "$KB_LEADER_POD_NAME" ${KB_SWITCHOVER_CANDIDATE_NAME:+--candidate "$KB_SWITCHOVER_CANDIDATE_NAME"}
+          if [ -z "$KB_SWITCHOVER_ROLE" ]; then
+              echo "role can't be empty"
+              exit 1
```
leon-inf, line 172: The KB_SWITCHOVER_ROLE might be empty by design, but a non-zero exit indicates that the switchover action execution failed.

cjc7373: I think in most cases we can rely on kb's role info. Engines who need more precise role info can later change to a way that gets role info directly from kernel.

leon-inf (approved):

zjx20 (approved):

PR #1353, opened 2024-12-20
fix: mysql addon error on mysql 5.7 with orc

### LINKED ISSUE #1352: [BUG]fix mysql addon error while using mysql 5.7 and orc
Cluster using mysql 5.7 community and orc can not startup or provision correctly.

There is two problems:

1. `init-mysql-instance-for-orc.sh` uses `GET_MASTER_PUBLIC_KEY` param which is support only in mysql 8.0
2. Var `COMPONENT_NAME` is shortName, should use componentName for complete cluster component name

lancelot1989:
1. check mysql major version
2. use cluster component name instead of short name

resolves #1352


zjx20 (approved):

kubeJocker (approved):
LGTM

PR #1404, opened 2025-01-14
chore: update proxysql addon for 1.0

kubeJocker:


addons/mysql/templates/_helpers.tpl
```diff
@@ -137,7 +139,23 @@ vars:
       credentialVarRef:
         name: root
         password: Required
+  - name: PROXYSQL_PASSWORD
+    value: {{ .Values.auth.proxysql.password }}
```
Y-Rookie, line 143: Why is proxysql's password static?

addons/mysql/config/proxysql.tpl
```diff
@@ -123,8 +123,7 @@ proxysql_servers=
 mysql_servers=
 (
 {{- range $i, $e := until $mysql_replicas }}
-  {{- $mysql_service_host := printf "%s-%s-mysql-%d.%s" $clusterName $mysql_component.name $i $namespace }}
-
+  {{- $mysql_service_host := printf "%s-%s-%d.%s-%s-headless.%s.svc.cluster.local" $clusterName $mysql_component.name $i $clusterName $mysql_component.name $namespace }}
```
Y-Rookie, line 126: Why not use vars to reference MySQL's podFQDNs? This approach has issues when instanceTemplate is present.

addons/mysql/templates/_helpers.tpl
kubeJocker: generated Password by systemaccount

addons/mysql/config/proxysql.tpl
kubeJocker: done

shanshanying (approved):

addons/mysql/scripts/proxysql-entry.sh
```diff
@@ -13,14 +20,25 @@ if [ $FRONTEND_TLS_ENABLED == "true" ]; then
     cp /var/lib/frontend/server/tls.key /var/lib/proxysql/proxysql-key.pem
 fi
 
-function timestamp() {
-    date +"%Y/%m/%d %T"
-}
+function gen_mysql_servers {
```
Y-Rookie, line 23: For future script writing, we can omit the 'function' keyword when defining functions, for the following reasons:

1. More concise
2. POSIX-compliant (better portability)
3. Recommended by most shell scripting style guides

Y-Rookie (approved):

PR #1428, opened 2025-01-20
chore: refactor roles

cjc7373:
relates to https://github.com/apecloud/kubeblocks/pull/8416

### DIFF
--- a/addons/apecloud-mysql/templates/_helpers.tpl
+++ b/addons/apecloud-mysql/templates/_helpers.tpl
@@ -133,17 +133,14 @@ tls:
   keyFile: key.pem
 roles:
   - name: leader
-    serviceable: true
-    writable: true
-    votable: true
+    updatePriority: 5
+    participatesInQuorum: true
   - name: follower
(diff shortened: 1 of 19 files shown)


cjc7373:
Owners, please help check the relavant changes in case I miss something.

zjx20 (dismissed):

wangyelei (dismissed):

addons/apecloud-mysql/templates/_helpers.tpl
```diff
@@ -133,17 +133,14 @@ tls:
   keyFile: key.pem
 roles:
   - name: leader
-    serviceable: true
-    writable: true
-    votable: true
+    updatePriority: 5
```
leon-inf, line 136: It is recommended to use 3, 2, 1, where the absolute priority of roles can be arbitrarily changed. Alternatively, leave room for expansion in between too, so the absolute priority of roles can remain unchanged.

leon-inf (approved):

PR #1432, opened 2025-01-22
chore: support mysql orc semi sync

kubeJocker:


addons/mysql/config/mysql8.0-config.tpl
```diff
@@ -68,7 +68,7 @@ join_buffer_size={{ $join_buffer_size }}
 sort_buffer_size={{ $sort_buffer_size }}
 
 # default_authentication_plugin=mysql_native_password    #From mysql8.0.23 is deprecated.
-# authentication_policy=mysql_native_password,
```
xuriwuyun, line 71: Avoid using mysql_native_password, as it is deprecated.

kubeJocker: When using caching_sha2_password, there seems to be an error with FAILOVER and SwitchOver in Orchestrator. I haven't found a way to configure it yet.

addons/mysql/templates/switchover.yaml
```diff
@@ -37,4 +37,8 @@ spec:
                 - /bin/sh
                 - -c
                 - |
```
shanshanying, line 39: maybe we can add one more parameter for namespace or infer it, if possible, from the context. The constraint of candidate name is much to implict.

PR #1438, opened 2025-01-24
fix: no syncer policy rules

wangyelei:


addons/kblib/templates/_common.tpl
```diff
@@ -0,0 +1,27 @@
+{{- define "syncer.policyRules" -}}
```
shanshanying, line 1: better to name the tempate with 'kblib.syncer....'

xuriwuyun (approved):

xuriwuyun (approved):

PR #1443, opened 2025-02-06
chore: replication orioledb with syncer

1aal:
Removing the previous implementation based on `patroni` and changing it to `syncer`, accessing orioledb in `standalone` and `replication` mode.


addons/orioledb/templates/_helpers.tpl
```diff
@@ -107,47 +79,269 @@ Define orioledb component config constraint name
...
+      command:
+        - /bin/sh
+        - -c
+        - |
+          if [ -z "$KB_SWITCHOVER_ROLE" ]; then
+              echo "role can't be empty"
+              exit 1
+          fi
+
+          if [ "$KB_SWITCHOVER_ROLE" != "primary" ]; then
```
xuriwuyun, line 226: It is not necessary to check KB_SWITCHOVER_ROLE; the syncer can be executed on any pod.

1aal:
It seems that the permission to `delete` configmap is missing from the `policyRules` setting right now, but it's covered in syncer.

When I trigger the switchover twice it reports the following error
image

addons/orioledb/templates/_helpers.tpl
```diff
@@ -107,47 +78,260 @@ Define orioledb component config constraint name
...
+        - name: PGPASSWORD
+          value: $(POSTGRES_PASSWORD)
+        - name: POSTGRESQL_PORT_NUMBER
+          value: '5432'
+        - name: PGDATA
+          value: {{ .Values.dataPath }}
+        - name: PGCONF
+          value: {{ .Values.confPath }}
+        - name: KB_ENGINE_TYPE
+          value: vanilla-postgresql
```
Y-Rookie, line 279: leave a TODO or comment here to explain why use the `vanilla-postgresql` engine type

addons/orioledb/templates/configconstraint.yaml
```diff
@@ -9,25 +9,34 @@ metadata:
...
+          - name: kb-tools
+            image: {{ .Values.image.registry | default "docker.io" }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}
+            asContainerImage: true
+
+  mergeReloadAndRestart: false
+  reloadStaticParamsBeforeRestart: true
   # ConfigurationSchema that impose restrictions on engine parameter's rule
   parametersSchema:
-    # top level orioledb configuration type
+    # top level mysql configuration type
```
Y-Rookie, line 30: mysql?

addons/orioledb/values.yaml
```diff
@@ -1,17 +1,37 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
 
 image:
-  # refer: https://github.com/orioledb/orioledb/blob/main/Dockerfile
   registry: docker.io
-  repository: apecloud/orioledb
-  tag: beta1
+  repository: orioledb/orioledb
+  tag: "latest-pg16"
```
Y-Rookie, line 8: latest tag? and cmpv mapping to `16.4.0`, there will be inconsistencies in the future

Y-Rookie (approved):

zjx20 (approved):

zjx20 (dismissed):

zjx20 (approved):

Y-Rookie (approved):

PR #1448, opened 2025-02-06
chore: support mysql orchestrator semi sync

kubejoker:


addons/mysql/templates/_helpers.tpl
```diff
@@ -214,13 +214,13 @@ serviceRefDeclarations:
         serviceVersion: "^*"
 services:
   - name: default
+    serviceName: server
```
xuriwuyun, line 217: The default service can eliminate the server suffix.

addons/mysql/templates/cmpd-mysql57-orc.yaml
```diff
@@ -11,6 +11,19 @@ spec:
   serviceVersion: 5.7.44
   lifecycleActions:
     {{- include "mysql-orc.spec.lifecycle.common" . | nindent 4 }}
+    accountProvision:
+      exec:
+        container: mysql
+        image: {{ .Values.image.registry | default "docker.io" }}/{{ .Values.image.repository }}:5.7.44
```
xuriwuyun, line 17: The account image can be specified in the cmpv and can use a common account provision across different cmpds.

addons/mysql/templates/_helpers.tpl
kubejoker: podService requires an empty ServiceName to achieve the purpose of having the same ServiceName and PodName, which can be more intuitive.

addons/mysql/templates/cmpd-mysql57-orc.yaml
kubejoker: done

shanshanying (dismissed):

Issue #1491, opened 2025-02-24
[Improvement]kafka addon support extra envs and volumeMounts


lancelot1989:
In certain On-Premise scenarios, container application logs need to be written to specific directories on the disk, and log collection components scan these `hostpath` directories to gather the container application logs.

Therefore, a mechanism needs to be designed to support addons in customizing additional `volumeMount` paths.
Additionally, due to the `subpathExpr` feature of volumeMount, which may utilize environment variable definitions, it is also necessary to support addons in customizing more environment variables.

PR #1492, opened 2025-02-24
feat: support custom volumeMount and envs

lancelot1989:
resovles #1491

### DIFF
--- a/addons/kafka/templates/cmpd-broker-27.yaml
+++ b/addons/kafka/templates/cmpd-broker-27.yaml
@@ -173,6 +173,9 @@ spec:
             value: /tools/server-jaas.properties
           - name: BROKER_MIN_NODE_ID
             value: {{ .Values.kafkaBroker.minNodeId | quote }}
+          {{- with .Values.extraEnvs.kafka }}
+          {{- toYaml . | nindent 10}}
+          {{- end }}
         ports:
           - name: kafka-client
             containerPort: 9092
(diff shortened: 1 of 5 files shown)


addons/kafka/templates/cmpd-broker-27.yaml
```diff
@@ -173,6 +173,9 @@ spec:
             value: /tools/server-jaas.properties
           - name: BROKER_MIN_NODE_ID
             value: {{ .Values.kafkaBroker.minNodeId | quote }}
+          {{- with .Values.extraEnvs.kafka }}
```
Y-Rookie, line 176: It is recommended to define extraEnvs in cmpd.Vars instead of directly in cmpd.Runtime(PodSpec).containers.env, which can avoid unexpected cmpd updates resulting in Runtime/PodSpec changes, and thus causing existing Pod restarts.

addons/kafka/values.yaml
```diff
@@ -80,4 +80,10 @@ container:
...
   securityContext:
     runAsNonRoot: true
-    runAsUser: 1001
\ No newline at end of file
+    runAsUser: 1001
+
+extraVolumeMounts:
+  kafka: []
+
+extraEnvs:
```
Y-Rookie, line 88: What is the expected form of content for extraEnvs and extraVolumeMounts? Are they shared by all components? Please add necessary comments and examples.

addons/kafka/templates/cmpd-broker-27.yaml
lancelot1989: But what if env want to use `fieldRef` for pod annotations or labels？It seems cmpd.Vars does not support `fieldRef`?

addons/kafka/values.yaml
lancelot1989: [3b389ae](https://github.com/apecloud/kubeblocks-addons/pull/1492/commits/3b389ae392d7e7a5d8b502d5fb3a5176ccca3605)

examples added.

Y-Rookie (approved):

caiq1nyu (approved):

PR #1544, opened 2025-03-20
support customise kafka's prometheus metric export config

### LINKED ISSUE #1543: [Improvement]support customize kafka's jmx-exporter config yaml
Sometimes users may want jmx-exporter can export more metrics or metric name should respect another pattern. Making jmx-exporter's config yaml can be configured through helm values can provide more flexibility.

lancelot1989:
resolves #1543


addons/kafka/values.yaml
```diff
@@ -115,4 +115,57 @@ extraEnvs:
...
+      - pattern : java.lang<type=(.*)>
+      - pattern: kafka.log<type=(.+), name=(.+), topic=(.+), partition=(.+)><>Value
+        name: kafka_log_$1_$2
+        labels:
+          topic: $3
+          partition: $4
+      - pattern: kafka.server<type=socket-server-metrics, listener=(.+), networkProcessor=(.+)><>(connection-count)
+        name: kafka_server_socket_server_metrics_$3
+        labels:
+          listener: $1
```
Y-Rookie, line 171: It is recommended to place the default configuration in the config template path rather than directly in the values.

lancelot1989: ok, placing default configuration to file, only custom config using values

zjx20 (approved):

Y-Rookie (approved):

PR #1672, opened 2025-05-09
fix: delete member if possible before pd joining a cluster

cjc7373:
This may later pick to the enterprise addon.


addons/tidb/scripts/pd_start.sh
```diff
@@ -16,13 +16,18 @@ set_join_args() {
...
-        # is unreliable. Consider a scenario with 3 pods: 2 start normally, while the 3rd pod 
-        # pulls image slowly and is still initializing. During this time, the PD cluster 
-        # achieves quorum and begins to work, thus KB's role probe succeeds. 
-        # When the third pod eventually starts, it mistakenly attempts to join the 
-        # cluster, leading to a failure.
         if [[ -n $PD_LEADER_POD_NAME ]]; then
+            echo "query member list from leader"
+            leader_fqdn=$(echo "$replicas" | grep "$PD_LEADER_POD_NAME")
+            members=$(/pd-ctl --pd "http://$leader_fqdn:2379" member | jq -r '.members[] | .name')
+            if echo "$members" | grep -q "$CURRENT_POD_NAME"; then
```
zjx20, line 23: Consider whether there is a more reliable matching method than grep. The grep-based method has flaws in some scenarios. For example, if there is an element named `mypod-10` in members and CURRENT_POD_NAME happens to be `mypod-1`, using grep for matching will incorrectly identify the member as existing. Of course, this situation should be relatively rare.

zjx20 (approved):

addons/tidb/scripts/pd_start.sh
cjc7373: Though the problem is valid, normally there won't be too much pd instances.

wangyelei (approved):

Issue #1691, opened 2025-05-21
[BUG] PostgreSQL AddOn: Secondary Instance Fails to Apply `max_connections` Reduction, Remains Pending Restart​


dingshun-cmss:
## Describe the bug
When modifying the `max_connections` parameter in a PostgreSQL cluster using the PostgreSQL AddOn, **only the primary instance restarts** with the updated value, while **the secondary instance remains pending for restart** and retains the previous value.

## To Reproduce
1. Create a PostgreSQL cluster using the PostgreSQL AddOn.
2. Create an `OpsRequest` to modify `max_connections` to a smaller value. Example:
```yaml
apiVersion: apps.kubeblocks.io/v1alpha1
kind: OpsRequest
metadata:
  generateName: dstest-reconfig-
  namespace: dbaas
spec:
  clusterRef: dstestpg
  reconfigure:
    componentName: postgresql
    configurations:
    - keys:
      - key: postgresql.conf
        parameters:
        - key: max_connections
          value: "50"
      name: postgresql-configuration
  type: Reconfiguring
```
3. Apply the `OpsRequest` and verify the `max_connections` value on both primary and secondary instances.
4. Observe that:
    - The primary reflects the new value (50).
    - The secondary remains pending for restart and retains the old value.

## Expected behavior
Both the primary and secondary instances should restart and apply the new `max_connections` value simultaneously.

## Information
- KubeBlocks version: 0.9.3

## Additional context
When increasing `max_connections`, both instances restart as expected. The issue only occurs when decreasing the value.

dingshun-cmss:
Based on the initial investigation, this bug occurs because the restart request after reconfiguration is only executed on the primary node.

In the PostgreSQL addon's `templates/patroni-reload.yaml`, the `dsn` in `reload.yaml` points to `localhost:8008`, which is the Patroni RESTful API endpoint for a single member. While I'm not entirely clear about the operator's internal workings, our observations confirm that the restart command only executes on the primary node.

Additional findings:
1. When decreasing `max_connections`:
   - Secondaries continue working without issues even if their `max_connections` value remains higher than the primary's
   - The secondary's state remains pending restart

2. When increasing `max_connections`:
   - After the primary restarts with a higher value, secondaries encounter a fatal error because their `max_connections` shouldn't be less than the primary's
   - This triggers an automatic PostgreSQL shutdown on secondaries
   - Patroni detects the shutdown and automatically restarts PostgreSQL
[...]

PR #1694, opened 2025-05-23
fix: pg addon only partial apply config changes

dingshun-cmss:
## Description

Fixed an issue where the PG addon was only partially applying configuration changes.

## Solution

- Use `shellTrigger` to execute a complex shell script on the master
- In the shell trigger:
   - Construct configuration changes as a JSON batch
   - Use the batch mode to pass the JSON to the script via STDIN
- In the script:
   - Apply configuration changes using Patroni's RESTful API
   - Reload the Patroni configuration
   - Restart the master node if required
   - Batch restart standby nodes in parallel if required

Also the unnecessary files after changes are cleaned up

## Related issue

Fix: #1691


wangyelei:
close with https://github.com/apecloud/kubeblocks-addons/pull/1695

PR #1746, opened 2025-06-24
chore: add redis podAntiAffinity

dongjiang1989:
add redis podAntiAffinity

### DIFF
--- a/addons-cluster/redis/templates/_helpers.tpl
+++ b/addons-cluster/redis/templates/_helpers.tpl
@@ -8,6 +8,9 @@ Define redis cluster shardingSpec with ComponentDefinition.
     name: redis
     componentDef: redis-cluster
     replicas: {{ .Values.replicas }}
+    {{- if .Values.podAntiAffinityEnabled }}
+    {{- include "redis-cluster.schedulingPolicy" . | indent 2 }}
+    {{- end }}
     {{- include "redis-cluster.exporter" . | indent 4 }}
     {{- if and .Values.nodePortEnabled (not .Values.hostNetworkEnabled)  (not .Values.fixedPodIPEnabled) }}
     services:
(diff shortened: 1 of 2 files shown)


shanshanying:
Instead of updating ComponentDefinition, I'd sugetst update the heml chart `addons-cluster/reids`, this chart is used to rener a cluster.yaml for redis.
you may add  scheduling policy to 
- https://github.com/apecloud/kubeblocks-addons/blob/main/addons-cluster/redis/

addons/redis/templates/cmpd-redis.yaml
```diff
@@ -320,7 +320,6 @@ spec:
           - /bin/bash
           - -c
           - /scripts/redis-switchover.sh > /tmp/switchover.log 2>&1
-  runtime:
```
wangyelei, line 323: revert it

addons-cluster/redis/templates/_helpers.tpl
```diff
@@ -188,6 +190,56 @@ app.kubernetes.io/managed-by: "kubeblocks"
 apps.kubeblocks.io/component-name: "redis"
 {{- end }}
 
+{{/*
+Redis Cluster schedulingPolicy
+*/}}
+{{- define "redis-cluster.schedulingPolicy" }}
```
wangyelei, line 196: provide a flag to control the use of pod anti-affinity

dongjiang1989: Add `podAntiAffinityEnabled` flag

addons-cluster/redis/templates/_helpers.tpl
```diff
@@ -188,6 +197,56 @@ app.kubernetes.io/managed-by: "kubeblocks"
...
+schedulingPolicy:
+  affinity:
+    podAntiAffinity:
```
zjx20, line 206: Why set both `preferredDuringSchedulingIgnoredDuringExecution` and `requiredDuringSchedulingIgnoredDuringExecution` at the same time? Moreover, their `labelSelector` and `topologyKey` look identical, so the result of this setup is that `requiredDuringSchedulingIgnoredDuringExecution` takes precedence, making `preferredDuringSchedulingIgnoredDuringExecution` redundant.

dongjiang1989: Thanks @zjx20 
Great. The `preferredDuringSchedulingIgnoredDuringExecution` setting has no practical effect in this case. For Redis deployments, it's more appropriate to use only `preferredDuringSchedulingIgnoredDuringExecution`.

wangyelei (approved):

PR #1770, opened 2025-07-07
chore: improve zookeeper jvm settings

kizuna-lek:


addons/zookeeper/scripts/startup.sh
```diff
@@ -31,8 +31,36 @@ compare_version() {
   esac
 }
 
+set_jvm_configuration() {
+  system_memory_in_mb=$(($(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)/1024/1024))
```
zjx20, line 35: This path is for cgroup v1. Currently, more and more distributions are using cgroup v2 by default, so please consider supporting it. At least the script should not crash when running under cgroup v2.

zjx20: Reviewed by copilot:

Yes, in cgroup v2, the value of /sys/fs/cgroup/memory.max can be a string such as "max" (meaning no limit). This is different from cgroup v1, where memory.limit_in_bytes is always an integer.

**Review of the PR:**

1. **Cgroup logic:**
   - The PR checks for /sys/fs/cgroup/memory/memory.limit_in_bytes (cgroup v1) and /sys/fs/cgroup/memory.max (cgroup v2).
   - However, the PR does not handle the case where /sys/fs/cgroup/memory.max is set to "max" (unlimited). Attempting arithmetic on "max" will cause a script error.
   - To be robust, you should check if the value of memory.max is "max" before attempting to use it as an integer. If it is "max", you may want to fallback to detecting system memory with another method (e.g., /proc/meminfo) or set a sensible default.

2. **Other logic:**
   - The JVM heap size calculation logic appears standard: it computes half and quarter of available memory, clamps them to 1024MB and 8192MB respectively, and picks the larger value for MAX_HEAP_SIZE.
   - The rest of the logic (setting JVMFLAGS, function structure, and invocation in start()) is clean and functional.

**Suggested fix for cgroup v2 edge case:**
Add a check for "max" in memory.max:
```bash
elif [ -f /sys/fs/cgroup/memory.max ]; then
    mem_limit=$(cat /sys/fs/cgroup/memory.max)
    if [ "$mem_limit" = "max" ]; then
        # fallback or set a default, e.g. use /proc/meminfo or set to a high value
    else
        system_memory_in_mb=$(($mem_limit/1024/1024))
    fi
```

**Summary:**  
- The PR improves JVM memory settings and covers both cgroup v1 and v2, but does not handle the "max" string in cgroup v2 which can cause failures in some environments.
- No other significant issues found in the rest of the logic.
- Recommend updating the script to account for "max" in /sys/fs/cgroup/memory.max for full robustness.

zjx20 (approved):
Should pick it to the release-1.0 branch?

kizuna-lek:
> Should pick it to the release-1.0 branch?

yes, i will pick it into release-1.0

PR #1774, opened 2025-07-07
feat: support client-service tls and tls between components 

cjc7373:


cjc7373:
suppressed by https://github.com/apecloud/kubeblocks-addons/pull/2635

PR #1844, opened 2025-07-28
chore: add pcr and pd for zk

kizuna-lek:


addons/zookeeper/templates/cmpd.yaml
```diff
@@ -75,7 +75,7 @@ spec:
       namespace: {{ .Release.Namespace }}
       defaultMode: 0755
   configs:
```
shanshanying, line 77: set `externalManaged: ture` API pls

shanshanying (approved):

PR #1854, opened 2025-07-31
chore: chagne pg scope label

ldming:
Use the existing label "workloads.kubeblocks.io/instance" as the scope label for PostgreSQL to avoid requiring users to explicitly specify the label when creating a cluster.


shanshanying (approved):

wangyelei (approved):

wangyelei:
fix https://github.com/apecloud/kubeblocks/issues/9461

PR #1974, opened 2025-09-08
fix: switchover failed when replicas is 1

wangyelei:


wangyelei:
fix https://github.com/apecloud/kubeblocks/issues/9717

PR #1977, opened 2025-09-08
fix: keeper switch over failed

wangyelei:


addons/clickhouse/scripts/keeper-switchover.sh
```diff
@@ -14,26 +14,21 @@ if [ $COMPONENT_REPLICAS -lt 2 ]; then
...
+    exit 0
 fi
 
-if [[ -z "$candidate_fqdn" ]]; then
-  echo "ERROR: Could not find a candidate follower."
-  exit 1
+if [ "$KB_SWITCHOVER_ROLE" != "leader" ]; then
+    echo "switchover not triggered for primary, nothing to do, exit 0."
+    exit 0
 fi
```
loomts, line 25: duplicate check with line8-11

PR #2230, opened 2025-11-12
chore: ck ping use wget instead of curl

loomts:


shanshanying (approved):
better to use CURL with ca info. but for the moment, it is acceptable to use wget with no-cert flag for readiness check.

PR #2259, opened 2025-11-20
feat: kafka support custom sasl

caiq1nyu:
there are some differences in the use of SASL across different versions and architectures of Kafka. this pull request provides an authentication solution that is compatible with all existing versions.

```
# SASL settings
sasl:
  enable: false
  # if true, the built-in SASL server callback handler in the knowledge base is used, and users are managed by modifying the key.
  # if false, use kafka native SASL callback handler
  # server.properties: 
  #   useKBBuildInSasl=true -> sasl.server.callback.handler.class=${KBClassName}
  #   useKBBuildInSasl=false -> sasl.server.callback.handler.class=''
  useKBBuildInSasl: true
  # [PLAIN,SCRAM-SHA-256,SCRAM-SHA-512]
  # when useKBBuildInSasl=true, only supported [PLAIN] yet. SCRAM-SHA-256, SCRAM-SHA-512 will be comming soon
  mechanisms:
    - PLAIN
  # should be one of sasl.mechanisms setting  
  interBrokerProtocol: PLAIN
```
this PR includes built-in SASL (useKBBuildInSasl=true) which only supports mechanisms=plain, future versions will support SCRAM.

users can dynamically maintain their accounts by directly editing the account-secret:
```
stringData:
  # accounts: |
  # [mechanism=PLAIN] format:
  #   {accountName}:{md5Password}
  accounts: ""
```

### DIFF
--- a/addons-cluster/kafka/templates/account-secret.yaml
+++ b/addons-cluster/kafka/templates/account-secret.yaml
@@ -3,8 +3,10 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ include "kafka-broker-accounts-secret-name" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
     app.kubernetes.io/name: {{ include "kblib.clusterName" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/component: {{ include "kafka-broker-component" . }}
 type: Opaque
 stringData:


addons-cluster/kafka/templates/account-secret.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: Secret
+metadata:
+  name: {{ include "kafka-broker-accounts-secret-name" . }}
+  labels:
+    app.kubernetes.io/name: {{ include "kblib.clusterName" . }}
+    app.kubernetes.io/component: {{ include "kafka-broker-component" . }}
+type: Opaque
+stringData:
+  # accounts: |
+  # [mechanism=PLAIN] format:
```
shanshanying: would it be better to set accounts through helm values

caiq1nyu: custom accounts are generally not created during cluster first create; addon users need to add or modify this secret during use.

shanshanying (approved):

PR #2426, opened 2026-02-04
chore: remove tidb slowlog container

cjc7373:
Slowlog is stored in file. This container merely displays it. Due to some signal handling problems, this container blocks the pod grace shutdown process. So I think the easiest way is to delete it.

### DIFF
--- a/addons/tidb/templates/cmpd-tidb.yaml
+++ b/addons/tidb/templates/cmpd-tidb.yaml
@@ -82,18 +82,6 @@ spec:
             mountPath: /var/log/tidb
           - name: config
             mountPath: /etc/tidb
-      - name: slowlog
-        imagePullPolicy: IfNotPresent
-        command:
-          - bin/sh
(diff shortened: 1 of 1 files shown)


shanshanying:
can we  TRAP the TERM signal ?

shanshanying:
It seems sleep blocks TERM.  can we for example, kill the  TAIL task by PID on receiving the signal
e.g
```
# trap TERM
trap graceful_exit TERM INT

# get tail pid and wait for it
tail -n0 -F /tmp/log.txt &
TAIL_PID=$!
wait $TAIL_PID
```

and this `graceful_exit` function kill the TAIL_PID, for example:

```
    graceful_exit() {
      kill -TERM $TAIL_PID 2>/dev/null
      exit 0
    }
```

wangyelei (approved):

shanshanying (approved):

PR #2573, opened 2026-04-13
feat(qdrant): add v1.17.1, bump appVersion

nayutah:
- Add qdrant 1.17.1 to versions list
- Bump appVersion to 1.17.1
- Add default: true to cluster topology
- Fix annotations: use qdrant.annotations helper consistently across cmpd/cmpv/clusterdefinition
- Remove duplicate skip-immutable-check from cmpd (now in helper)
- Remove kblib.helm.resourcePolicy from qdrant.annotations helper


addons/qdrant/templates/_helpers.tpl
```diff
@@ -54,7 +54,6 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 Common annotations
 */}}
 {{- define "qdrant.annotations" -}}
-{{ include "kblib.helm.resourcePolicy" . }}
```
wangyelei, line 57: why remove it? keep cmpd for upgrading helm

nayutah: add it back

PR #2592, opened 2026-04-20
feat(valkey): add Valkey addon with sentinel-based replication

### LINKED ISSUE #2591: feat(valkey): add Valkey addon with sentinel-based replication
## Summary

Add a new Valkey addon supporting sentinel-based replication topology for KubeBlocks.

## Features

- Sentinel-based HA with automatic and targeted switchover
- Support for Valkey 8.x and 9.x
- TLS support
- ACL management
- Backup and restore via datafile method
- Full ShellSpec unit test coverage for all scripts

## Operations Verified (EKS, KubeBlocks 1.0.2)

All 20 operations pass: cluster creation, stop/start/restart, vertical/horizontal scaling, volume expansion, expose, upgrade (8→9), reconfiguring, switchover (auto + targeted), backup/restore, TLS, scheduled backup, custom secret, decommission.

weicao:
Closes #2591

## Summary

- Add Valkey addon supporting sentinel-based replication topology
- Supports Valkey 8.x and 9.x with separate ComponentDefinitions
- TLS, ACL, backup/restore, reconfiguring, and switchover (auto + targeted)

## Script Highlights

- `valkey-start.sh`: bootstrap-primary election via sentinel query + lexicographic fallback; cascading-chain detection with `follow_slave_to_master()`
- `valkey-sentinel-start.sh`: `_background_monitor_discovery` for simultaneous pod restart recovery; TLS-aware sentinel replication
- `switchover.sh`: strict candidate verification — OpsRequest fails if wrong node elected
- `valkey-member-leave.sh`: config-epoch–based sentinel selection; skip RESET on fast-path
- `check-role.sh`: sync-stall detection with SIGTERM recovery

## Test Results (EKS ap-southeast-1, KubeBlocks 1.0.2)

All 20 documented operations pass. Full suite: 620 PASS, 0 FAIL across smoke, regression, chaos, TLS, upgrade, sentinel, ACL, custom-secret, and rebuild suites.


weicao (commented):
## Blake Review: COMMENT

Well-engineered Valkey addon with impressive depth on sentinel quorum correctness. The quorum-invalid fallback logic in check-role.sh is sound, unit test coverage is thorough, and architecture follows KubeBlocks addon patterns. No critical safety blockers, but two moderate issues worth discussing.

### Moderate Issues

**1. Password injection via word-splitting** (check-role.sh:214-221, all scripts)
Passwords injected into CLI command strings without quoting (`cmd="${cmd} -a ${VALKEY_DEFAULT_PASSWORD}"`). Safe under current passwordGenerationPolicy (numSymbols: 0), but fragile if policy is relaxed. Consider using bash arrays for CLI commands instead of string concatenation. Pervasive across ~10 scripts.

**2. Authoritative role ignores local INFO when quorum agrees** (check-role.sh:472-478)
Role decision based purely on run_id match without checking local role:master. By design (round-3 race fix), but creates a theoretical window during sentinel convergence lag. Mitigated by sentinel's fast runid update, isExclusive: true, and self-heal daemon. Worth documenting. Also: test helper `decide_role` semantically diverges from production code (test checks role_line as fallback, production does not).

### Low / Nit Items
- Debug JSON output lacks escaping (debug-only, low impact)
- `sentinel masters` (plural) parsing assumes single-master topology (correct for KB, fragile for reuse)
- roleProbe timeoutSeconds: 3 may be tight for multi-sentinel sequential queries
- Chart.yaml appVersion only references 9.0.0 despite dual 8.x/9.x support

### Strengths
- Quorum fallback logic correct and well-reasoned (no_sentinel_env/no_sentinel_configured vs insufficient_valid/split_view)
- FQDN-only gate fix correctly closes cross-mode label-strip race
- Fork-and-zombie discipline in check-role.sh is exemplary
- Test coverage thorough (unanimous, split, partial, unreachable, 3- and 5-sentinel topologies)
- Outstanding inline documentation with evidence anchors

weicao:
Thanks for the thorough review, @blake-reviewer.

**Issue 1 (Password word-splitting)**: Acknowledged. This is pre-existing across all scripts and currently safe under numSymbols: 0 policy. We'll track this as a follow-up improvement (bash array refactor). Not blocking for this release since the password policy is controlled.

**Issue 2 (Authoritative role ignores local INFO)**: This is intentional — see PR #2676 discussion where we specifically moved away from trusting local INFO when sentinel quorum is ambiguous. The sentinel runid match is the higher-fidelity signal. Will add a comment documenting the design rationale at lines 472-478. The test helper `decide_role` divergence is a valid nit; will fix in a follow-up.

**Note**: PR #2733 (child PR into feat/valkey-addon) addresses an improvement to the quorum-invalid fallback — `role:master` now exits non-zero (preserving existing label) instead of emitting `secondary`. Once #2733 merges into feat/valkey-addon, this PR's head will be rebased to include it.

Low/nit items acknowledged for tracking.

weicao:
## Response to Blake's Review

Thanks for the thorough review. Both moderate findings are valid observations. Here is how we will handle them:

**1. Password injection via word-splitting** (check-role.sh:214-221)
Acknowledged. Currently safe under `passwordGenerationPolicy.numSymbols: 0`, but fragile if policy changes. This is a **post-merge hardening item** — we will track it as a follow-up PR to switch CLI argument construction from string concatenation to bash arrays across affected scripts. Not a release blocker because the current policy guarantees no special characters.

**2. Authoritative role ignores local INFO when quorum agrees** (check-role.sh:472-478)
This is by design — the round-3 race fix intentionally trusts sentinel run_id match over local INFO to prevent the exact label-cascade bug that #2676 addresses. The theoretical convergence lag window is mitigated by three layers: (a) sentinel fast runid update after failover, (b) `isExclusive: true` preventing dual-primary, (c) self-heal daemon. The `decide_role` test helper divergence is also intentional — test checks both paths for full branch coverage while production takes the safe-only path. We will add an inline comment documenting this design rationale.

Both items will be addressed as follow-up PRs after the current release validation completes.

addons/valkey/templates/paramsdef.yaml
```diff
@@ -0,0 +1,143 @@
...
+---
+apiVersion: parameters.kubeblocks.io/v1alpha1
+kind: ParametersDefinition
+metadata:
+  name: {{ printf "valkey%s-pd" .major }}
+  labels:
+    {{- include "valkey.labels" $ | nindent 4 }}
+  annotations:
+    {{- include "valkey.annotations" $ | nindent 4 }}
+spec:
```
leon-ape, line 11: This ParametersDefinition is missing the direct binding fields required by the new parameter path. Please declare the matching `componentDef`, `templateName: valkey-replication-config`, `fileName: valkey.conf`, and `fileFormatConfig` here instead of relying on the legacy PCR fallback.

addons/valkey/templates/pcr.yaml
```diff
@@ -0,0 +1,23 @@
+{{- range .Values.valkeyVersions }}
+---
+apiVersion: parameters.kubeblocks.io/v1alpha1
+kind: ParamConfigRenderer
```
leon-ape: ParamConfigRenderer is deprecated and should not be introduced for a new addon. The PD should carry the binding to the ComponentDefinition/config item directly, and this PCR resource should be removed.

addons/valkey/config/valkey9-config.tpl
```diff
@@ -0,0 +1,89 @@
+# Valkey 9.x default configuration template.
```
leon-ape: This file currently only differs from `valkey8-config.tpl` by the comment. Please avoid splitting config templates by major version unless there is an actual config difference; otherwise a single template is easier to maintain and validate.

addons/valkey/templates/cmpd.yaml
```diff
@@ -0,0 +1,619 @@
...
+                - -c
+                - /scripts/pre-stop.sh
+
+      {{- if $.Values.metrics.enabled }}
+      - name: metrics
+        imagePullPolicy: IfNotPresent
+        securityContext:
+          runAsNonRoot: true
+          runAsUser: 1001
+        resources:
```
leon-ape: The ComponentDefinition should not bake in default container resources. Resource sizing should come from the Cluster/addon-cluster values so the addon definition does not impose one test-sized default on every user deployment. The sentinel CMPD has the same issue.

addons/valkey/templates/cmpd.yaml
```diff
@@ -0,0 +1,619 @@
...
+              fieldRef:
+                apiVersion: v1
+                fieldPath: metadata.name
+          - name: KB_HOST_IP
+            valueFrom:
+              fieldRef:
+                apiVersion: v1
+                fieldPath: status.hostIP
+          - name: KB_POD_FQDN
+            value: "$(CURRENT_POD_NAME).$(VALKEY_COMPONENT_NAME)-headless.$(CLUSTER_NAMESPACE).svc.{{ $.Values.clusterDomain }}"
```
leon-ape, line 399: Please avoid hand-building the current Pod FQDN from naming conventions. The component already exposes `podFQDNs`; use that mapping to resolve the current pod instead of depending on headless service name, cluster domain, and pod naming details.

addons/valkey/dataprotection/post-restore-sentinel.sh
```diff
@@ -0,0 +1,191 @@
...
+comp_prefix="${pod_name%-*}"           # e.g. mycluster-valkey
+cluster_prefix="${comp_prefix%-*}"     # e.g. mycluster
+cluster_domain="${CLUSTER_DOMAIN:-cluster.local}"
+
+echo "INFO: resolved target context pod=${pod_name} namespace=${namespace} comp=${comp_prefix}"
+
+# Sentinel component name follows the ClusterDefinition topology:
+# component name "valkey-sentinel" → pod name "<cluster>-valkey-sentinel-<n>"
+sentinel_comp="${cluster_prefix}-valkey-sentinel"
+sentinel_headless="${sentinel_comp}-headless.${namespace}.svc.${cluster_domain}"
```
leon-ape, line 98: This derives Sentinel and data Pod FQDNs from naming conventions and a fixed topology name. That is fragile for restore paths; prefer explicit injected topology/FQDN information or another KubeBlocks-provided mapping rather than reconstructing service names manually.

addons/valkey/templates/clusterdefinition.yaml
```diff
@@ -0,0 +1,49 @@
...
+    #
+    # Failover sequence (automatic, no human action needed):
+    #   1. Sentinel detects primary unreachable (down-after-milliseconds = 20 s)
+    #   2. 2/3 sentinels reach quorum and elect a sentinel leader
+    #   3. Leader promotes the best replica: SLAVEOF NO ONE
+    #   4. Leader reconfigures other replicas to point to new primary
+    #   5. Leader updates its own conf + broadcasts to peer sentinels
+    #   6. roleProbe in KubeBlocks detects new roles, updates Service routing
+    - name: replication
+      components:
```
leon-ape, line 36: The replication topology resolves data and sentinel components independently through broad regexes, but the CMPDs later reference each other by matching major-specific compDefs. A valid spec can therefore select data=8.x and sentinel=9.x, making the cross-component vars empty and breaking sentinel registration/discovery. Please enforce or document a single version contract for the pair.

addons/valkey/templates/cmpd.yaml
```diff
@@ -0,0 +1,619 @@
...
+Each CMPD is immutable after creation — we version the name with Chart.Version
+so upgrades create new resources rather than patching existing ones.
+In-use clusters keep pointing at the old CMPD; new clusters use the new one.
+*/}}
+{{- range .Values.valkeyVersions }}
+---
+apiVersion: apps.kubeblocks.io/v1
+kind: ComponentDefinition
+metadata:
+  name: {{ printf "%s-%s" .componentDef $.Chart.Version }}
```
leon-ape: The CMPD name embeds `Chart.Version`, while the ClusterDefinition selects CMPDs through a broad regex and KubeBlocks picks the latest match by lexicographic name order. SemVer strings are not lexicographically ordered (`0.9.0` sorts after `0.10.0`), so future chart upgrades can keep selecting an older CMPD. Please use a selection/naming strategy that is deterministic across SemVer upgrades.

leon-ape (commented):
Review comments focused on addon contract completeness and KubeBlocks addon conventions.

leon-ape:
This PR also lacks the user-facing contract for the new addon: there is no `addons-cluster/valkey` chart and no `examples/valkey`. For a new addon, the contract should cover not only the internal CD/CMPD/CMPV/PD/BPT/ActionSet resources, but also a renderable Cluster chart/example that proves the supported topologies and operations can be used end-to-end. Without that, standalone/replication/TLS/backup/restore/reconfigure/switchover are not easy to validate from the repo.

addons/valkey/scripts/switchover.sh
```diff
@@ -0,0 +1,429 @@
...
+    # Restore priorities on both success and failure paths — Sentinel has now
+    # committed +switch-master (or timed out), so the bias is no longer needed.
+    IFS=',' read -ra all_fqdns <<< "$(pod_fqdns_with_candidate "${candidate_fqdn}")"
+    for fqdn in "${all_fqdns[@]}"; do
+      set_replica_priority "${fqdn}" 100 || true
+    done
+    return "${wfnm_rc}"
+  else
+    # No candidate: any new master is a valid outcome — best-effort wait only.
+    wait_for_new_master "" "${KB_SWITCHOVER_CURRENT_FQDN}" || true
```
leon-ape: No-candidate switchover should not ignore this failure. `SENTINEL FAILOVER` only means Sentinel accepted the asynchronous failover command; it does not prove a new primary was elected. Since KubeBlocks has no candidate role to verify later in this path, this can report the switchover Ops as successful while no new master was confirmed. Please return the `wait_for_new_master` result here and fail the action if no new master is observed.

addons/valkey/templates/backupactionset.yaml
```diff
@@ -0,0 +1,65 @@
...
+  env:
+    - name: DATA_DIR
+      value: {{ .Values.dataMountPath }}
+    - name: TLS_MOUNT_PATH
+      value: {{ .Values.tlsMountPath }}
+  backup:
+    preBackup: []
+    postBackup: []
+    backupData:
+      image: {{ include "valkey.defaultImage" . }}
```
leon-ape: This fixed default image does not use the `IMAGE` env produced by the BackupPolicyTemplate versionMapping, so the version-specific mapping described there will not affect backup jobs. DP expands ActionSet images from env only when the image contains a placeholder such as `$(IMAGE)`. As written, Valkey 9.x backups can still run with the default Valkey 8.x image. Please use the mapped image for backup, prepareData, and postReady, or remove the versionMapping claim and document the intentional fixed image.

addons/valkey/templates/backupactionset.yaml
```diff
@@ -0,0 +1,65 @@
...
+        - -c
+        - |
+          {{- .Files.Get "dataprotection/restore.sh" | nindent 10 }}
+    postReady:
+      # Re-register the primary with Sentinel (Sentinel's own PVC is not backed up).
+      # This is best-effort because current DP BackupPolicyTemplate env cannot
+      # inject cross-component Sentinel credentials.
+      - job:
+          image: {{ include "valkey.defaultImage" . }}
+          onError: Continue
```
leon-ape: If this postReady job is part of the restore contract for re-establishing Sentinel monitoring, `Continue` can turn a failed Sentinel re-registration into a successful Restore. That leaves the data restored but HA monitoring not restored. Please either make this fail the Restore, or make it explicit that Sentinel's own background discovery is the authoritative recovery path and this job is only non-contract best effort.

addons/valkey/dataprotection/post-restore-sentinel.sh
```diff
@@ -0,0 +1,191 @@
...
+# Use full component name as master-name (matches register-to-sentinel logic)
+master_name="${comp_prefix}"
+
+# ── find current primary ─────────────────────────────────────────────────────
+comp_headless="${comp_prefix}-headless.${namespace}.svc.${cluster_domain}"
+
+find_primary_fqdn() {
+  local ordinal fqdn role
+  for ordinal in 0 1 2 3 4; do
```
leon-ape: This restore hook hard-codes the data pod search to ordinals 0..4 and derives component/headless names from naming conventions. For larger clusters, custom component names, or any future naming/topology change, it can miss the actual primary and still be hidden by `onError: Continue`. Please use an explicit KB/DP-provided pod/FQDN mapping or pass the expected component topology into the job instead of deriving and bounding it here.

addons/valkey/dataprotection/post-restore-sentinel.sh
```diff
@@ -0,0 +1,191 @@
...
+if ! ${_probe_base} PING 2>/dev/null | grep -q "PONG"; then
+  if ${_probe_base} --tls --insecure PING 2>/dev/null | grep -q "PONG"; then
+    _tls_args="--tls --insecure"
+    echo "INFO: TLS detected via connection probe — using --tls --insecure"
+  fi
+fi
+
+# Build a valkey-cli prefix for the data nodes
+if [ -n "${DP_DB_PASSWORD}" ]; then
+  data_cli_base="valkey-cli --no-auth-warning ${_tls_args} -p ${data_port} -a ${DP_DB_PASSWORD}"
```
leon-ape: Please avoid building `valkey-cli` commands as strings containing secret values. Passwords with spaces or shell glob characters will be split/expanded when `${data_cli_base}` or `${sentinel_cli_base}` is executed, and the same pattern exists in backup/restore Sentinel ACL scripts. Build the command as an array, like the lifecycle scripts already do, and pass password values as quoted array elements.

leon-ape (commented):
Additional review comments for the Valkey addon implementation.

weicao:
## Review response — all 13 comments addressed

Head: `3fb9c9a4`  
CI: all green (helm lint, shell-check, shellspec-test, generate-files, addons-cluster-helm, label-check)

### Fixed in this head (10 items)

| # | File | Comment | Fix |
|---|------|---------|-----|
| 1 | `paramsdef.yaml` | PD missing direct binding fields | Added `componentDef`, `templateName`, `fileName`, `fileFormatConfig` |
| 2 | `pcr.yaml` | PCR deprecated, should not be introduced | Deleted `pcr.yaml` entirely |
| 3 | `valkey9-config.tpl` | Config template dedup | Merged into single `valkey-config.tpl` |
| 4 | `cmpd.yaml:571` | No default container resources | Removed hardcoded resources from exporter and sentinel CMPD |
| 7 | `clusterdefinition.yaml:36` | Version contract for data+sentinel pair | Per-major topologies (`replication-8`, `replication-9`) with exact compDef references; default `replication` uses latest major |
| 8 | `cmpd.yaml:12` | CMPD name SemVer ordering | Removed `Chart.Version` suffix; names are now major-scoped and stable |
| 9 | `switchover.sh:381` | No-candidate ignores wait_for_new_master failure | Now returns error if no new primary confirmed; added shellspec test |
| 10 | `backupactionset.yaml:35` | Fixed image ignores versionMapping | Changed all 3 image references to `$(IMAGE)` |
| 11 | `backupactionset.yaml:59` | onError:Continue hides Sentinel failure | Changed to `onError: Fail` |
| 12 | `post-restore-sentinel.sh:99` | Hardcoded ordinals 0..4 | Dynamic scan via `DATA_REPLICA_COUNT` with consecutive-unreachable early exit |

### Acknowledged as follow-up (3 items)

| # | File | Comment | Status |
|---|------|---------|--------|
| 5 | `cmpd.yaml:399` | Hand-building Pod FQDN | The `POD_FQDN` env var is built from `$(CURRENT_POD_NAME).$(COMPONENT)-headless...`. This pattern is shared with Redis and other addons; KB does not yet inject a direct `KB_POD_FQDN` for the main container (only for roleProbe). Will address when KB provides a standard FQDN injection mechanism. |
| 6 | `post-restore-sentinel.sh:90` | Naming convention derivation for Sentinel FQDNs | Current DP BackupPolicyTemplate env schema cannot inject cross-component Sentinel FQDNs into the restore job. The script must derive them from naming conventions. Will revisit when DP supports cross-component topology injection. |
| 13 | `post-restore-sentinel.sh:47` | Password word-splitting in CLI string commands | Valid hardening. Will convert `data_cli_base`/`sentinel_cli_base` from string to array in a follow-up commit. |

### Verification
- Offline: helm lint/render/YAML parse, structure assertions, shellspec (56 examples, 0 failures)
- CI: all checks pass on `3fb9c9a4`
- Peer review: Bob reviewed Alice's changes (scripts/DP/ActionSet); Alice reviewed Bob's changes (templates/chart/examples)
- Runtime full-suite run scheduled as next step

weicao:
@Leon11 您的第二轮 4 条 findings 已在 `37b591f5` 全部处理：

1. **postReady fail-closed** (line 157): 脚本改为至少成功配置 1 个 Sentinel 才 exit 0；当 `SENTINEL_REPLICA_COUNT` 明确指定时，按预期数量验收，不足则 exit 1。
2. **clusterDomain / naming** (line 88): `cluster_domain` 从 `DP_DB_HOST` FQDN 自动推导，不再硬编码 `cluster.local`；data/sentinel ordinal scan 改为 bounded scan，支持 `DATA_REPLICA_COUNT` / `SENTINEL_REPLICA_COUNT` / `POST_RESTORE_*_SCAN_LIMIT` override。
3. **restore.yaml API 合同** (line 7): 示例已改为 `Cluster.spec.restore.source` + `parameters` (key: `dataprotection.kubeblocks.io/volume-restore-policy`)，不再使用旧 annotation。
4. **password word-splitting** (line 47): `data_cli_base` / `sentinel_cli_base` / `_probe_base` 全部改为 bash array，密码含空格或通配符不会被拆词。

新增 ShellSpec 合同测试覆盖以上 4 点。CI 验证中。

leon-ape (approved):

PR #2676, opened 2026-05-25
fix(valkey): narrow legacy roleProbe fallback when sentinel quorum is invalid

weicao:
## Summary

`check-role.sh` no longer emits legacy single-token `primary` from `local INFO replication role:master` when the Sentinel quorum is transiently invalid but Sentinel is configured. In that window it emits `secondary` instead. The standalone / no-Sentinel paths (`__quorum_decision_reason` = `no_sentinel_env` / `no_sentinel_configured`) keep the original local-INFO mapping.

## Why

This addon-side patch addresses a Bug-B-class regression observed when the addon is paired with the post-apecloud/kubeblocks#10283 controller (which splits `LastRoleEventVersionAnnotationKey` from `LastRoleEngineVersionAnnotationKey`). The controller's `removeExclusiveRoleLabels()` only gates peer cleanup against the incoming event's own mode. A legacy `primary` event from one Pod therefore strips a sibling Pod's engine-versioned `primary` label, and the strict `engine:>` staleness gate then rejects subsequent same-version repair events because the previousRole annotation has been cleared.

The trigger is exactly the transient window in which check-role.sh falls back to legacy output:

- Pod A has Sentinel quorum, emits `primary <epoch>` (engine-versioned). Controller writes Pod A's role label + `LastRoleEngineVersionAnnotationKey=<epoch>`.
- Pod B has the same local `INFO replication role:master` view (transitional state during a switchover, scale-out, or a Sentinel split) but its quorum is invalid (`insufficient_valid` / `split_view`). Pre-patch, Pod B emits plain `primary` from local INFO. Post-patch, Pod B emits `secondary`.

On the closed-PR apecloud/kubeblocks#10280 path this issue did not surface because the controller did not split annotation keys. On the apecloud/kubeblocks#10283 path the issue becomes visible whenever two pods straddle modes, which is exactly the case across the switchover convergence window.

## Reproduction reference

This PR's RED case is sourced from the live freeze archive sha `575ed6c444cf6224c74629a2abc28f16f80c35e62f13f6cada4c5ae782302920`, captured from a full-suite acceptance round on idc vcluster valkey-fresh-0008 (cluster `vlk-smoke-lc-...` after T09 pre-scale-in switchover). The controller timeline showed `valkey-3 primary engine:3 accepted (exclusiveClean=true)` followed seconds later by `valkey-2 primary legacy accepted (exclusiveClean=true)` and then `valkey-2 secondary engine:3 accepted`, after which the controller rejected every subsequent `valkey-3 primary engine:3` event as `staleRoleEventVersion`. Cluster `Component.phase` stayed `Failed` with `RoleProbeNotDone: some instances do not have roles`; the engine data plane was healthy throughout. The companion controller-side mitigation (cross-mode aware peer cleanup / same-version engine repair) lives in apecloud/kubeblocks#10283.

## Scope

- `addons/valkey/scripts/check-role.sh`: split the legacy fallback into two branches keyed by `__quorum_decision_reason`. Sentinel-configured-but-quorum-invalid now emits `secondary` for both `role:master` and `role:slave`. Sentinel-not-configured keeps the original local-INFO mapping. Unknown-role guard unchanged.
- `addons/valkey/scripts-ut-spec/check_role_spec.sh`: add a `legacy fallback cross-mode safety` Describe block covering four checks (Sentinel configured + quorum invalid + INFO=master → secondary; Sentinel not configured + INFO=master → primary; role:slave maps to secondary in the no_sentinel branch; unknown role exits non-zero in both branches). Total shellspec count holds; the new tests pass alongside the existing engine-version-trailer suite.

Out of scope:

- Controller-side cross-mode peer cleanup / same-version engine repair — that lives in apecloud/kubeblocks#10283.
- Sentinel `failover-abort-slave-timeout` tuning — unrelated to the cross-mode write-then-strip mechanism.
- Standalone / no-Sentinel topology fallback — explicitly preserved by the `no_sentinel_env` / `no_sentinel_configured` branch.

## Test plan

- [x] `shellspec` passes on the updated `check_role_spec.sh` (legacy fallback cross-mode safety: Sentinel configured + INFO=master → secondary; no_sentinel + INFO=master → primary; role:slave → secondary; unknown role → exit 1).
- [x] Existing engine-version-trailer Describe blocks unchanged and pass.
- [ ] End-to-end: rerun the full-suite acceptance loop on idc vcluster valkey-fresh-0008 with `apecloud/kubeblocks:1.2.0-alpha.1-pr10283-714f684b` controller image + this addon HEAD, looking for T09 pre-scale-in switchover to converge without `RoleProbeNotDone` deadlock.

## Compatibility

This patch only narrows the fallback path. Engine-version trailer output is unchanged for any path where Sentinel quorum is achieved. Standalone topologies still emit `primary` from local `INFO=master`. The companion controller-side patch (apecloud/kubeblocks#10283) is the contract-level fix; this PR is a defence-in-depth that reduces the trigger surface on the addon side and unblocks acceptance runs while the controller patch lands.

## Public hygiene checks

- [x] Commit message and PR body state public facts only (no internal task numbers, channel names, or agent code names).
- [x] All cross-references use the upstream PR / archive identifiers.
- [x] Status: defence-in-depth `narrow legacy fallback`; not a contract change.

### DIFF
--- a/addons/valkey/scripts-ut-spec/check_role_spec.sh
+++ b/addons/valkey/scripts-ut-spec/check_role_spec.sh
@@ -841,6 +841,62 @@ Describe "Valkey Check-Role Bash Script Tests"
         The stdout should equal "2"
       End
     End
+
+    Context "quorum-entry gate depends only on SENTINEL_POD_FQDN_LIST"
+      # Background: Bob2 review on PR apecloud/kubeblocks-addons#2676.
+      # The earlier gate `[ -n "${SENTINEL_PASSWORD}" ] && [ -n "${SENTINEL_POD_FQDN_LIST}" ]`
+      # would route a sentinel-configured-but-password-missing topology
+      # to `no_sentinel_env`, and the no-sentinel fallback emits legacy
+      # `primary` from `role:master` — which is exactly the cross-mode
+      # strip path this PR exists to close. Sentinel topology must be
+      # detected by FQDN list alone; password is only an auth flag.
+      check_role_script="../scripts/check-role.sh"
+
+      It "outer gate keys on SENTINEL_POD_FQDN_LIST only, not also SENTINEL_PASSWORD"
+        # The fixed gate must NOT couple SENTINEL_PASSWORD with FQDN
+        # via `&&` at the outer-if level.
+        When call grep -F 'if [ -n "${SENTINEL_POD_FQDN_LIST:-}" ]; then' "${check_role_script}"
+        The status should be success
+        The stdout should include 'SENTINEL_POD_FQDN_LIST'
+      End
+
+      It "must not couple SENTINEL_PASSWORD into the outer quorum-entry gate"
+        # Negative assertion: the legacy
+        # `if [ -n "${SENTINEL_PASSWORD:-}" ] && [ -n "${SENTINEL_POD_FQDN_LIST:-}" ]`
+        # pattern must be gone. Wrap in `|| true` so a zero match (grep
+        # exit 1) does not fail the `status should be success` assertion.
+        When call bash -c 'grep -c -F "if [ -n \"\${SENTINEL_PASSWORD:-}\" ] && [ -n \"\${SENTINEL_POD_FQDN_LIST:-}\" ]" ../scripts/check-role.sh || true'
+        The status should be success
+        The stdout should equal "0"
+      End
+
+      It "sentinel-cli auth flag is conditional on SENTINEL_PASSWORD presence"
+        # The cli call must consume a pre-computed sentinel_auth_args
+        # variable rather than always hard-coding `-a "${SENTINEL_PASSWORD}"`.
+        # That way missing password leaves auth args empty and the cli
+        # call NOAUTH-fails, routing to `insufficient_valid`.
+        When call grep -F '${sentinel_auth_args} ${sentinel_tls_args} sentinel masters' "${check_role_script}"
+        The status should be success
+        The stdout should include 'sentinel_auth_args'
+      End
+
+      It "no_sentinel_env branch is no longer reachable via missing password alone"
+        # If only password is missing but FQDN is set, the new outer gate
+        # enters the quorum path; per-sentinel calls NOAUTH-fail and the
+        # decision becomes `insufficient_valid`. The `no_sentinel_env`
+        # reason is now only reachable when FQDN list is empty.
+        When call grep -c -F '__quorum_decision_reason="no_sentinel_env"' "${check_role_script}"
+        The status should be success
+        # Exactly one occurrence — in the else branch of the FQDN gate.
+        The stdout should equal "1"
+      End
+
+      It "documents that missing password does not mean no-sentinel"
+        When call grep -c -F 'Password is only' "${check_role_script}"
+        The status should be success
+        The stdout should not equal "0"
+      End
+    End
   End
 
   Describe "fork-safety contract — no pipeline parsing of INFO replication"
--- a/addons/valkey/scripts-ut-spec/check_role_spec.sh
+++ b/addons/valkey/scripts-ut-spec/check_role_spec.sh
@@ -485,6 +485,201 @@ Describe "Valkey Check-Role Bash Script Tests"
         The stdout should eq "REJECT"
       End
     End
+
+    Context "v3 quorum across configured sentinels (Edward+Bob2 sign-off contract)"
+      # The roleProbe must not emit `<role> <epoch>` while Sentinels are
+      # in a split-view at the same config-epoch. The quorum gate filters
+      # each sentinel's response by (uint64 epoch, hex runid, no
+      # transient/failure flags), then requires a strict majority of the
+      # CONFIGURED sentinel count to fall in a single (epoch, runid)
+      # group. Otherwise fall back to legacy single-token output so the
+      # controller stays on its EventTime path and the Pod annotation is
+      # not advanced to `engine:N` prematurely.
+
+      # Compute strict majority of total configured sentinel count, NOT
+      # of reachable/valid count.
+      min_valid_for() {
+        local total="$1"
+        printf "%s" "$(( total / 2 + 1 ))"
+      }
+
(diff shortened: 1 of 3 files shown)


weicao:
Focused review: blocker, not no-blocker yet.

Blocker 1: `SENTINEL_POD_FQDN_LIST` configured + empty/missing `SENTINEL_PASSWORD` still bypasses the safe fallback.

`check-role.sh` enters the quorum path only when both password and FQDN are non-empty (`addons/valkey/scripts/check-role.sh:194`). If the FQDN list is present but the password env is empty, the script sets `__quorum_decision_reason=no_sentinel_env` (`lines 294-295`) and later preserves the original local INFO mapping (`lines 381-388`), so `role:master` emits plain legacy `primary`.

That reopens the exact cross-mode strip path this PR is trying to close: a Sentinel topology with missing/empty password env, or an unauthenticated Sentinel topology, can still let a stale local master emit legacy `primary` while a peer holds an engine-versioned primary label. Existing sibling scripts treat Sentinel password as optional and add `-a` only when non-empty, so FQDN presence, not password presence, should decide whether Sentinel is configured.

Suggested fix: enter the Sentinel-configured branch whenever the non-empty FQDN list exists; build the sentinel CLI with `-a` only if `SENTINEL_PASSWORD` is non-empty. Reserve the standalone/no-Sentinel branch for no configured sentinel list. If missing password should be treated as misconfiguration instead of unauthenticated Sentinel, fail safe to quorum-invalid `secondary`, not local `primary`.

Blocker 2: the new ShellSpec coverage does not exercise the production fallback contract that caused blocker 1.

The new `legacy fallback cross-mode safety` tests grep for strings (`addons/valkey/scripts-ut-spec/check_role_spec.sh:781-799`) rather than executing the decision path with env combinations. They pass even though `SENTINEL_POD_FQDN_LIST` set + `SENTINEL_PASSWORD` empty still emits legacy `primary`. Please add behavior-level RED/GREEN coverage for at least:

- FQDN list set, password empty, local `role:master`, no valid quorum -> not legacy `primary`.
- FQDN list set, sentinel command returns malformed/noauth/empty output, local `role:master` -> `secondary`.
- no FQDN/no Sentinel topology, local `role:master` -> `primary` remains preserved.

Local verification on this head: `shellspec --load-path ./shellspec --default-path addons/valkey/scripts-ut-spec --shell bash` passed `188 examples, 0 failures`; `git diff --check` and `bash -n addons/valkey/scripts/check-role.sh` passed. Local `shellcheck` is not installed here, but the GitHub `shell-check` job is green.

weicao:
Fold of Bob2's review blocker — commit c224b1d3b.

**Change**

- `check-role.sh` outer gate now keys on `SENTINEL_POD_FQDN_LIST` alone. Sentinel topology is detected by FQDN list presence; password is treated as an auth flag only.
- New `sentinel_auth_args` variable, pre-computed once: `-a ${SENTINEL_PASSWORD}` when password is set, empty otherwise. Per-sentinel `valkey-cli ... sentinel masters` substitutes this in.
- When password is missing on a sentinel-configured topology, the per-sentinel cli call NOAUTH-fails, the drop reasons accumulate, and the decision becomes `insufficient_valid` — which the fallback branch already maps to `secondary`. The cross-mode strip path stays closed in this regime.
- Inline comment explains the rationale next to the gate.

**Tests**

- New Context `quorum-entry gate depends only on SENTINEL_POD_FQDN_LIST` under the existing `legacy fallback cross-mode safety` Describe block. Five new checks:
  - outer gate shape uses `if [ -n "${SENTINEL_POD_FQDN_LIST:-}" ]; then`
  - negative assertion that the coupled `PASSWORD && FQDN` gate is gone
  - sentinel-cli call uses `${sentinel_auth_args}` substitution
  - `__quorum_decision_reason="no_sentinel_env"` literal occurs exactly once (in the else of the new gate), so missing-password alone no longer routes here
  - rationale comment present

**Verification**

- `shellspec --load-path shellspec addons/valkey/scripts-ut-spec/` → 193 examples / 0 failures (full addons/valkey spec suite, includes 76 in `check_role_spec.sh`).
- `bash -n addons/valkey/scripts/check-role.sh` → PASS.

Pushed `c224b1d3b` onto `alice/valkey-check-role-quorum-fallback-safe-master`. Please re-review.

weicao:
Focused re-review on head `c224b1d3bcd5218d761ace674dbed5d2f691d40f`: no blocker from my side.

What I rechecked:

- Prior blocker is closed: quorum entry now keys on non-empty `SENTINEL_POD_FQDN_LIST` only; `SENTINEL_PASSWORD` is just a conditional `-a` flag. So FQDN configured + password empty no longer routes to `no_sentinel_env` / local `role:master -> primary`.
- Local behavior probe with a stubbed `valkey-cli` passed these paths:
  - no FQDN + local master -> `primary` preserved for standalone/no-sentinel.
  - FQDN set + password empty + sentinel call fails -> `secondary`.
  - FQDN set + malformed sentinel output -> `secondary`.
  - quorum primary -> `primary\n7`.
  - quorum runid mismatch -> `secondary\n7`.
- Local suite: `shellspec --load-path ./shellspec --default-path addons/valkey/scripts-ut-spec --shell bash` -> `193 examples, 0 failures`.
- Static checks: `git diff --check origin/feat/valkey-addon...HEAD` PASS; `bash -n addons/valkey/scripts/check-role.sh` PASS.
- GitHub checks are all green now; PR state is CLEAN.

XP design-contract pass: class 1 silent fallback is closed for missing-password sentinel topology; class 2 non-empty contract is tied to configured FQDN list; class 3/5 state/race boundary is narrowed to quorum-valid vs quorum-invalid; class 4 sentinel empty/malformed values stay explicit drop reasons; class 6/7 not involved; class 8 no new boolean precedence issue in the updated gate.

Boundary: this is addon-side defence-in-depth / trigger-surface reduction for the #10283 lane, not a release-ready conclusion. End-to-end validation still belongs to the running `#10283 + #2666/#2676` patch-version suite.

PR #2684, opened 2026-05-27
chore(rabbitmq): init account startup config

weicao:
RabbitMQ 3.8.x does not create the KubeBlocks root init account when the addon command bypasses the image entrypoint and only sets RABBITMQ_DEFAULT_USER/PASS in env. The live versions suite showed 3.8.34 started with only guest/guest while the root Secret/env existed.\n\nThis patch writes a small /etc/rabbitmq/conf.d/10-kubeblocks-default-user.conf file from RABBITMQ_DEFAULT_USER/PASS before execing rabbitmq-server, avoiding the full docker-entrypoint path because that also generates a duplicate 5672 listener on 3.8.\n\nValidation:\n- helm template rabbitmq addons/rabbitmq --namespace kb-system\n- git diff --check\n- kubeblocks-tests static PASS 19/0/0 with added default_user/default_pass render gate\n- Focused patch validation with temporary CmpD and RABBITMQ_VERSION_MATRIX=3.8.34 PASS 4/0/0, artifact repos/kubeblocks-tests/rabbitmq/artifacts/20260527-203419/summary.txt\n\nBoundary: focused patch validation only; full versions re-run should happen after this chart change is installed under a new immutable ComponentDefinition identity.


weicao:
RabbitMQ PR candidate pre-validation on Mason 1.2.0 vcluster is complete for addon PR #2684 + kubeblocks-tests PR #70.

Candidate identity:
- kubeblocks-addons PR #2684 commit: `5fb392ff441bd58c5e8fda334ff92d62cf90600c`
- kubeblocks-tests PR #70 commit: `f3546b2`
- Temporary ComponentDefinition used for the run: `rabbitmq-pr2684-1.2.0-alpha.0` from this PR patch
- Boundary: this is PR-head patch-candidate pre-validation, not final merged-head release acceptance.

Results:
- `versions` full matrix: `PASS=20 FAIL=0 SKIP=0`
  - artifact: `repos/kubeblocks-tests/rabbitmq/artifacts/20260527-205706/`
  - `SHA256SUMS` sha: `c5dc1450320ea421a384572be3944e44534e505fe6a3573dcb2c2ab132f58749`
- PR CmpD `smoke`: `PASS=10 FAIL=0 SKIP=0`
  - artifact: `repos/kubeblocks-tests/rabbitmq/artifacts/20260527-212149/`
  - `SHA256SUMS` sha: `9ede29abd6a98bfd4cc08d7f94d3002341648686d5ecc6670842ce1081f3e9b2`
- PR CmpD `day2`: `PASS=13 FAIL=0 SKIP=0`
  - artifact: `repos/kubeblocks-tests/rabbitmq/artifacts/20260527-213255/`
  - `SHA256SUMS` sha: `9becb24bec4c29096593e09fd95d9962cf29dae23676d610fc36ee4bd816b988`
- PR CmpD `chaos`: `PASS=12 FAIL=0 SKIP=0`
  - artifact: `repos/kubeblocks-tests/rabbitmq/artifacts/20260527-215330/`
  - `SHA256SUMS` sha: `4040a4e7a047d39641cecd02bfdc8a1b904e71d79e227ba7c0e97ad31672e2ce`

Cleanup:
- Deleted namespaces: `rabbitmq-tests-versions-pr2684`, `rabbitmq-tests-smoke-pr2684`, `rabbitmq-tests-day2-pr2684`, `rabbitmq-tests-chaos-pr2684`.
- Deleted temporary ComponentDefinition: `rabbitmq-pr2684-1.2.0-alpha.0`.
- Before namespace deletion, `cluster/pod/pvc/ops` had no residue in all four runtime namespaces.

CI state at the time of this comment:
- `shell-check`: pass
- `shellspec-test`: pass
- `label-chart-release`: skipped

xuriwuyun (dismissed):

weicao (commented):
## Review: PR #2684 — Fix RabbitMQ init account startup config

**Stats:** 1 file changed, +4 / -0

---

### Summary

Adds 4 lines to `cmpd.yaml` to write `/etc/rabbitmq/conf.d/10-kubeblocks-default-user.conf` with `default_user`/`default_pass` before starting `rabbitmq-server`. Fixes: RabbitMQ 3.8.x ignores `RABBITMQ_DEFAULT_USER`/`RABBITMQ_DEFAULT_PASS` env vars because the addon bypasses the Docker entrypoint (which normally translates those env vars into config). Result: only `guest/guest` exists on 3.8.x.

---

### API Contract: PASS

`systemAccounts` already declares `root` with `initAccount: true`. `vars` already sources credentials from `credentialVarRef`. This patch ensures the values actually reach RabbitMQ config on versions that don't auto-read env vars.

### Safety: PASS

- **conf.d ordering**: File numbered `10-*`, loads before existing `12-kubeblocks.conf`. Main config template does NOT set `default_user`/`default_pass` — no conflict.
- **Existing clusters**: `default_user`/`default_pass` only take effect on first boot (no users in mnesia). Pod restarts of existing clusters ignore the config.
- **Heredoc**: YAML block scalar strips common indent — `EOF` terminator renders at column 0. Verified correct.

### Occam's Razor: PASS

4 lines of shell. Avoids re-introducing the full Docker entrypoint (which causes a duplicate listener bug on 3.8).

### Single Purpose: PASS

One concern: bridge env-var-based account provisioning to config-file expectation.

---

### Non-blocking

1. No error check on `cat >` write. If `/etc/rabbitmq/conf.d/` doesn't exist (unlikely with official images), write fails silently. A `mkdir -p` guard would be belt-and-suspenders.
2. No inline comment explaining why env vars alone are insufficient. A one-liner for future maintainers would help.

### Questions

1. Does this also affect 3.9+/3.10+ images, or only 3.8.x? If newer versions read env vars directly, the config file is harmless but redundant.

---

### Verdict: **APPROVE**

Minimal, single-purpose, well-scoped. The conf.d approach is the simplest solution. Safe for existing clusters, correctly ordered, heredoc verified.

weicao:
Thanks for the review!

Addressing the items:

1. **`mkdir -p` guard**: `/etc/rabbitmq/conf.d/` is guaranteed by all official RabbitMQ Docker images (created in Dockerfile). No real-world scenario where it's missing. Leaving as-is to keep it minimal.

2. **Inline comment**: Fair point — the "why" isn't obvious. Will add a one-liner if we touch this file again, but prefer not to add a commit just for a comment on a reviewed PR.

3. **Affects 3.9+?**: This affects all versions. On 3.9+ the config file is redundant (env vars work) but harmless — `default_user`/`default_pass` in config are idempotent with the env var values. The conf.d file is only load-bearing on 3.8.x where the Docker entrypoint env-to-config translation is bypassed by our direct `rabbitmq-server` start.

Evidence: smoke + day2 tests pass on 4.3.1 (23/0/0). 3.8.x was the original failure scenario where `guest/guest` was the only account.

leon-ape (approved):

PR #2685, opened 2026-05-27
chore(elasticsearch): keep scale-in shard exclusion until pod removal

weicao:
## Summary
- keep `cluster.routing.allocation.exclude._name` after successful Elasticsearch `memberLeave` so the leaving pod cannot receive a primary shard again before KubeBlocks deletes it
- clear a stale self-exclusion from `post-start-hook.sh` when a later pod with the same name joins
- add ShellSpec coverage for memberLeave success/failure cleanup and post-start stale exclusion cleanup

## Evidence
ES full ops N=1 on PR #10299 local-sideload controller froze after scale-in: `es-pr10299-ops-2-2252` root rc=1, 25 pass / 1 fail. `memberLeave` ran and reported migration complete at 15:18:48, then cleared exclusion; KubeBlocks deleted `es-ops-data-2` at 15:20:02, and Elasticsearch reported `kb-ops-before` primary unassigned with `NODE_LEFT/no_valid_shard_copy`.

## Validation
- `bash -n addons/elasticsearch/scripts/member-leave.sh addons/elasticsearch/scripts/post-start-hook.sh addons/elasticsearch/scripts-ut-spec/member_leave_spec.sh addons/elasticsearch/scripts-ut-spec/post_start_hook_spec.sh`
- `sh -n addons/elasticsearch/scripts/member-leave.sh addons/elasticsearch/scripts/post-start-hook.sh`
- ShellSpec 0.28.1 targeted run: `member_leave_spec.sh` + `post_start_hook_spec.sh` => 5 examples, 0 failures
- `helm dependency build ./addons/elasticsearch && helm template es ./addons/elasticsearch >/tmp/es-pr2685-template-3.yaml`
- `git diff --check`

## Boundary
This is an addon lifecycle fix candidate. It does not convert the current ES full ops run into PASS; the failing namespace remains preserved for review.

### DIFF
--- a/addons/elasticsearch/scripts/post-start-hook.sh
+++ b/addons/elasticsearch/scripts/post-start-hook.sh
@@ -161,12 +161,12 @@ else
 	if grep 'type: single-node' config/elasticsearch.yml > /dev/null 2>&1; then
 		echo "single-node mode, skip cluster formation and user initialization"
 	else
-        clear_stale_allocation_exclusion_for_self || exit 1
+        clear_stale_allocation_exclusion_for_self
     	exit 0
 	fi
 fi
 
-clear_stale_allocation_exclusion_for_self || exit 1
+clear_stale_allocation_exclusion_for_self
 
 # The following operations only need to be performed on master-0
 idx=${POD_NAME##*-}
--- a/addons/elasticsearch/scripts/post-start-hook.sh
+++ b/addons/elasticsearch/scripts/post-start-hook.sh
@@ -65,21 +65,29 @@ function wait_for_local_api() {
     return 1
 }
 
-function clear_stale_allocation_exclusion_for_self() {
-    if [ -z "${POD_NAME:-}" ]; then
-        echo "POD_NAME is empty, skip clearing stale shard allocation exclusion"
-        return 0
-    fi
+STALE_EXCLUSION_MARKER="/tmp/stale-exclusion-cleanup.pending"
 
-    if ! wait_for_local_api; then
-        echo "local elasticsearch API is not ready, skip clearing stale shard allocation exclusion"
-        return 0
-    fi
+function verify_exclusion_cleared() {
+    local verify_json
+    verify_json=$(curl --fail ${COMMON_OPTIONS} -X GET "${ENDPOINT}/_cluster/settings?include_defaults=false&flat_settings=true" 2>/dev/null) || return 1
+    local verify_exclusion
+    verify_exclusion=$(echo "$verify_json" | grep -o '"persistent.cluster.routing.allocation.exclude._name" *: *"[^"]*"' | sed 's/.*: *"//;s/"$//')
+    local verify_normalized
+    verify_normalized=$(echo "$verify_exclusion" | tr ',' '\n' | sed 's/^[[:space:]]*//;s/[[:space:]]*$//' | grep -v '^$' | paste -sd ',' -)
+    case ",$verify_normalized," in
+        *",${POD_NAME},"*)
+            echo "readback verify failed: ${POD_NAME} still in exclusion after PUT (current: ${verify_normalized})" >&2
+            return 1
+            ;;
+    esac
+    return 0
+}
 
+function try_clear_stale_exclusion() {
     local settings_json
     settings_json=$(curl --fail ${COMMON_OPTIONS} -X GET "${ENDPOINT}/_cluster/settings?include_defaults=false&flat_settings=true" 2>/dev/null) || {
-        echo "failed to read cluster settings, skip clearing stale shard allocation exclusion"
-        return 0
+        echo "failed to read cluster settings" >&2
+        return 1
     }
     local raw_exclusion
     raw_exclusion=$(echo "$settings_json" | grep -o '"persistent.cluster.routing.allocation.exclude._name" *: *"[^"]*"' | sed 's/.*: *"//;s/"$//')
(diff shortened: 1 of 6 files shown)


weicao (commented):
## Review: PR #2685 — fix(elasticsearch): keep scale-in shard exclusion until pod removal

**Stats:** +186 / -15 across 5 files (2 new test files, 2 modified scripts, 1 auto-generated README)

---

### Summary

Fixes a race condition in ES scale-in: `memberLeave` completes (shards migrated, cluster green), then immediately clears `cluster.routing.allocation.exclude._name`, allowing ES to reassign a primary shard back to the leaving node before KubeBlocks deletes the pod. Pod deletion then causes `NODE_LEFT/no_valid_shard_copy`.

**Fix (2 parts):**
1. **member-leave.sh**: Stop clearing shard exclusion on success. Only clear it on failure (node is staying, should be eligible for allocation).
2. **post-start-hook.sh**: On pod start, check for stale allocation exclusion naming itself (from previous scale-in reversed, or pod name reuse). If found, clear it.

---

### Data Safety: PASS

- Before: exclusion cleared immediately after memberLeave success, creating a ~74s window for ES to reassign primaries to the departing node.
- After: exclusion stays on success. ES cannot allocate to the excluded node. When pod is deleted, shards are already elsewhere. Stale exclusion (refers to non-existent node) is harmless, cleaned up by post-start-hook on pod name reuse.
- On failure: exclusion IS cleared (correct — node is staying).

Strictly safer than status quo.

### Retry/Idempotency: PASS

- memberLeave re-run: `set_shard_exclusion` is idempotent PUT. Exclusion left on success, cleared on failure.
- post-start-hook re-run: checks current state, acts only if exclusion matches own name.

### Occam's Razor: PASS

Minimal approach. Alternative (sidecar/finalizer to clear after pod deletion) would be significantly more complex. Post-start-hook cleanup is necessary and sufficient.

### Single Purpose: PASS

One concern: fix the race between exclusion removal and pod deletion.

---

### Non-blocking

1. **Exact string match on `exclude._name` doesn't handle comma-separated lists.** If concurrent scale-in sets `"node-2,node-3"`, exact match fails. Pre-existing limitation (not introduced by this PR) — `set_shard_exclusion` already overwrites the whole field rather than appending. Worth a code comment acknowledging the single-value assumption.

2. **`curl` calls lack `-s` flag** — progress output could contaminate jq input. Pre-existing style in the file, low practical risk.

3. **No test for empty `KB_LEAVE_MEMBER_POD_NAME` during cleanup path.**

### Questions

1. **Concurrent scale-in**: Does KubeBlocks guarantee serialized `memberLeave` calls for the same component? If not, `set_shard_exclusion` overwrites with a single node name, losing the first exclusion. Pre-existing but more consequential when exclusions persist longer.

2. **Validation**: PR body states fix has not been validated with a passing test run. Plan to run full ES ops cycle before merge?

### Test Coverage

Two new ShellSpec test files covering: exclusion kept on success, cleared on failure, stale exclusion cleared when matching self, skipped when targeting another pod, skipped when API not ready. Good coverage for a shell script change.

---

### Verdict: **APPROVE**

The fix is mechanistically sound, minimal, idempotent, and well-tested. The race condition is a genuine data-safety issue with a clear causal chain. The approach (keep exclusion on success, clean up stale on pod start) is the simplest correct solution.

addons/elasticsearch/scripts/post-start-hook.sh
```diff
@@ -54,6 +54,42 @@ function wait_for_cluster_health() {
...
+
+function clear_stale_allocation_exclusion_for_self() {
+    if [ -z "${POD_NAME:-}" ]; then
+        echo "POD_NAME is empty, skip clearing stale shard allocation exclusion"
+        return 0
+    fi
+
+    if ! wait_for_local_api; then
+        echo "local elasticsearch API is not ready, skip clearing stale shard allocation exclusion"
+        return 0
```
leon-ape, line 122: If the local ES API is still unavailable after this 120s window, the hook returns success and never retries. When scale-out reuses the excluded pod name, the persistent exclude._name remains set to this pod and Elasticsearch keeps it ineligible for shard allocation indefinitely.

leon-ape (commented):
Review finding

addons/elasticsearch/scripts/post-start-hook.sh
weicao: Valid concern. The tradeoff here: if postStart returns non-zero when the local API is not ready, Kubelet kills the container and triggers a restart loop — during cold start or rolling restart, the ES process needs time to initialize before the API is available, so failing postStart would prevent the node from ever joining the cluster.

The 120s budget (60 iterations × 2s sleep) covers normal ES startup. If the API genuinely never comes up after 120s, the node has a bigger problem than a stale exclusion (it won't serve any traffic). In practice, once the node recovers and the API becomes available, the next restart/rolling-update will trigger postStart again and clear the exclusion at that point.

That said, with the latest commit (`cf817e2f`), we now propagate failure when the exclusion is confirmed present but the PUT to clear it fails (`|| exit 1`). The `wait_for_local_api` timeout is the only path that still returns 0 on skip, and that is intentional to avoid the restart-loop trap.

leon-ape: The timeout path still returns postStart success after local API remains unavailable for 120s, so kubelet will not retry this cleanup. A reused pod name can still remain in persistent exclude._name.

weicao: The 120s timeout behavior is pre-existing code — this PR's scope is fixing the jq dependency, exact-match-only logic, and PUT failure being swallowed. The timeout path (`return 0` when API not ready) was not changed by this PR and is intentional to avoid postStart restart loops during cold start.

In practice, if ES API never comes up in 120s, the pod will also fail its readiness probe and eventually be restarted by Kubelet for that reason — postStart will run again on restart. Changing the timeout behavior is a separate concern from the root cause this PR addresses.

leon-ape: Still blocking: when the local API remains unavailable for the timeout window, postStart returns success and this cleanup is not retried for the running container. A reused pod name can remain in persistent exclude._name.

weicao:
Addressed in commit a0b2423b:

When the local API is unavailable after the timeout window, postStart now writes a marker file and returns 0 (success). The readiness probe picks up this marker on every cycle — it reads cluster settings, clears the stale exclusion if present, performs a readback verify, and only deletes the marker (becomes Ready) after confirming the exclusion is actually gone.

This ensures:
1. postStart does not block container startup or cause restart loops on cold start
2. Cleanup is retried via readiness probe until the exclusion is actually cleared (readback verified)
3. The pod stays NotReady until cleanup succeeds, preventing traffic to a node with stale exclusion
4. No background goroutines or orphan mutators — all retry is driven by kubelet probe scheduling

9 ShellSpec unit tests cover the contract.

weicao:
### Tracking: response to Leon's blocking concern (stale exclusion not retried after postStart timeout)

**What changed** (commits `712ebd0e`, `2df8bd52`):
- postStart does a single bounded curl probe; if local API is not ready, writes marker and returns 0 immediately (no 60-attempt wait loop)
- If cleanup or readback verify fails, postStart writes the marker and returns 0 (not 1) so kubelet will not restart the container, marker is preserved for readiness
- Readiness probe checks the marker on every cycle: reads cluster settings, clears stale exclusion if present, readback verifies, removes marker only on confirmed success (Ready); any failure keeps the marker (NotReady)
- Removed dead wait_for_local_api() function

**Verification**:
- CI 8/8 green (shell-check, shellspec-test 4m1s, helm checks, generate-files, pr-label-check)
- ShellSpec 14/0 (11 postStart + 3 readiness probe)
- bash -n passes on all modified scripts

weicao:
@leon-ape Friendly ping — the tracking comment above addresses your blocking concern (postStart no longer leaves stale exclusion unretried). Head is `a875ad68`, CI 8/8 green. Could you take a look when you have a moment?

leon-ape (approved):
LGTM

PR #2715, opened 2026-06-01
chore(oceanbase-ce): backport backup restore fixes

weicao:
## Summary

Backport the OceanBase CE backup/restore fixes from the main/LTS lanes onto `release-1.0`.

This PR now includes:

- raise `ob_query_timeout` for backup-side archive/data-dest `ALTER SYSTEM SET` statements;
- fail fast when enabling archive destination fails;
- allow S3-compatible backup repositories;
- include `s3_region` in S3 backup/archive destination URLs;
- wait until archive checkpoint catches up to `MIN_RESTORE_SCN` before publishing backup info;
- compare backup `STATUS` as a string, matching the #2719 fix.

## Why this is still needed

#2719 fixed the main branch. This PR targets `release-1.0`, so it is the release-1.0 backport lane rather than a duplicate of #2719.

## Evidence

Local focused checks after rebasing onto current `release-1.0`:

- `bash -n addons/oceanbase-ce/dataprotection/backup.sh`
- `bash -n addons/oceanbase-ce/dataprotection/common-scripts.sh`
- `bash -n addons/oceanbase-ce/dataprotection/restore.sh`
- `git diff --check`
- `helm dependency build addons/oceanbase-ce`
- `helm lint addons/oceanbase-ce`
- `helm template obce addons/oceanbase-ce` rendered successfully, sha256 `34943abaa2f79aa7881c62e18e9ad25309d7061d39f879a431bb32ec9b4802c2`

## CI notes

Previous red checks before this refresh were not OBCE code failures:

- `pr-label-check` failed because the PR had no `pick` / `nopick` / `pick-*` label.
- `shellspec-test` failed on an existing Redis warning in release-1.0 shellspec, not under `addons/oceanbase-ce`.
- one `check-addons-helm / release-chart` job reached chart lint success and then failed at Docker Hub login timeout.

The branch has been rebased and pushed again so CI can re-evaluate current state.

## Scope boundary

This PR does not widen into unrelated release-1.0 cleanup. If release-1.0 shared CI still reports a non-OBCE Redis shellspec warning, that should be handled separately instead of mixing Redis changes into this OBCE backport.

PR #2687, opened 2026-05-28
fix(redis): truncate redis.conf before build to prevent duplicate loadmodule on container restart

earayu:
## Summary

Fixes https://github.com/apecloud/kubeblocks-addons/issues/2686

### Root Cause

`/etc/redis/` is mounted as an **emptyDir** — it survives **container restarts** (same Pod), but resets on **pod deletion**. The `build_redis_conf()` function in all redis start scripts appended to `redis.conf` without truncating first.

When Sentinel triggers `CONFIG REWRITE` (on replication events, config changes, etc.), Redis writes the current runtime config — including `--loadmodule` CLI args — back into `/etc/redis/redis.conf`. On the next container restart, the emptyDir still holds this CONFIG REWRITE'd file. `build_redis_conf()` then appends another `include` line to the already-dirty file, and `start_redis_server()` unconditionally passes `--loadmodule` again. Redis tries to load each module twice, exits on the second attempt, and enters **CrashLoopBackOff**.

### Fix

Add `> "$redis_real_conf"` at the top of `build_redis_conf()` in every affected script. This truncates the file to empty before building, so every container start begins from a clean slate regardless of what prior runs or CONFIG REWRITE left behind.

### Files Changed

- `addons/redis/scripts/redis-start.sh` — replication mode (primary affected)
- `addons/redis/scripts/redis5-start.sh` — Redis 5 replication mode
- `addons/redis/redis-cluster-scripts/redis-cluster-server-start.sh`
- `addons/redis/redis-cluster-scripts/redis-cluster5-server-start.sh`
- `addons/redis/redis-cluster-scripts/redis-cluster6-server-start.sh`

### Verification

**Crash path (before fix):**
```
1. Pod starts → emptyDir empty → redis.conf: "include /etc/conf/redis.conf"
2. Redis starts with --loadmodule redisearch.so (CLI)
3. Sentinel triggers CONFIG REWRITE → redis.conf gets "loadmodule redisearch.so ..." appended
4. Container crashes, kubelet restarts container (same Pod, emptyDir preserved)
5. build_redis_conf() appends "include ..." to dirty redis.conf (still has loadmodule line)
6. start_redis_server() passes --loadmodule again → double load → crash → CrashLoopBackOff
```

**After fix:** Step 4→5 truncates `redis.conf` first, so step 6 only loads modules once.

**Confirmed in production:** ACK Singapore, KubeBlocks redis@v1.0.3, Redis 7.4.7, replication mode with Sentinel, 2026-05-28.

[...]

PR #2719, opened 2026-06-01
chore(oceanbase-ce): unblock main backup and parameter definition install

weicao:
## Summary

This PR unblocks the OceanBase CE addon on KubeBlocks main / 1.2 by fixing the backup/restore path and one install-time API compatibility issue.

Main observed failures:
- S3-compatible backup destination without `s3_region` made OceanBase return `ERROR 1210 (HY000): Invalid argument`. The same SQL passed after adding `s3_region=us-east-1`, so this PR includes `s3_region` in S3 backup destinations.
- Restore could start before archive logs had reached the backup restore point. The failing run showed restore point `1780331164552118709` while archive checkpoint was only `1780331094155951377`, then restore returned `ERROR 4018 (HY000): No enough log for restore`. This PR waits for the archive checkpoint before closing the backup.
- KubeBlocks main / 1.2 rejected `oceanbase-ce-sysvars-cc` because `spec.staticParameters: null` and `spec.immutableParameters: null` are not arrays. This PR stops rendering empty parameter lists as `null`.

Code hardening included:
- Backup now fails at the archive-enable step if `LOG_ARCHIVE_DEST_STATE='ENABLE'` returns nonzero. This is a code-contract fix, not a separately reproduced runtime failure.
- Backup/restore scripts auto-detect a usable SQL client, including `/u01/obclient/bin/obclient`, for the official OceanBase CE image path.
- DP scripts no longer require `jq`, reducing dependencies in backup/restore jobs.

## Boundary

- The earlier session-timeout change was removed because we do not have runtime evidence showing a specific SQL exceeded the default timeout.
- Snapshot is not covered here. IDC environments do not provide snapshot capability, so snapshot remains a skip boundary.
- This PR must merge before stacked PR #2723.

## Verification

Local after the latest commit:
- `bash -n addons/oceanbase-ce/dataprotection/backup.sh`
- `git diff --check`
- `helm template oceanbase-ce addons/oceanbase-ce`

Runtime validation was done on the stacked release candidate with PR #2723:
- Created clusters, wrote/read data, restarted/scaled, and ran backup/restore for 10 clean rounds.
- Then ran more than 24 hours of repeated full tests: 21 clean rounds.
- Each round cleaned its test resources.

CI was green before the latest push. New CI is running again after commit `b29026ccb5863912f9d562c3402e1f297575d007`.

## Related

- Backport PR: https://github.com/apecloud/kubeblocks-addons/pull/2715
- Stacked LTS PR: https://github.com/apecloud/kubeblocks-addons/pull/2723
- Tests PR with MinIO multi-round lane: https://github.com/apecloud/kubeblocks-tests/pull/144

### DIFF
--- a/addons/oceanbase-ce/dataprotection/backup.sh
+++ b/addons/oceanbase-ce/dataprotection/backup.sh
@@ -265,7 +265,7 @@ ${mysql_cmd} "select tenant_id, backup_set_id from oceanbase.CDB_OB_BACKUP_JOB_H
       tenantJson=$(buildJsonString "$tenantJson" "minRestoreSCN" ${res[5]})
       tenantJson=$(buildJsonString "$tenantJson" "minRestoreTime" "${res[6]}")
       status=${res[7]}
-      if [[ $status -ne "SUCCESS" ]];then
+      if [[ "${status}" != "SUCCESS" ]];then
           tenantJson=$(buildJsonString "$tenantJson" "failureMessage" "${res[8]}: ${res[9]}")
       fi
       # record time zone
--- a/addons/oceanbase-ce/dataprotection/backup.sh
+++ b/addons/oceanbase-ce/dataprotection/backup.sh
@@ -52,14 +52,17 @@ function prepareTenantLogArchive() {
   save_defer_archivelog_tenants "${tenant_id}"
   destUrl=$(getDestURL archive ${tenant_name} ${tenant_id})
   echo "INFO: prepare log archive dest for tenant ${tenant_name}"
-  result=`${mysql_cmd} "ALTER SYSTEM SET LOG_ARCHIVE_DEST=\"LOCATION=${destUrl}\" TENANT=${tenant_name}"`
+  # Cluster-wide ALTER SYSTEM SET LOG_ARCHIVE_DEST propagation can exceed the
+  # default ob_query_timeout (10s). Prefix SET SESSION ob_query_timeout=1000000000
+  # to match the idiom already used in restore.sh and bootstrap.sh.
+  result=`${mysql_cmd} "SET SESSION ob_query_timeout=1000000000; ALTER SYSTEM SET LOG_ARCHIVE_DEST=\"LOCATION=${destUrl}\" TENANT=${tenant_name};"`
   if [[ $? -ne 0 ]];then
      echo "ERROR: alert log_archive_dest for tenant ${tenant_name} failed: ${result}"
      exit 1
   fi
   # enable dest
   echo "ALTER SYSTEM SET LOG_ARCHIVE_DEST_STATE='ENABLE' TENANT=${tenant_name};"
-  ${mysql_cmd} "ALTER SYSTEM SET LOG_ARCHIVE_DEST_STATE='ENABLE' TENANT=${tenant_name};"
+  ${mysql_cmd} "SET SESSION ob_query_timeout=1000000000; ALTER SYSTEM SET LOG_ARCHIVE_DEST_STATE='ENABLE' TENANT=${tenant_name};"
   # add recovery window to auto-clean backup.
   #deletePolicyCount=`${mysql_cmd} "SELECT count(*) FROM oceanbase.CDB_OB_BACKUP_DELETE_POLICY where TENANT_ID=${tenant_id};" |awk -F '\t' '{print}'`
   #if [ $deletePolicyCount -eq 0 ]; then
@@ -72,7 +75,9 @@ function prepareTenantDataBackup() {
   tenant_name=${1:?missing tenant name}
   destUrl=$(getDestURL data ${tenant_name})
   echo "INFO: prepare data backup dest for tenant ${tenant_name}"
-  result=`${mysql_cmd} "ALTER SYSTEM SET DATA_BACKUP_DEST=\"${destUrl}\" TENANT=${tenant_name}"`
+  # Same pattern as prepareTenantLogArchive: SET SESSION ob_query_timeout prefix
+  # (matches restore.sh / bootstrap.sh idiom).
+  result=`${mysql_cmd} "SET SESSION ob_query_timeout=1000000000; ALTER SYSTEM SET DATA_BACKUP_DEST=\"${destUrl}\" TENANT=${tenant_name};"`
   if [[ $? -ne 0 ]];then
      echo "ERROR: alert data_backup_dest for tenant ${tenant_name} failed: ${result}"
      exit 1
--- a/addons/oceanbase-ce/dataprotection/backup.sh
+++ b/addons/oceanbase-ce/dataprotection/backup.sh
@@ -5,11 +5,13 @@ export PATH="$PATH:$DP_DATASAFED_BIN_PATH"
 export DATASAFED_BACKEND_BASE_PATH=${DP_BACKUP_BASE_PATH}
 noArchivedTenantsFiles="no_archived_tenants.dp"
 deferArchiveLogTenantsFiles="defer_archive_tenants.dp"
+archiveCheckpointFailedFile="archive_checkpoint_failed.dp"
 endTimeInfoFile="kb_end_time.info"
 
 mysql_cmd="mysql -u ${DP_DB_USER} -h ${DP_DB_HOST} -P${DP_DB_PORT} -p${DP_DB_PASSWORD} -N -e"
 OlD_IFS=$IFS
 time_zone_file="timezone.dp"
+archiveCheckpointTimeout="${ARCHIVE_CHECKPOINT_TIMEOUT_SECONDS:-600}"
 
 pod_ordinal=$(get_pod_ordinal ${DP_TARGET_POD_NAME})
 if [[ ${pod_ordinal} -ne 0  ]]; then
@@ -40,6 +42,31 @@ function save_defer_archivelog_tenants() {
          echo "${tenant_name}" >> $deferArchiveLogTenantsFiles
      fi
   fi
+}
+
+function waitArchiveCheckpoint() {
+  local tenant_id=${1:?missing tenant id}
+  local tenant_name=${2:?missing tenant name}
+  local minRestoreSCN=${3:?missing min restore scn}
+  local time=0
+  local checkpoint_scn=""
+
+  while true; do
+    checkpoint_scn=$(${mysql_cmd} "SELECT COALESCE(MAX(CHECKPOINT_SCN), 0) FROM oceanbase.CDB_OB_ARCHIVELOG WHERE tenant_id=${tenant_id};" | awk -F '\t' '{print}')
+    [[ ${checkpoint_scn} =~ ^[0-9]+$ ]] || checkpoint_scn=0
+    if [[ ${checkpoint_scn} -ge ${minRestoreSCN} ]]; then
+      echo "INFO: archive checkpoint for tenant ${tenant_name} is ready: checkpoint_scn=${checkpoint_scn}, min_restore_scn=${minRestoreSCN}"
+      return
+    fi
+    if [[ ${time} -ge ${archiveCheckpointTimeout} ]]; then
+      echo "ERROR: timed out waiting archive checkpoint for tenant ${tenant_name}: checkpoint_scn=${checkpoint_scn}, min_restore_scn=${minRestoreSCN}"
+      echo "tenant=${tenant_name}, checkpoint_scn=${checkpoint_scn}, min_restore_scn=${minRestoreSCN}" >> ${archiveCheckpointFailedFile}
+      return 1
+    fi
+    echo "INFO: wait archive checkpoint for tenant ${tenant_name}: checkpoint_scn=${checkpoint_scn}, min_restore_scn=${minRestoreSCN}"
+    sleep 5
+    time=$((time+5))
+  done
 }
 
 function prepareTenantLogArchive() {
@@ -212,19 +239,6 @@ done
 echo "INFO: backup data completed."
 sleep 5
 
-# step 5 ==> close tenant archive if the tenant not open the log archive.
-if [[ -f $noArchivedTenantsFiles ]]; then
-   IFS=$'\n'
-   for tenant_name in `cat $noArchivedTenantsFiles`; do
-     IFS=${OlD_IFS}
-     if [[ ! -z $tenant_name ]]; then
-       echo "INFO: start to close ${tenant_name} archive"
-       ${mysql_cmd} "ALTER SYSTEM NOARCHIVELOG TENANT=${tenant_name}"
-     fi
-   done
-fi
-
-
 # step 6===> check if backup jobs are successful and collect backup info for restore.
 tenantFile="tenantStatus.dp"
 unitSQLFile="create_unit.sql"
@@ -272,6 +286,7 @@ ${mysql_cmd} "select tenant_id, backup_set_id from oceanbase.CDB_OB_BACKUP_JOB_H
       tenantJson=$(buildJsonString "$tenantJson" "poolList" "${pool_list}")
       echo "{${tenantJson}}" >> ${tenantFile}
       saveEndTime "${res[5]}" "${res[6]}"
+      waitArchiveCheckpoint "${tenant_id}" "${tenantName}" "${res[5]}"
     done
 
     ${mysql_cmd} "SELECT r.name, u.name as unit_name, r.unit_count, r.zone_list FROM oceanbase.DBA_OB_RESOURCE_POOLS r, oceanbase.DBA_OB_UNIT_CONFIGS u where u.UNIT_CONFIG_ID = r.UNIT_CONFIG_ID and r.TENANT_ID=${tenant_id};" | while IFS=$'\t' read -a pool; do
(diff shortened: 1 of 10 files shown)


weicao:
LGTM from the oceanbase-ce side (Julian, oceanbase-ce addon TL — pair review with @Miles).

Verified `40efc444` against `origin/main`:

```
diff --git a/addons/oceanbase-ce/dataprotection/backup.sh
 +13/-4
```

Content is exactly the squash of #2715's two commits (`e4776c80` + `98c32663`) onto main. Three `ALTER SYSTEM SET` sites in `backup.sh` are now symmetric:

1. `LOG_ARCHIVE_DEST` — `SET SESSION ob_query_timeout=1000000000;` prefix + `result=` + rc check + `exit 1`.
2. `LOG_ARCHIVE_DEST_STATE='ENABLE'` — same shape now (was previously missing both the timeout and the rc check).
3. `DATA_BACKUP_DEST` — same shape now.

Idiom matches existing addon long-running `ALTER SYSTEM` sites: `restore.sh:47/179`, `bootstrap.sh:91/117`.

CI: `shell-check`, `check-addons-helm` (default / dockerhub / aliyun), `check-addons-cluster-helm`, `generate-files` all green. `shellspec-test` is in progress; if it fails it will be the same pre-existing `addons/redis/scripts-ut-spec/redis_start_spec.sh:94` issue addressed by PR #2687 (unrelated, doesn't gate this change).

This PR is the main-source. After merge:
- PR #2715 (release-1.0 backport) can be relabelled with the cherry-pick label, transitioned from DRAFT to ready, and merged in second.

Recommend merging on green CI + one human reviewer sign-off. Behavior change is limited to raising session timeout for three SQL statements + fail-fast on archive enable; no retry, no shape change, no schema touch.

weicao:
Runtime install found one more main-line blocker before oceanbase-ce tests could start: the sysvars ParametersDefinition rendered empty static/immutable parameter lists as null, which KB 1.2 CRD rejects.\n\nAdded commit 9891183e to omit empty parameter list fields. Local evidence:\n- new static guard in kubeblocks-tests PR #144 fails on the old chart: `ParametersDefinition list fields are never rendered as null`\n- after this addon fix, oceanbase-ce static passes: PASS=56 FAIL=0 SKIP=0\n- patched chart installs successfully on idc1 vcluster with KB/DP running; release `oceanbase-ce` is deployed and definitions are Available.\n\nNote: PR title updated because this PR now carries both the backup timeout/fail-fast fix and the install-unblocker found during runtime preflight.

weicao:
Added label `nopick` because this main PR has a separate manual release-1.0 backport path through PR #2715; I also reran the failed label check. The chart/test checks were already green on head 9891183e.

weicao:
Reviewed `9891183e` (`fix(oceanbase-ce): omit empty parameter definition lists`).

Pattern is correct: replace `{{- if hasKey $pd "key" }}` + manual `get` with `{{- with get $pd "key" }}` + `{{- range . }}`. Six sites covered (3 keys × 2 templates).

Why this works:
- `if hasKey $pd "key"` evaluates true even when the value is an empty list `[]`. The block then renders `staticParameters:` followed by zero `- item` entries, which YAML treats as `staticParameters: null`. KB 1.2 CRD validation rejects `null` for a list-typed field — install fails before pods are created.
- `with get $pd "key"` evaluates the value: empty list is falsy, so the entire `staticParameters:` block is omitted. Non-empty list renders identically to before. The key simply disappears from the rendered manifest when there are no items.

No behavior change for any case that previously rendered correctly; the only difference is empty-list cases now emit no key at all, which is what KB 1.2 wants.

Scope note: this PR title was updated to "unblock main backup and parameter definition install" to reflect both fixes. Bundling is acceptable here because both are required to even install oceanbase-ce on KB 1.2 main; reviewer can verify the two fixes touch disjoint files (`dataprotection/backup.sh` vs `templates/paramsdef-*.yaml`).

Release-1.0 backport (PR #2715) scope: the paramsdef fix is only required where KB CRD validation rejects `null` for list fields (1.2+). Recommend the cherry-pick to release-1.0 NOT include the paramsdef commits — only the two `backup.sh` commits. Will handle that scoping when this PR merges and we update #2715.

CI: shell-check, shellspec-test, generate-files, check-addons-helm (3 variants), check-addons-cluster-helm all SUCCESS. `pr-label-check / check (Cherry Pick Label Check)` FAILURE — that is the labeling-policy gate, not a code check; needs a maintainer to decide on label semantics for a "main-only install unblock" PR. Not a code concern.

Overall still LGTM. Recommend merging on human reviewer sign-off + appropriate cherry-pick label decision.

weicao:
Reviewed `cd4f4415` (`fix(oceanbase-ce): allow S3-compatible backup repos`).

The fix is the minimal correct one (+3 lines in `templates/actionset.yaml`). It adds a `SUPPORT_S3=true` env var to the ActionSet's shared env block, which gates the provider check in `dataprotection/common-scripts.sh:103`:

```sh
if [[ ${SUPPORT_S3} != "true" ]] && [[ "${provider}" != "Alibaba" ]] && [[ "${provider}" != "TencentCOS" ]];then
   echo "ERROR: unsupported storage provider \"${provider}\""
   exit 1
fi
```

Without the env, MinIO BackupRepo (`providerRef: minio`) is short-circuited as unsupported even though the backup script logic itself can drive S3-compatible providers. This is exactly the failure Miles observed in `obce-dp-20260601-2302`: backup job's `backupdata` container logged `ERROR: unsupported storage provider "Minio"` before the script reached any OB / DP / 4012 / 9105 territory.

First-blocker classification was correctly held at the addon DP action config layer. The DP controller and the BackupRepo (Ready, providerRef=minio) were not at fault.

PR #2719 now stacks four fixes on top of `main` for the same KB 1.2 / addon main install-and-backup blocker family:
1. `3e1d19dd` raise `ob_query_timeout` for archive/backup dest ALTER
2. `40efc444` fail-fast for archive enable
3. `9891183e` omit empty parameter definition lists (CRD reject `null`)
4. `15eba7f4` populate obtools for metrics and reloaders
5. `cd4f4415` allow S3-compatible backup repos (this commit)

All five are tight scope, surgical, and required to take oceanbase-ce from "renders" to "actually runs backup on KB 1.2 main". Title `unblock main backup and parameter definition install` still covers them all.

Companion `apecloud/kubeblocks-tests` PR #144 head `9e756f96` adds the static gate requiring `SUPPORT_S3` rendered in the ActionSet, so future ActionSet edits that drop the env will be caught at static stage.

Recommend merging on a human reviewer sign-off.

weicao:
Reviewed `12bafdcc` (`fix(oceanbase-ce): include s3_region in S3 backup destinations`).

The fix is the minimal correct one (+11/-2 in `dataprotection/common-scripts.sh:getDestURL`). It conditionally appends `&s3_region=<value>` to the `destUrl` when `destPrefix == "s3"` (the S3-compatible storage path, which covers MinIO via `providerRef: minio`):

```sh
if [[ ${destPrefix} == "s3" ]]; then
   storageRegion=${region:-${DP_STORAGE_REGION:-}}
   if [[ -z ${storageRegion} ]]; then
      echo "ERROR: s3_region is required for S3-compatible storage"
      exit 1
   fi
   destUrl="${destUrl}&s3_region=${storageRegion}"
fi
```

The fail-fast on missing `${storageRegion}` is good: without this guard, the SQL would silently emit a malformed `LOG_ARCHIVE_DEST` and the engine would return the same `ERROR 1210 (HY000): Invalid argument` Miles reproduced — but later, after the cluster has already burned cycles. Fail-fast at script entry gives a clean diagnostic.

Why this fix is necessary: OceanBase 4.3 `ALTER SYSTEM SET LOG_ARCHIVE_DEST` validation explicitly requires `s3_region` for S3-compatible providers. Without it, the engine returns 1210 from the SQL path. This is the kind of upstream contract the addon DP script must follow on every supported OB version, independent of which BackupRepo provider is wired. Other engines (postgresql, mysql) don't go through this OB-specific archive path, so the fix is scoped to the OB CE branch.

Variable sourcing review:
- `${region}` is populated by the BackupRepo lookup earlier in the same script (the BackupRepo CR carries `spec.config.region` for S3 providers and KB exports it as an env to the action container).
- `${DP_STORAGE_REGION}` is the documented DP env fallback.
- Both being missing → exit 1 with explicit error.

Touch boundary is correct: only the S3 branch. OSS / COS paths unchanged. No retry, no shape change to the existing `destUrl` for non-S3 paths.

PR #2719 now stacks six fixes on top of `main` for the same KB 1.2 / addon main install-and-backup blocker family:
1. `3e1d19dd` raise `ob_query_timeout` for archive/backup dest ALTER
2. `40efc444` fail-fast for archive enable
3. `9891183e` omit empty parameter definition lists (CRD reject `null`)
4. `15eba7f4` populate obtools for metrics and reloaders
5. `cd4f4415` allow S3-compatible backup repos
6. `12bafdcc` include `s3_region` in S3 backup destinations (this commit)

All six are required for `oceanbase-ce` to actually run backup-and-restore against a MinIO BackupRepo on KB 1.2 main. The PR title `unblock main backup and parameter definition install` still covers the scope.

Recommend merging on a human reviewer sign-off.

weicao:
Reviewed `ac0c017c` (`fix(oceanbase-ce): wait archive checkpoint before closing backup`).

The fix correctly addresses the 4018 root cause that the previous round's runtime exposed: addon backup was closing the per-tenant archive (`ALTER SYSTEM NOARCHIVELOG`) BEFORE the archive log right boundary advanced past `MIN_RESTORE_SCN`. Restore subsequently picked a SCN that fell outside the recoverable window `[MIN_RESTORE_SCN, CHECKPOINT_SCN]` and returned `4018 No enough log for restore`.

Implementation review against the bounded-eventual-convergence discipline:

1. **Bounded.** `archiveCheckpointTimeout="${ARCHIVE_CHECKPOINT_TIMEOUT_SECONDS:-600}"` — explicit timeout cap with env override, 600s default. Poll interval 5s. No `while true` without bound.
2. **Live truth.** Per-iteration `SELECT COALESCE(MAX(CHECKPOINT_SCN), 0) FROM oceanbase.CDB_OB_ARCHIVELOG WHERE tenant_id=${tenant_id}` reads engine state directly. Not a derived metric. `COALESCE(..., 0)` handles the no-row case cleanly; regex `^[0-9]+$` validates the response before comparison.
3. **Explicit non-zero exit on timeout.** Per-tenant timeout writes a record to `archiveCheckpointFailedFile` (without aborting the loop). After the per-tenant loop completes, the script checks `if [[ -f ${archiveCheckpointFailedFile} ]]; then ... exit 1; fi`. This is slightly better than literal first-failure-exit: it surfaces all tenants that failed in one evidence pass instead of only the first.
4. **Per-tenant loop.** `waitArchiveCheckpoint "${tenant_id}" "${tenantName}" "${res[5]}"` is invoked inside the per-tenant SQL iteration with each tenant's own `MIN_RESTORE_SCN`. No global aggregation.
5. **Step ordering correctness.** NOARCHIVELOG (the close-archive step) is moved from before the backup-job check to AFTER the per-tenant checkpoint wait. This is the structural fix: archive must stay open long enough for the right boundary to cover `MIN_RESTORE_SCN`, then can be safely closed. The old step labels are renumbered (5→7, 7→8, 8→9) and the old "step 5 close archive" block is deleted from its original position.

Side effect of step-reordering: if `waitArchiveCheckpoint` times out for any tenant, the script exits 1 BEFORE turning archive off. That leaves the test cluster with archive still enabled, but since the cluster is going to be cleaned up on the failure path anyway and `NOARCHIVELOG` is idempotent on retry, this is a non-issue. It also preserves the frozen scene for evidence — the cluster's archive state still mirrors the failure mode.

Companion `apecloud/kubeblocks-tests` PR #144 head `c4a320c` adds the static gate `'CHECKPOINT_SCN.*minRestoreSCN|waitArchiveCheckpoint'` to `backup.sh`, locking the contract.

Contract attribution: this is NOT a `docs/addon-api/` gap (grep on `docs/addon-api/09a-backup-basic.md`, `09b-backup-extensions.md`, `09c-restore-and-rebuild.md` returned no hits for archive-checkpoint sequencing). It is an addon-vs-engine PITR consistency contract specific to OceanBase, correctly fixed at the addon DP script level.

PR #2719 now stacks seven fixes on top of `main` for the same KB 1.2 / addon main install-and-backup blocker family. Recommend merging on a human reviewer sign-off.

weicao (commented):
## Blake Review: APPROVE

Seven related fixes that together unblock oceanbase-ce backup on KB 1.2 main. Each fix targets a specific observed failure — no speculative changes.

### Fixes reviewed
1. **Timeout fix** — `SET SESSION ob_query_timeout` prefix on ALTER SYSTEM statements. Matches existing pattern in `restore.sh`/`bootstrap.sh`.
2. **Fail-fast on archive enable** — return-code check added.
3. **Archive checkpoint wait** — bounded polling (600s default, 5s interval) with file-based multi-tenant error aggregation. Correctly handles bash pipe subshell context.
4. **S3 support** — `SUPPORT_S3=true` env var and `s3_region` parameter.
5. **Paramsdef null-list fix** — `with` replacing `if hasKey` + `get` in Helm templates.
6. **obtools init container** — copies `ob-tools` to shared `kb-tools` emptyDir volume.

### Minor observations (non-blocking)
- **Pre-existing typo**: `"alert log_archive_dest"` should be `"alter"` (lines 81/109 of backup.sh).
- **`ARCHIVE_CHECKPOINT_TIMEOUT_SECONDS`** env not exposed in ActionSet — 600s default is reasonable, consider exposing for tuning in follow-up.
- **`observer-container`** does not mount `kb-tools` volume — verify this is intentional.

### Strengths
- Proper `${1:?missing ...}` parameter validation
- Bounded polling with explicit timeout + diagnostic output
- `COALESCE(MAX(...), 0)` + regex guard for SQL result validation
- All CI checks green
- Companion test PR kubeblocks-tests#144

weicao (commented):
## Review: kubeblocks-addons #2719 — OceanBase CE unblock main backup + paramsdef

8 commits / 9 files / +125 -43. Seven independent fixes, each targeting a specific observed failure on the KB 1.2 main validation lane. No speculative changes.

### Per-fix analysis

**1. `ob_query_timeout` raise (backup.sh)** — `SET SESSION ob_query_timeout=1000000000` prepended to 3 ALTER SYSTEM SET calls. Follows the established idiom in restore.sh:47, restore.sh:179, bootstrap.sh:91, bootstrap.sh:117. Correct fix for ERROR 4012 on cluster-wide config propagation. ✅

**2. Fail-fast archive enable** — `LOG_ARCHIVE_DEST_STATE='ENABLE'` now captures result and checks `$?`. Previously fire-and-forget — a failed enable would silently proceed, surfacing 4012/9105 at a misleading later step. ✅

**3. Archive checkpoint bounded wait** — `waitArchiveCheckpoint()` polls `CDB_OB_ARCHIVELOG.CHECKPOINT_SCN` until ≥ `minRestoreSCN`. Default 600s timeout, 5s interval, configurable via `ARCHIVE_CHECKPOINT_TIMEOUT_SECONDS`. Key strengths:
- `COALESCE(MAX(...), 0)` + `^[0-9]+$` regex guard — robust against NULL/empty/non-numeric results
- `${1:?missing ...}` parameter validation
- File-based error aggregation (`archiveCheckpointFailedFile`) — all tenants checked before failing, providing comprehensive diagnostics
- Step reordering: "close archive" moved AFTER checkpoint wait (step 5 → step 7). Correct — must verify archive checkpoint before cleanup. If checkpoint fails, archive stays open for diagnosis. ✅

**4. S3 support + s3_region** — Verified: `SUPPORT_S3=true` in ActionSet gates provider validation at common-scripts.sh:103. `s3_region` appended from `${region:-${DP_STORAGE_REGION:-}}`, required for S3 destinations. `destPrefix` defaults to `s3` (line 189), overridden only by cos/oss host patterns. ✅

**5. Paramsdef null-list fix** — `{{- with get $pd "staticParameters" }}` replacing `{{- if hasKey ... }}` + `{{- get ... }}`. Correct: `hasKey` returns true for null values, causing `range` over nil → empty list header with no items. `with` skips the block entirely when value is nil/empty. Applied consistently across both paramsdef templates. ✅

**6. obtools init container** — Copies `/bin/ob-tools` to shared `kb-tools` emptyDir. Re Blake's observation about observer-container: verified `kb-tools` is mounted by the metrics/obagent container (componentdefinition.yaml:184-185), which is where obtools is consumed. ✅

**7. Restore idempotent guard (ERROR 4179)** — `DBA_OB_TENANTS` check before `ALTER SYSTEM RESTORE`. Handles PRIMARY/STANDBY/RESTORE states with skip + continue; unexpected roles → exit 1. `2>/dev/null` on the query is safe — connection failure → empty result → proceeds to RESTORE attempt → fails naturally. Evidence: 2/13 rounds (15%) reproduced 4179. ✅

### Agrees with Blake's non-blocking observations
- "alert" → "alter" typo is pre-existing (lines 81/109) — trivial follow-up
- `ARCHIVE_CHECKPOINT_TIMEOUT_SECONDS` not exposed in ActionSet — 600s default is reasonable, consider follow-up
- `observer-container` kb-tools mount: resolved — obagent container mounts it

No blocking concerns.

leon-ape:
`addons/oceanbase-ce/dataprotection/backup.sh` uses `[[ $status -ne "SUCCESS" ]]` to check `CDB_OB_BACKUP_SET_FILES.STATUS`. This is an arithmetic comparison; string statuses such as `FAILED` and `SUCCESS` both evaluate to 0, so failed backup sets can skip `failureMessage` and let the backup pass.

weicao:
Addressed @leon-ape review comment in `2cd03750`.

Root cause: `CDB_OB_BACKUP_SET_FILES.STATUS` is a string (`SUCCESS` / `FAILED` / etc.), but `backup.sh` used arithmetic comparison:

```sh
[[ $status -ne "SUCCESS" ]]
```

That can misclassify string statuses and skip `failureMessage` for failed backup sets.

Fix: changed the check to string comparison:

```sh
[[ "${status}" != "SUCCESS" ]]
```

Companion test guard pushed to apecloud/kubeblocks-tests#144 in `969a973` so the static suite now requires this string comparison shape. Local static evidence against the patched addon:

```text
PASS: 81  FAIL: 0  SKIP: 0
```

Scope is intentionally one line in the addon plus one static guard in the test PR; no unrelated cleanup.

weicao:
Follow-up on `2cd03750`: all GitHub checks are now green on the review-fix head.

Current gate after the fix:
- head: `2cd037504af93ee67a70c372ccc54467acb4b3fa`
- checks: shell-check, shellspec-test, generate-files, pr-label-check, and chart checks all SUCCESS
- remaining gate: `REVIEW_REQUIRED` / `BLOCKED`, no merge commit yet

No further code action from my side unless reviewer asks for another change.

leon-ape (approved):

PR #2720, opened 2026-06-01
test(mariadb): lock replication mode contract wording

weicao:
## Summary
- correct MariaDB merged replication topology comments so they match the implemented install-time `mariadb.replication.mode` path
- remove stale wording that described `replicationMode` as a per-Cluster ComponentSpec parameter
- extend the existing replication mode ShellSpec to guard this contract wording from drifting again

## Review basis
- addon API: user-selectable deployment shape must be represented clearly by ClusterDefinition topology; comments must not imply a per-Cluster API that the chart does not implement
- MySQL comparison: MySQL exposes semisync as an explicit topology, while current MariaDB uses one `replication` topology with install/render-time mode wiring

## Validation
- `sh -n addons/mariadb/scripts-ut-spec/replication_merged_replication_mode_env_wire_spec.sh`
- `git diff --check`
- grep-equivalent assertions for the new guard conditions
- `helm dependency build addons/mariadb`
- `helm template test addons/mariadb`

## Boundary
This is a contract/comment/static-test guard only. It does not change rendered runtime resources, does not validate live K8s, and does not prove MariaDB runtime acceptance. Runtime lifecycle fixes remain in the separate MariaDB alpha chain PRs.

### DIFF
--- a/addons/mariadb/Chart.yaml
+++ b/addons/mariadb/Chart.yaml
@@ -2054,11 +2054,9 @@ version: 1.1.1-alpha.110
 #   primary/secondary (plus `standalone` and `galera` unchanged).
 #   `async` and `semisync` are NOT listed: Option B is a breaking
 #   change; users with old Cluster CRs must actively migrate to
-#   `spec.topology=replication` plus a ComponentSpec parameter
-#   `replicationMode: async | semisync`. Auto-rebind by KB on a
-#   topology compDef rename is exploratory (verified in upgrade
-#   gate N=1) and is NOT promised in release notes (Jack design
-#   Gate 6 / Blocker 4).
+#   `spec.topology=replication`. Auto-rebind by KB on a topology
+#   compDef rename is exploratory (verified in upgrade gate N=1) and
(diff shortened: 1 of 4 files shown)


weicao:
CI triage for head `e91e15c97271b96a4e80ce2040456812feabc78a`:

- Green: `shell-check`, `generate-files`, `check-addons-cluster-helm`.
- Red: `shellspec-test`, `check-addons-helm`, dockerhub/aliyun image gates.

`#2720` changes only comments plus static ShellSpec guard wording. The 4 new guard examples are not listed in the ShellSpec failure examples. The ShellSpec red list is existing MariaDB alpha-cycle drift / older spec failures, including pre-existing failures in the same `replication_merged_replication_mode_env_wire_spec.sh` render block.

The chart image gates are also outside this PR: this diff does not change any image field. Treat them as existing image-supply/chart-gate state, not a runtime or image change introduced by this PR.

Validation I ran locally: `sh -n`, `git diff --check`, grep-equivalent guard assertions, `helm dependency build addons/mariadb`, and `helm template test addons/mariadb`.

PR #2729, opened 2026-06-02
fix(mariadb): galera reconfigure auth — set MARIADB_INTERNAL_ROOT_USER=root

weicao:
## Summary

- Set `MARIADB_INTERNAL_ROOT_USER=root` in galera CmpD vars section
- Bumps alpha.116 → alpha.117

## Root cause

The shared reconfigure helper (`_helpers.tpl:288-291`) defaults `MARIADB_INTERNAL_ROOT_USER` to `kb_internal_root`. Standalone/replication/semisync CmpDs provision this account via initdb, but galera CmpD does not. When a galera Reconfiguring OpsRequest fires (e.g. `slow_query_log=ON`), the helper connects as `kb_internal_root@127.0.0.1` — which does not exist in the DB — and fails with `ERROR 1045 (28000): Access denied`.

## Why root is correct for galera

Galera does not apply the alpha.64 SUPER-stripping security hardening. The user-facing `root@%` retains `ALL PRIVILEGES` including `SUPER`, so `root` can execute `SET GLOBAL` without issue.

## Verification

- `helm dependency build` + `helm template` confirms `MARIADB_INTERNAL_ROOT_USER=root` renders in galera CmpD vars
- `git diff --check` clean
- Reconfigure helper resolution: `INTERNAL_ROOT_USER="${MARIADB_INTERNAL_ROOT_USER:-kb_internal_root}"` resolves to `root` with the new env var

## Scope

2 files / +13 / -1. Only galera CmpD affected. No script change, no standalone/replication/semisync change.

### DIFF
--- a/addons/mariadb/Chart.yaml
+++ b/addons/mariadb/Chart.yaml
@@ -1985,7 +1985,17 @@ type: application
 # for any Galera reconfigure touching wsrep_slave_threads, which
 # causes unnecessary Serial pod restarts on a 3-node cluster.
 # Test galera CM3 reconfigures wsrep_slave_threads from 4 to 8.
-version: 1.1.1-alpha.116
+#
+# alpha.117 (Helen 2026-06-02 — galera reconfigure auth fix):
+# set MARIADB_INTERNAL_ROOT_USER=root in cmpd-galera.yaml vars.
+# The shared reconfigure helper defaults to kb_internal_root, but
+# galera CmpD does not provision this account (no initdb block like
+# standalone/replication/semisync). Galera root retains full
+# privileges (no alpha.64 SUPER stripping), so root is the correct
+# reconfigure executor. Without this, any galera Reconfiguring
+# OpsRequest fails with ERROR 1045 (Access denied for
+# kb_internal_root). CmpD spec change requires version bump.
+version: 1.1.1-alpha.117
 
 # alpha.101 v1 (Helen 2026-05-25 — Phase 1 mitigation for orphan
 # event / GTID divergence cascade surfaced after alpha.100): add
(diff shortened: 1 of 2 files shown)


weicao:
LGTM for the PR scope.

What I checked:
- Diff scope is narrow: `Chart.yaml` version/comment bump plus `cmpd-galera.yaml` vars addition only.
- Contract path: the shared reconfigure helper resolves `INTERNAL_ROOT_USER="${MARIADB_INTERNAL_ROOT_USER:-kb_internal_root}"`; Galera now sets `MARIADB_INTERNAL_ROOT_USER=root`, so it no longer falls through to the non-provisioned `kb_internal_root` account.
- Render check: `helm dependency build addons/mariadb` and `helm template test addons/mariadb` pass locally. Rendered `mariadb-galera-1.1.1-alpha.117` contains:
  - `name: MARIADB_INTERNAL_ROOT_USER`
  - `value: root`
- Hygiene: `git diff --check` clean; PR body and commit body clean for generated/AI attribution markers.
- Design-contract review classes 1-8: class 1 is the key fix (removes the bad fallback for Galera); class 2 write-site is explicit in CmpD vars; classes 3-8 do not apply to this two-line env/Chart bump change.

CI note: chart/generate/shell-check jobs are green. ShellSpec is red, but the failed examples are in existing merged-CmpV/PD/schema/mode-wire/switchover/user-convergence/semisync specs; no `cmpd-galera.yaml` failure appears in the failed list. I do not see a PR-scoped blocker there.

Boundary: this is code/render LGTM. The Galera T5 runtime closeout still needs a fresh alpha.117 reconfigure run after merge/deploy.

PR #2733, opened 2026-06-02
fix(valkey): keep primary label on invalid sentinel quorum

weicao:
## Summary
- change Valkey check-role fallback for sentinel-configured quorum-invalid windows
- local `role:master` now exits non-zero without stdout so the controller skips this sample and keeps the previous trusted label
- local `role:slave` still emits `secondary` so demoted pods do not keep a stale primary label
- update shellspec guards/comments for the new fallback contract

## Validation
- `bash -n addons/valkey/scripts/check-role.sh`
- `bash -n addons/valkey/scripts-ut-spec/check_role_spec.sh`
- `git diff --check HEAD~1..HEAD`
- local stub: `role:master` + invalid sentinel quorum -> `rc=1`, stdout empty
- local stub: `role:slave` + invalid sentinel quorum -> `rc=0`, stdout `secondary`

## Boundary
This PR does not provide final release evidence. After merge, kubeblocks-tests T10 needs to be updated to this contract and the Valkey full suite rerun on the new addon head.

### DIFF
--- a/addons/valkey/scripts-ut-spec/check_role_spec.sh
+++ b/addons/valkey/scripts-ut-spec/check_role_spec.sh
@@ -769,12 +769,16 @@ Describe "Valkey Check-Role Bash Script Tests"
     #
     # check-role.sh defence: when Sentinel is configured (any of
     # `insufficient_valid`, `split_view`, etc.) but quorum is invalid,
-    # do NOT emit legacy `primary` from `local INFO=master`. Emit
-    # `secondary` instead, so the primary Service may briefly have no
-    # endpoint while Sentinel converges but no engine-versioned peer
-    # can be cleaned. `no_sentinel_env` / `no_sentinel_configured`
-    # paths keep the original local-INFO mapping because they are the
-    # only authority for standalone topologies.
+    # do NOT emit legacy `primary` from `local INFO=master`. Fail the
+    # current probe sample instead. KB controller skips non-zero
+    # roleProbe events, so the last trusted primary label is preserved
+    # while Sentinel converges, and a new pod with no trusted label is
+    # not promoted from a stale local INFO result. But local
+    # `role:slave` is safe and should emit `secondary` so a demoted pod
+    # does not keep a stale primary label. `no_sentinel_env` /
+    # `no_sentinel_configured` paths keep the original local-INFO
+    # mapping because they are the only authority for standalone
+    # topologies.
     Context "Sentinel configured + quorum invalid + INFO=master"
       check_role_script="../scripts/check-role.sh"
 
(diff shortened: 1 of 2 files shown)


weicao (commented):
## Review: PR #2733 — fix(valkey): roleProbe fail-on-invalid-quorum

**Stats:** 2 files changed, +64 / -38

---

### Summary

Changes roleProbe behavior when Sentinel quorum is invalid. Previously both `role:master` and `role:slave` collapsed to `secondary`. Now they're split:

- `role:master` + no quorum -> **exit 1, no stdout** (controller skips sample, preserves existing label)
- `role:slave` + no quorum -> **emit `secondary`** (safe demotion)
- unknown role -> **exit 1** (fail-safe)

---

### Contract Safety Analysis

| Scenario | Stdout | Exit | Controller action | Safe? |
|---|---|---|---|---|
| master + no quorum | empty | 1 | Skip sample, keep previous label | YES |
| slave + no quorum | `secondary` | 0 | Update label to secondary | YES |
| unknown + no quorum | empty | 1 | Skip sample | YES |

**Critical invariant maintained:** No branch emits `primary` without Sentinel quorum confirmation in a Sentinel-configured topology.

**Branch 1 (master + no quorum -> fail): CORRECT.** A node claiming `role:master` without Sentinel quorum could be a stale master after partition. The old behavior (emit `secondary`) risked false-demotion of a legitimate primary during a quorum gap. The new behavior (skip sample) is more conservative — avoids both false-promotion AND false-demotion.

**Branch 2 (slave + no quorum -> secondary): CORRECT.** The asymmetry is intentional and correct. Trusting demotion is safe: a pod reporting `role:slave` has already accepted `replicaof`. Emitting `secondary` ensures a demoted primary gets its label updated — failing the probe here would leave a stale `primary` label on a replica, which is a split-brain risk.

**Branch 3 (unknown role -> fail): CORRECT.** Empty/garbage/unrecognized role must never be silently accepted.

### Idempotency: PASS

Stateless — reads fresh `INFO replication` and Sentinel state each invocation. Writes no persistent state. Safe to call repeatedly.

### Occam's Razor: PASS

Minimal change: splits one `case` arm into two with different behaviors. No new variables, no structural reorganization. The asymmetric treatment is the simplest correct solution.

### Test Coverage: ADEQUATE

Structural grep-based tests cover all three branches. Acknowledged gap: no end-to-end behavioral tests mocking `valkey-cli` — mitigated by the boundary statement that T10 integration tests will be updated post-merge.

---

### Minor Note

PR title "keep primary label on invalid sentinel quorum" describes the controller-side effect accurately, but the script-side change is "fail the probe when master + no quorum." Acceptable since it describes user-visible outcome.

---

### Verdict: **APPROVE**

All three branches are contract-safe. The critical invariant (never emit `primary` without quorum) is maintained. The asymmetric master/slave treatment is the correct design: trust demotion, distrust promotion without quorum. Minimal, idempotent, well-documented.

weicao (commented):
**Alice LGTM** — focused review on the differentiated fallback:

**check-role.sh** (lines 418-427):
- `role:master` + quorum-invalid → `exit 1` with stderr diagnostic, stdout empty. Controller skips, preserves existing label. Correct: prevents false primary without killing existing primary service.
- `role:slave` + quorum-invalid → `printf %s "secondary"`. Correct: demotion is always safe, avoids keeping stale primary label on a pod that was actually demoted.
- Unknown role → `exit 1` unchanged. Correct.

**Edge case verification**:
- Established primary + sentinel down → roleProbe exit 1 → controller keeps primary label → service endpoint preserved ✓
- Pod just demoted + sentinel briefly invalid → role:slave → reports secondary → label updated to secondary ✓  
- New pod (no prior label) + sentinel invalid → exit 1 → no label set → no false primary ✓
- Sentinel recovers → next probe succeeds → correct label applied ✓

**check_role_spec.sh**: comment block updated, 3 new positive/negative assertions for the split behavior, existing tests adjusted. Count assertions for unknown-role (2 occurrences) still hold.

**Minor note**: no-sentinel path has `"role:slave")  printf` (2 spaces) while quorum-invalid has `"role:slave") printf` (1 space). Both are grepped by separate tests — no functional issue, just a style inconsistency. Not blocking.

PR #2734, opened 2026-06-02
fix(mariadb): galera non-pod-0 waits for Primary before joining

weicao:
## Summary
- Non-pod-0 nodes now wait (up to 120s) for a peer with wsrep_cluster_status=Primary before starting MariaDB in join mode.
- Prevents 1+2 split where pod-1/pod-2 form a dead non-Primary partition before pod-0 bootstraps.
- Bumps chart to 1.1.1-alpha.121.

## Context
Alpha.120 galera full r5 CM4: pod-0 correctly bootstrapped as 1-node Primary, but pod-1/pod-2 had already formed a 2-node non-Primary Galera partition. The two partitions never merge. CM4 focused PASS'd because pod-0 bootstrapped faster (less prior state).

## Test plan
- [ ] Galera full suite: CM4 static reconfigure cluster recovers to wsrep_cluster_size=3
- [ ] Galera T6 Stop/Start: no regression
- [ ] Single pod restart scenario: no unnecessary wait

### DIFF
--- a/addons/mariadb/Chart.yaml
+++ b/addons/mariadb/Chart.yaml
@@ -1986,6 +1986,36 @@ type: application
 # causes unnecessary Serial pod restarts on a 3-node cluster.
 # Test galera CM3 reconfigures wsrep_slave_threads from 4 to 8.
 #
+# alpha.121 (Helen 2026-06-02 — galera non-pod-0 wait for Primary):
+# Fix 1+2 split after parallel restart. Alpha.120 correctly made pod-0
+# bootstrap when peers are non-Primary, but pod-1/pod-2 start MariaDB
+# in join mode BEFORE pod-0 finishes bootstrapping. Pod-1/pod-2 connect
+# to each other via Galera group communication, form a 2-node non-Primary
+# partition (seqno=-1, can't elect primary). Pod-0 then bootstraps as a
+# separate 1-node cluster. Result: pod-0 = 1-node Primary, pod-1/pod-2 =
+# 2-node non-Primary — two separate Galera partitions that never merge.
+#
+# Evidence: alpha.120 galera full r5 CM4 — pod-0 wsrep_cluster_size=1
+# Primary/Synced; pod-1/pod-2 wsrep_cluster_size=2 non-Primary/Initialized.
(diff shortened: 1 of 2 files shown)


weicao (commented):
## Review: PR #2734 — galera non-pod-0 waits for Primary before joining

**Verdict: APPROVE**

### Summary
Single-purpose fix for a real Galera race condition: during full cluster restart with `podManagementPolicy=Parallel`, pod-1/pod-2 could form a dead 2-node non-Primary partition before pod-0 finishes bootstrapping. The fix adds a polling wait (`_wait_for_primary_peer`) so non-pod-0 nodes delay `JOIN` until at least one peer reports `wsrep_cluster_status=Primary`.

### What I checked

1. **Control flow correctness**:
   - `should_bootstrap()` still governs bootstrap vs join — unchanged and correct.
   - `_wait_for_primary_peer` is only called for non-pod-0 in the join branch. Pod-0 never hits this path. ✅
   - Single pod restart: peer already Primary → `_any_peer_alive()` returns 0 on first call → no wait. ✅
   - Pod-0 restart with healthy peers: enters join via `should_bootstrap` returning 1, skips wait (pod_index=0). ✅

2. **Timeout fallback (120s)**:
   - Returns 0 (proceed with join) after timeout. Reasonable — Kubernetes will restart on join failure, and subsequent attempts re-enter the wait loop. Convergence is guaranteed. ✅
   - This is a container entrypoint script, NOT a kbagent lifecycle action, so the 60s kbagent timeout cap does not apply. ✅

3. **Occam's razor**: Minimal change — one new function (polling loop over existing `_any_peer_alive`), 4 lines in `main()`. No spec changes. No new dependencies. ✅

4. **Code-description consistency**: PR description matches diff exactly. ✅

5. **Single purpose**: Yes — one race condition fix + version bump for traceability. ✅

### Minor observation (non-blocking)

Each `_any_peer_alive()` call logs per-peer status lines. Over 40 iterations (120s / 3s), this could produce ~80–120 lines of log noise during a full cluster restart wait. Consider suppressing per-peer status inside the polling loop (e.g., only log on final success/timeout) in a future cleanup pass. Not worth changing now.

PR #2735, opened 2026-06-02
fix(mariadb): galera self-healing watcher kills stuck non-Primary nodes

weicao:
## Summary
- Background watcher now detects `wsrep_cluster_status=non-Primary` persisting for 90s and kills mariadbd to force container restart
- Fixes TOCTOU race where pod-1/2 lose pod-0 mid-SST during second restart wave, forming a dead non-Primary partition that Kubernetes never restarts (MariaDB alive but irrecoverable)
- Counter resets on Synced (state=4) or any non-non-Primary cluster status; 90s threshold avoids false kills during normal SST/IST

## Context
Alpha.121 added `_wait_for_primary_peer` so non-pod-0 waits for pod-0 to bootstrap. But InstanceSet controller can issue a second restart wave: pod-0 restarts again after pod-1/2 started joining. Pod-1/2 fall to non-Primary and stay stuck. This self-healing closes the gap.

## Test plan
- [ ] CM4 static reconfigure (back_log=200): Stop/Start → verify all 3 nodes reach wsrep_cluster_size=3, wsrep_cluster_status=Primary
- [ ] Full galera suite pass (T1-T6, CM1-CM5)

### DIFF
--- a/addons/mariadb/Chart.yaml
+++ b/addons/mariadb/Chart.yaml
@@ -1986,6 +1986,30 @@ type: application
 # causes unnecessary Serial pod restarts on a 3-node cluster.
 # Test galera CM3 reconfigures wsrep_slave_threads from 4 to 8.
 #
+# alpha.122 (Helen 2026-06-02 — galera self-healing non-Primary watcher):
+# Fix TOCTOU race in parallel restart. Alpha.121 added _wait_for_primary_peer
+# so pod-1/pod-2 wait for pod-0 to bootstrap. But the InstanceSet controller
+# can issue a second restart wave: pod-0 restarts AGAIN after pod-1/pod-2
+# already saw it as Primary and started joining. Pod-1/pod-2 lose pod-0
+# mid-SST, fall to non-Primary, and form a dead 2-node partition. New pod-0
+# bootstraps as a 1-node singleton. The 1+2 split persists because the stuck
+# nodes' MariaDB is alive (not crashing), so Kubernetes never restarts them.
+#
+# Evidence: alpha.121 galera full r1 CM4 — pod-1/pod-2 wait log shows they
+# DID find pod-0 Primary, started join, then pod-0 restarted again.
+# Pod-1/pod-2 stuck at wsrep_cluster_status=non-Primary indefinitely.
+#
+# Fix: add self-healing to the background watcher. The watcher now also
+# queries wsrep_cluster_status each cycle. If the node stays in non-Primary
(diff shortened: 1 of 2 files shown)


weicao (commented):
## Review: PR #2735 — galera self-healing watcher kills stuck non-Primary nodes

**Verdict: APPROVE**

### Summary
Adds self-healing to the existing background watcher: if `wsrep_cluster_status=non-Primary` persists for 90s (30 cycles × 3s), the watcher kills mariadbd to force a container restart. This closes the TOCTOU gap left by alpha.121: InstanceSet controller can restart pod-0 again after pod-1/2 started joining, leaving them stuck in a dead non-Primary partition that Kubernetes won't restart (MariaDB is alive, just irrecoverable).

### What I checked

1. **Signal correctness — `non-Primary` is the right trigger**:
   - During normal SST/IST, the joining node is part of the primary component and reports `wsrep_cluster_status=Primary` even while `wsrep_local_state` is 1 (Joining). The `non-Primary` status specifically indicates a dead partition that can't form a primary component. ✅
   - 90s threshold is conservative enough to avoid false kills during legitimate state transitions. ✅

2. **Counter logic**:
   - Increment: only when socket exists AND `CLUSTER_STATUS=non-Primary` AND `STATE≠4`. ✅
   - Reset: when `STATE=4` (Synced), OR socket unavailable, OR `CLUSTER_STATUS≠non-Primary`. ✅
   - This correctly scopes the kill to confirmed, persistent dead-partition states only. ✅

3. **Kill mechanism**:
   - SIGTERM → 5s grace → SIGKILL is standard graceful shutdown. ✅
   - `pkill mariadbd` targets the correct process (exec'd PID 1 in the container). ✅
   - Container restart triggers galera-start.sh re-evaluation → `_wait_for_primary_peer` (alpha.121) → joins now-stable pod-0. ✅

4. **Convergence / no infinite loop**:
   - Each kill+restart cycle gives pod-0 a ~120s window to stabilize before pod-1/2 join again. If pod-0 is stable, the join succeeds. If instability persists, CrashLoopBackOff correctly escalates. ✅

5. **Occam's razor**: Minimal addition — counter + threshold + kill in existing watcher loop. No new processes, no new dependencies, no spec changes. ✅

6. **Single purpose**: Yes — one self-healing mechanism for one specific TOCTOU gap. ✅

7. **Code-description consistency**: PR description matches the diff. ✅

### No issues found. Clean approve.

PR #2740, opened 2026-06-02
fix(mariadb): reduce galera polling log noise during wait-for-Primary

weicao:
## Summary
- Add quiet parameter to `_any_peer_alive` to suppress non-Primary peer status lines during polling loop
- Success lines ("Peer X is alive with Primary") always log regardless of quiet mode
- First call in `_wait_for_primary_peer` and all `should_bootstrap` calls remain verbose

## Context
Addresses Jade's non-blocking review feedback on PR #2734: during a full cluster restart, the 120s polling loop calls `_any_peer_alive` up to 40 times, each logging per-peer status. This produces ~80-120 lines of repetitive log noise.

## Test plan
- [ ] Galera full suite still passes (no behavioral change, log-only)
- [ ] During wait-for-Primary, only success/timeout lines appear (no per-peer non-Primary lines)

### DIFF
--- a/addons/mariadb/Chart.yaml
+++ b/addons/mariadb/Chart.yaml
@@ -1986,6 +1986,14 @@ type: application
 # causes unnecessary Serial pod restarts on a 3-node cluster.
 # Test galera CM3 reconfigures wsrep_slave_threads from 4 to 8.
 #
+# alpha.123 (Helen 2026-06-02 — galera reduce polling log noise):
+# Address Jade's non-blocking review feedback on PR #2734: _any_peer_alive
+# logs per-peer status on every call, producing ~80-120 lines during the
+# _wait_for_primary_peer polling loop (40 iterations × 2 peers). Add a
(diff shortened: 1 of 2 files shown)


weicao (commented):
## Review: kubeblocks-addons #2740 — MariaDB galera log noise reduction

Clean, minimal change. Addresses the non-blocking feedback I raised on PR #2734 about repetitive polling log noise in `_wait_for_primary_peer`.

### Analysis

- `_any_peer_alive()` gains a `local quiet="${1:-}"` parameter. When `quiet` is non-empty, the per-peer "not Primary" status lines are suppressed via `[ -z "${quiet}" ] && echo ...`.
- `_wait_for_primary_peer` passes `quiet` to the inner function, silencing the repetitive non-Primary lines during the polling loop. The important "cluster has Primary" success message still prints.
- The `quiet` parameter pattern is idiomatic for shell scripts — clean and zero-overhead.
- Version bump alpha.122 → alpha.123 is correct.

No concerns. Single-purpose PR, code matches description.

PR #2741, opened 2026-06-02
fix(redis): dynamic reconfigure fails for hyphenated parameters

weicao:
## Summary

- Fix dynamic reconfigure (CONFIG SET) failing for Redis parameters with hyphens in their names (e.g. `maxmemory-policy`, `repl-backlog-size`, `slowlog-max-len`)
- Root cause was a three-layer bug stack:
  1. `/bin/sh` strips env vars with invalid POSIX names (hyphens) from its environment, making `env`/`printenv` unable to see them
  2. `set -eu` + pipe `while` loop (`tr | while`) runs in subshell where `set -e` silently exits on first `redis-cli` stderr warning
  3. Redis Cluster scripts ConfigMap was missing `reload-parameter.sh`, breaking reconfigure on cluster topology
- Fix reads from `/proc/self/environ` (preserves original `execve` environment), uses file-redirect loop to avoid subshell, and includes `reload-parameter.sh` in redis-cluster scripts

## Changes

- `_helpers.tpl` `redis.config.reconfigureAction`: Replace `env | cut` + pipe while with `/proc/self/environ` + file-redirect while
- `_helpers.tpl` `redis-cluster.extend.scripts`: Add `reload-parameter.sh` to redis-cluster scripts ConfigMap

## Test plan

- [x] Standalone: 4-param reconfigure (maxmemory-policy, hz, slowlog-max-len, repl-backlog-size) — CONFIG GET all updated, no pod restart
- [x] Replication: Same 4 params — both data pods updated
- [x] Twemproxy: Same 4 params — both backing Redis data pods updated
- [x] Redis Cluster: Same 4 params — all 6 pods (3 shards x 2 replicas) updated
- [ ] Full parameters test suite regression

### DIFF
--- a/addons/redis/templates/_helpers.tpl
+++ b/addons/redis/templates/_helpers.tpl
@@ -200,14 +200,16 @@ reconfigure:
       - /bin/sh
       - -c
       - |
+        rc=0
         tr '\0' '\n' < /proc/self/environ > /tmp/_reconf_env.txt
         while IFS= read -r line; do
           key="${line%%=*}"
           printf '%s\n' "$key" | grep -qE '^[a-z0-9_.-][a-z0-9_.-]*$' || continue
           value="${line#*=}"
-          /scripts/reload-parameter.sh "$key" "$value"
+          /scripts/reload-parameter.sh "$key" "$value" || rc=$?
         done < /tmp/_reconf_env.txt
         rm -f /tmp/_reconf_env.txt
+        exit "$rc"
 {{- end -}}
 
 {{- define "apeDts.reshard.image" -}}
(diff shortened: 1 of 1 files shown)


wangyelei:
https://github.com/apecloud/kubeblocks/pull/10182/changes look at this pr. only can reconfigure parameters by reconfigure operation.

addons/redis/templates/_helpers.tpl
```diff
@@ -196,12 +200,16 @@ reconfigure:
       - /bin/sh
       - -c
       - |
-        set -eu
-
-        env | cut -d= -f1 | grep -E '^[a-z0-9_.-][a-z0-9_.-]*$' | sort -u | while IFS= read -r param; do
-          [ -n "${param}" ] || continue
-          /scripts/reload-parameter.sh "${param}" "$(printenv "${param}")"
```
wangyelei, line 203: only replace param `_` to `-`

wangyelei: replace `^[a-z0-9_.-][a-z0-9_.-]*$` to `^[a-zA-Z0-9_.-]+$`

wangyelei (approved):

PR #2742, opened 2026-06-02
chore(mongodb): avoid rendering null volumes in CmpD templates

weicao:
## Summary
- When `logCollector.enabled=false` (default), the `volumes:` key in MongoDB CmpD templates was rendered with no items, producing a null value
- KB 1.2 strict CRD validation rejects null arrays for `spec.runtime.volumes`
- Fix: move the `volumes:` key inside the `{{- if .Values.logCollector.enabled }}` block so it is only emitted when logCollector is enabled

## Affected templates
- `cmpd.yaml` (mongodb)
- `cmpd-config-server.yaml` (mongo-config-server)
- `cmpd-mongodb-shard.yaml` (mongo-shard)
- `cmpd-mongos.yaml` is not affected (has unconditional volume entries)

## Test plan
- [x] `helm lint addons/mongodb` passes
- [x] `helm template` renders without `volumes: null`
- [x] Verified on KB 1.2.0-alpha.1: CmpDs install and reach Available status
- [x] MongoDB 8.0.17 cluster creates and reaches Running with 1P/2S

### DIFF
--- a/addons/mongodb/templates/cmpd-config-server.yaml
+++ b/addons/mongodb/templates/cmpd-config-server.yaml
@@ -242,13 +242,13 @@ spec:
         env:
           - name: MONGODB_URI
             value: mongodb://127.0.0.1:$(KB_SERVICE_PORT)
+    {{- if .Values.logCollector.enabled }}
     volumes:
-      {{- if .Values.logCollector.enabled }}
       - name: log-data
         hostPath:
           path: /var/log/kubeblocks
           type: DirectoryOrCreate
-      {{- end }}
+    {{- end }}
   services:
     - name: default
       spec:
--- a/addons/mongodb/templates/cmpd-mongodb-shard.yaml
+++ b/addons/mongodb/templates/cmpd-mongodb-shard.yaml
@@ -289,13 +289,13 @@ spec:
         env:
           - name: MONGODB_URI
             value: mongodb://127.0.0.1:$(KB_SERVICE_PORT)
+    {{- if .Values.logCollector.enabled }}
     volumes:
-      {{- if .Values.logCollector.enabled }}
       - name: log-data
         hostPath:
           path: /var/log/kubeblocks
           type: DirectoryOrCreate
-      {{- end }}
+    {{- end }}
   vars:
     - name: CLUSTER_NAME
       valueFrom:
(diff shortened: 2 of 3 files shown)


weicao (commented):
Cross-review from Amber — **MongoDB fix APPROVE, but Redis changes need split.**

**MongoDB `volumes: null` fix (3 files) — APPROVE:**
- `cmpd.yaml`, `cmpd-config-server.yaml`, `cmpd-mongodb-shard.yaml`: moved `{{- if .Values.logCollector.enabled }}` to wrap both `volumes:` key and content, so `logCollector.enabled=false` produces no `volumes:` key at all (instead of `volumes:` with null value)
- Fix is correct, minimal, consistent across all 3 CmpD templates
- KB 1.2 strict CRD validation will no longer reject the rendered manifest

**Redis `_helpers.tpl` changes — SCOPE ISSUE:**
This PR title is `fix(mongodb): avoid rendering null volumes in CmpD templates` but includes Redis changes:
1. Added `reload-parameter.sh` to scripts configmap
2. Refactored reconfigure action: replaced `env | cut | grep` pipeline with `/proc/self/environ` parsing + per-key `rc` capture

These are unrelated to the MongoDB volumes fix. Per team convention (one PR one purpose), these should be a separate PR. Jade will likely flag this too.

**Recommendation:** split Redis changes into a separate PR. MongoDB fix can merge independently.

addons/clickhouse/README.md
```diff
@@ -830,6 +830,30 @@ spec:
 
 ```
 
+```yaml
```
leon-ape: This MongoDB PR adds ClickHouse README content, so the diff is changing another addon.

weicao:
Cleaned up the PR branch after Leon11's review: force-pushed from d33c3033 back to 40a64718 so the unrelated addons/clickhouse/README.md generated diff is removed.\n\nCurrent PR diff is only the intended MongoDB CmpD template change:\n- addons/mongodb/templates/cmpd.yaml\n- addons/mongodb/templates/cmpd-config-server.yaml\n- addons/mongodb/templates/cmpd-mongodb-shard.yaml\n\nNo runtime behavior changed beyond the existing MongoDB volumes-null fix.

leon-ape (approved):

PR #2744, opened 2026-06-03
fix(valkey): use bash arrays for CLI commands to prevent password word-splitting

weicao:
## Summary
- Converts all 13 Valkey addon scripts from string-based CLI command construction to bash array-based construction to prevent word-splitting when passwords contain spaces or special characters
- Updates 3 ShellSpec test files with helper wrappers to accommodate the array-based API change
- Changes shebangs from `#!/bin/sh` to `#!/bin/bash` for scripts that now use bash arrays

## Context
Addresses Issue 1 (password quoting / word-splitting) from Blake review COMMENT on PR #2592.

## Files changed (16)
**Scripts (13):** check-role.sh, switchover.sh, sync-acl.sh, valkey-self-heal.sh, valkey-register-to-sentinel.sh, valkey-member-leave.sh, valkey-sentinel-start.sh, valkey-start.sh, post-provision.sh, pre-stop.sh, reload-parameter.sh, account-provision.sh, valkey-sentinel-account-provision.sh

**Tests (3):** check_role_spec.sh, valkey_switchover_spec.sh, valkey_member_leave_spec.sh

## Test plan
- [x] bash -n passes on all 13 modified scripts
- [x] grep confirms zero residual unquoted password patterns
- [x] ShellSpec tests updated with array-to-string helper wrappers
- [ ] ShellSpec unit tests pass
- [ ] Full suite regression (covered by r34 forensic on parent branch; this PR is array-only refactor with no logic change)

### DIFF
--- a/addons/valkey/scripts-ut-spec/check_role_spec.sh
+++ b/addons/valkey/scripts-ut-spec/check_role_spec.sh
@@ -929,7 +929,7 @@ Describe "Valkey Check-Role Bash Script Tests"
         # variable rather than always hard-coding `-a "${SENTINEL_PASSWORD}"`.
         # That way missing password leaves auth args empty and the cli
         # call NOAUTH-fails, routing to `insufficient_valid`.
-        When call grep -F '${sentinel_auth_args} ${sentinel_tls_args} sentinel masters' "${check_role_script}"
+        When call grep -F '"${sentinel_auth_args[@]}" ${sentinel_tls_args} sentinel masters' "${check_role_script}"
         The status should be success
         The stdout should include 'sentinel_auth_args'
       End
--- a/addons/valkey/scripts-ut-spec/check_role_spec.sh
+++ b/addons/valkey/scripts-ut-spec/check_role_spec.sh
@@ -28,6 +28,11 @@ Describe "Valkey Check-Role Bash Script Tests"
   AfterAll "cleanup"
 
   Describe "build_cli_cmd()"
+    _cli_cmd_as_string() {
+      build_cli_cmd
+      printf '%s\n' "${cli_cmd[*]}"
+    }
+
     Context "without password or TLS"
       setup() {
         unset VALKEY_DEFAULT_PASSWORD
@@ -36,7 +41,7 @@ Describe "Valkey Check-Role Bash Script Tests"
       Before "setup"
 
       It "builds a basic valkey-cli command"
-        When call build_cli_cmd
+        When call _cli_cmd_as_string
         The status should be success
         The stdout should include "valkey-cli --no-auth-warning"
         The stdout should include "-h 127.0.0.1"
@@ -57,7 +62,7 @@ Describe "Valkey Check-Role Bash Script Tests"
       After "teardown"
 
       It "includes -a flag"
-        When call build_cli_cmd
+        When call _cli_cmd_as_string
         The status should be success
         The stdout should include "-a secret"
       End
@@ -77,7 +82,7 @@ Describe "Valkey Check-Role Bash Script Tests"
       After "teardown"
 
       It "uses the custom port"
-        When call build_cli_cmd
+        When call _cli_cmd_as_string
         The status should be success
         The stdout should include "-p 6380"
       End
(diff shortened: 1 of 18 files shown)


weicao:
COMMENT verdict from Bob review: **BLOCKER on current head `77548c5a`**.

The array conversion direction is right, but the caller/test contract migration is incomplete and ShellSpec is currently red.

Evidence:
- GitHub `shellspec-test` failed: 851 examples, 11 failures.
- I also checked the changed scripts locally: `bash -n` for the 13 modified production scripts passed, and `git diff --check origin/feat/valkey-addon...HEAD` passed. So this is not syntax whitespace; it is contract drift.

Concrete blockers:

1. `reload_parameter_spec.sh` still invokes the now-bash script through `sh`.
   - Lines 47/53/59/65/72 use `When run sh ../scripts/reload-parameter.sh ...`.
   - CI stderr is `../scripts/reload-parameter.sh: 24: Syntax error: "(" unexpected` because the script now uses bash arrays.
   - Fix the spec to run via bash or execute the script by path/shebang, and keep a production-path check that the lifecycle command still executes the shebang rather than `sh script`.

2. `valkey_member_leave_spec.sh` still uses the old "builder echoes command string" contract in the role detection cases.
   - Lines 150/161 do `_data_cli=$(build_data_cli ...)`, but `build_data_cli` now sets `_data_cli_cmd` and emits no stdout.
   - Lines 151/162 then run an empty command prefix, so `leaving_role` is empty; CI fails both slave/master assertions.
   - Fix these to `build_data_cli ...; leaving_role=$("${_data_cli_cmd[@]}" INFO replication ...)`.

3. `valkey_self_heal_spec.sh` was not migrated with the production signature change.
   - Lines 44/65 still assert stdout from `cascade_build_local_cli_cmd`, but the function now sets `_cli_cmd` and emits nothing.
   - Lines 374/383/392 still pass a command string into `dual_master_confirm_demote`; production now reads `_dm_cli_cmd[@]` and takes only the expected master host.
   - Update the specs to assert the arrays and to seed `_dm_cli_cmd=(...)` before calling `dual_master_confirm_demote`.

4. `check_role_spec.sh` has a stale static assertion.
   - Line 932 greps for literal `${sentinel_auth_args} ${sentinel_tls_args} sentinel masters`.
   - The production line now correctly uses `"${sentinel_auth_args[@]}" ${sentinel_tls_args} sentinel masters`, so the old grep fails. Update the assertion to the array form.

XP design-contract review notes:
- Class 3/4 blocker: the command-builder contract changed from "stdout string" to "global array side effect". Several callers/specs still use the old continuous/sentinel-string contract.
- Classes 1/2/5/6/7/8: I did not see a new production blocker in the password-array paths; auth args themselves are quoted as array elements.
- Residual non-blocking scope note: TLS args are still intentionally split from `VALKEY_CLI_TLS_ARGS` in several places. That is pre-existing and not the password word-splitting issue, but the PR body should avoid implying every CLI argument source is now fully array-safe.

Please fix the contract drift and get ShellSpec green before merge. r34 forensic does not validate this PR because it is pinned to #2743 head, not this head.

Note: GitHub does not allow this shared account to mark Request changes on its own PR, so this is a COMMENT verdict rather than a formal review state.

weicao:
COMMENT verdict from Bob re-review on head `ad8f9e8e8`: **LGTM**.

Rechecked the four blockers from my previous review:
1. `reload_parameter_spec.sh` now runs `reload-parameter.sh` through bash instead of `sh`.
2. `valkey_member_leave_spec.sh` role detection now invokes `"${_data_cli_cmd[@]}"` after `build_data_cli`.
3. `valkey_self_heal_spec.sh` now asserts the array-builder contract via helper output and seeds `_dm_cli_cmd=(...)` before `dual_master_confirm_demote`.
4. `check_role_spec.sh` now matches the array-form sentinel auth args.

Validation:
- `bash -n` passed for the 13 modified production scripts.
- `git diff --check origin/feat/valkey-addon...HEAD` passed.
- PR body and commits are clean for AI attribution after the amend.
- GitHub checks are all green, including `shellspec-test`; merge state is `CLEAN`.

I also walked the XP design-contract checklist again. The previous class 3/4 contract drift is closed. I do not see a remaining production blocker in the password-array paths. Residual TLS arg splitting is pre-existing and outside this password word-splitting fix scope.

Note: GitHub still cannot record my formal approval on this shared-account-owned PR, so this is a COMMENT verdict, not a formal APPROVE state.

PR #2771, opened 2026-06-04
chore(valkey): add weicao as code owner

weicao:
Add maintainers field with weicao to addons/valkey/Chart.yaml. When update-codeowners.sh runs, this will map addons/valkey/ to @weicao in .github/CODEOWNERS. Per weston directive.


weicao:
Closing in favor of #2772 which has clean attribution. My commit had a Co-Authored-By trailer that shouldn't be in public commits.

PR #2791, opened 2026-06-08
feat(postgresql): add standby DR promote operation

weicao:
## Summary
- add pg-promote-standby OpsDefinition for PostgreSQL standby-cluster DR cutover
- implement Patroni standby_cluster promotion by requiring force=true, removing standby_cluster from dynamic config, and waiting for read-write status
- add shellspec coverage for force/standby guards and non-standby-leader no-op

## Validation
- bash -n addons/postgresql/scripts/promote_standby.sh
- shellspec --helperdir shellspec --shell /opt/homebrew/bin/bash addons/postgresql/scripts-ut-spec/promote_standby_spec.sh
- helm dependency build addons/postgresql
- helm template pg addons/postgresql >/tmp/pg-dr-cutover-rendered.yaml
- helm lint addons/postgresql
- git diff --check

### DIFF
--- a/addons/postgresql/scripts/promote_standby.sh
+++ b/addons/postgresql/scripts/promote_standby.sh
@@ -31,12 +31,12 @@ warn_if_source_reachable() {
   fi
 
   local host="${endpoint%:*}" port="${endpoint##*:}"
-  if [ -z "$host" ] || [ -z "$port" ] || [ "$host" = "$port" ]; then
+  if [ -z "$host" ] || [ -z "$port" ] || [ "$host" = "$port" ] || ! [[ "$port" =~ ^[0-9]+$ ]]; then
     echo "WARNING: invalid sourceEndpoint '$endpoint'; expected host:port. Skipping source reachability probe."
     return 0
   fi
 
-  if timeout 3 bash -c "cat < /dev/null > /dev/tcp/${host}/${port}" 2>/dev/null; then
+  if timeout 3 bash -c 'cat < /dev/null > "/dev/tcp/$1/$2"' _ "$host" "$port" 2>/dev/null; then
     echo "WARNING: sourceEndpoint ${endpoint} is still TCP-reachable while promoting standby. Ensure the source is fenced/stopped to avoid split-brain."
   else
     echo "sourceEndpoint ${endpoint} is not TCP-reachable in best-effort probe."
--- a/addons/postgresql/scripts/promote_standby.sh
+++ b/addons/postgresql/scripts/promote_standby.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+# Promote a Patroni standby_cluster by removing the dynamic standby_cluster
+# configuration. Patroni documents this as the supported manual promotion path
+# for a remote standby cluster; the caller must first fence/stop the source.
+set -euo pipefail
+
+patroni_url() {
+  printf 'http://%s:8008' "${CURRENT_POD_IP:-localhost}"
+}
+
+is_true() {
+  case "$(printf '%s' "${1:-}" | tr '[:upper:]' '[:lower:]')" in
+    true|yes|1) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
+require_force() {
+  if ! is_true "${force:-}"; then
+    echo "force=true is required; promote only after the source cluster is fenced/stopped."
+    return 1
+  fi
+}
+
+require_standby_mode() {
+  if ! printf '%s' "${PG_MODE:-}" | tr '[:upper:]' '[:lower:]' | grep -q 'standby'; then
+    echo "PG_MODE=${PG_MODE:-<unset>} is not standby; refusing DR standby promotion."
+    return 1
+  fi
+}
+
+is_standby_leader() {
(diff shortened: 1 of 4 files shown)


addons/postgresql/scripts/promote_standby.sh
```diff
@@ -0,0 +1,104 @@
...
+    return 0
+  fi
+
+  local host="${endpoint%:*}" port="${endpoint##*:}"
+  if [ -z "$host" ] || [ -z "$port" ] || [ "$host" = "$port" ]; then
+    echo "WARNING: invalid sourceEndpoint '$endpoint'; expected host:port. Skipping source reachability probe."
+    return 0
+  fi
+
+  if timeout 3 bash -c "cat < /dev/null > /dev/tcp/${host}/${port}" 2>/dev/null; then
```
leon-ape: sourceEndpoint is interpolated into a bash -c string before parsing, so a crafted endpoint can execute additional shell commands in the action pod.

weicao: Addressed in commit `3eafa3bc`.

The TCP probe no longer interpolates `sourceEndpoint` into the `bash -c` command string. It now parses `host`/`port`, rejects non-numeric ports, and passes host/port as positional parameters:

```bash
timeout 3 bash -c 'cat < /dev/null > "/dev/tcp/$1/$2"' _ "$host" "$port"
```

I also added a ShellSpec regression case with a crafted `sourceEndpoint` containing shell metacharacters and a `touch` payload; the test asserts the payload file is not created.

Local validation: `bash -n` and `git diff --check` passed. Local ShellSpec could not run in this PR worktree because the repo `shellspec/` helper load path is absent there; GitHub `shellspec-test` is running on the pushed commit.

addons/postgresql/templates/pcr.yaml
```diff
@@ -0,0 +1,29 @@
+{{- range .Values.versions }}
+---
+apiVersion: parameters.kubeblocks.io/v1alpha1
+kind: ParamConfigRenderer
```
leon-ape, line 4: ParamConfigRenderer is deprecated in the target KB API, but this PR adds new PCR resources.

addons/postgresql/templates/promote_standby.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: operations.kubeblocks.io/v1alpha1
+kind: OpsDefinition
```
leon-ape, line 2: This PR adds a new OpsDefinition for the DR promote operation, which is not the recommended API path for new addon operations.

addons/postgresql/templates/pcr.yaml
weicao: Ack. I agree this should not add new ParamConfigRenderer resources on the current target API path. I have moved the PR back to draft and will either remove this resource from the PR or rework it through the non-deprecated config binding path before asking for review again.

addons/postgresql/templates/promote_standby.yaml
weicao: Ack. I agree this PR should not proceed as a new OpsDefinition-based operation if that is not the recommended new addon operation API path. I have moved the PR back to draft. Next step is to redesign the DR promote entry on the recommended operation path, or split this PR down to only pieces that are valid under the current API contract.

weicao:
Status refresh after Leon/Jade reviewer follow-up:

- This PR should remain draft. Leon's two API-path comments are real blockers, not just review nits:
  - adding new `ParamConfigRenderer` resources on the target API path is not acceptable;
  - adding the DR promote flow as a new `OpsDefinition` is not the recommended API path for new addon operations.
- The earlier shell-injection review comment is addressed in head `3eafa3bc6608ce7113a48ddc065bc515444d8cfd`: `sourceEndpoint` is no longer interpolated into the `bash -c` string, and a ShellSpec regression checks a crafted endpoint payload does not execute.
- The branch is also behind `main` by 90 commits, so it needs a rebase after redesign.

Owner: Violet. Next: redesign/split this PR before requesting review again. The valid path is either to remove PCR/OpsDefinition pieces from this PR or rework the operation through the current recommended addon operation/config APIs, then rebase and rerun shellspec/render/runtime DR validation. No ready-for-review ETA until that redesign lands.

weicao:
Closing this draft instead of keeping it in the active review queue.

Reason: the current PR mixes the useful shell-injection hardening with the DR promote operation shape that still has real API-path blockers from review: the PCR addition and new OpsDefinition path need redesign/split before this can be reviewable. Rebasing the current branch would not make it ready; it would only refresh a blocked design.

What remains valid from this branch: head 3eafa3bc6608ce7113a48ddc065bc515444d8cfd fixed the sourceEndpoint shell-interpolation issue and added a ShellSpec regression for a crafted endpoint payload. That can be salvaged into a smaller follow-up PR if we still need that hardening independently.

Next: reopen as separate scoped work only after redesign:
1. a small script/test hardening PR if the injection regression is still wanted;
2. a separate DR promote design PR only after choosing the current recommended addon operation/config API path and adding runtime validation.

No ready-for-review ETA for this closed draft.

PR #2796, opened 2026-06-09
chore(valkey): param ranges, io-threads, backup/restore hardening, dead code cleanup

weicao:
## Summary

Comprehensive Valkey addon cleanup and hardening:

- practical parameter range fixes and `io-threads` defaulting
- backup / restore script hardening
- dead-code cleanup in addon scripts and templates
- sentinel startup / registration timeout and deadline guards
- reconfigure action rewrite to read mounted ConfigMap content directly, apply dynamic parameters via `CONFIG SET`, and fail closed on stale projection or failed runtime readback
- focused fixes for the ConfigMap projection / marker / retry edge cases found during T13 and VScale validation
- ShellSpec coverage for reload, restore, postProvision budget, startup, and related script contracts

## Validation

- [x] GitHub CI green: shellcheck, ShellSpec, helm/chart checks, label check
- [x] PR #2796 exact candidate validated:
  - addon: `2ccf8a8476cd1962f1424b37a1dde694100a4b73`
  - tests: `ef77e573e58842da918e5958135be6481a982dd8`
  - chart: `0.1.1`
  - Valkey: `9.0.4`
- [x] 10 clean full-suite rounds reached release-ready gate:
  - counted rounds: `r1/r2/r3/r5/r6/r7/r8/r9/r10/r11`
  - invalidated round: `r4` due to environment/VPN interruption
  - coverage: `17 suites x 10 rounds = 170 suite-runs`
  - result: `170/170` suite-runs passed, `runner_rc=0` for counted rounds, cleanup `NotFound`
- [x] Final round r11:
  - finished: `2026-06-17T11:28:58+0800`
  - result: `17/17 PASS`
  - `runner_rc=0`
  - all cleanup `NotFound`

## Review Focus

- `addons/valkey/scripts/reload-config.sh`: projection freshness, marker handling, retry-safe failures, runtime readback
- Valkey lifecycle scripts: sentinel registration/startup timeout behavior
- `addons/valkey/templates/paramsdef.yaml`: practical parameter bounds
- ShellSpec coverage under `addons/valkey/scripts-ut-spec/`


weicao:
Focused review note: not ready to merge yet.

1. Parameter ranges are not yet a practical-range closeout. Several high-risk numeric params now only have lower bounds, for example maxmemory >=0, proto-max-bulk-len >=1, auto-aof-rewrite-percentage >=0. That blocks negative values but still allows obviously dangerous huge values. Please add a small parameter range matrix with evidence/source for each practical lower/upper bound, plus negative schema tests. Params without enough evidence should not be claimed as practical-range closed in this PR.

2. This combined PR should stay draft until the combined head has review + CI + focused runtime validation. The current GitHub red check appears to be DockerHub login timeout, not code, but it is still a red gate; and the old 6/0 belongs to the pre-combine candidate, not this exact head.

Checked OK locally: restore.sh dirty-data fail-closed shape, restore contract behavior tests, dead-code removal references, bash -n, helm lint/template, ShellSpec 224/0.

addons/valkey/scripts/valkey-register-to-sentinel.sh
```diff
@@ -124,25 +124,25 @@ register_to_one_sentinel() {
   master_addr=$("${cli[@]}" SENTINEL get-master-addr-by-name "${master_name}" 2>/dev/null || true)
   if is_empty "${master_addr}" || [ "${master_addr}" = "(nil)" ]; then
     echo "Sentinel not yet monitoring '${master_name}' — issuing SENTINEL monitor..."
-    call_func_with_retry 3 5 execute_sentinel_cmd "${sentinel_fqdn}" \
+    call_func_with_retry 3 2 execute_sentinel_cmd "${sentinel_fqdn}" \
       SENTINEL monitor "${master_name}" "${primary_host}" "${primary_port}" 2 || return 1
```
leon-ape: This still registers the monitor with fixed quorum 2. In a 5-Sentinel cluster, only 2 S_DOWN votes are required instead of strict majority 3, weakening Sentinel failover safety.

weicao: 已修复。cherry-pick 了 PR #2801 的改动（commit `969182e`），quorum 现在从 `SENTINEL_POD_FQDN_LIST` 动态计算 `(n/2)+1`：

```bash
sentinel_monitor_quorum=$(( sentinel_count / 2 + 1 ))
```

- 3 Sentinel → quorum=2（和之前一样）
- 5 Sentinel → quorum=3（strict majority，修复了你指出的安全问题）

新 head: `77b6650e`，shellspec 34/0 全绿。

addons/valkey/scripts/reload-config.sh
```diff
@@ -0,0 +1,252 @@
+#!/bin/sh
+set -eu
+
+CONFIG_FILE="${CONFIG_FILE:-/etc/conf/valkey.conf}"
+DATA_LINK="${DATA_LINK:-/etc/conf/..data}"
+RELOAD_PARAM_SCRIPT="${RELOAD_PARAM_SCRIPT:-/scripts/reload-parameter.sh}"
+RELOAD_VERIFY_CMD="${RELOAD_VERIFY_CMD:-}"
+MAX_WAIT="${MAX_WAIT:-15}"
+MTIME_FRESH="${MTIME_FRESH:-10}"
+MARKER_FILE="${MARKER_FILE:-/data/reload-config-applied.marker}"
```
leon-ape, line 10: MARKER_FILE defaults to /data, while Valkey physical backup archives the whole data directory. A restored pod with the same hostname/path/config checksum can reuse a stale marker and report reconfigure success from old state.

addons/valkey/dataprotection/post-restore-sentinel.sh
```diff
@@ -209,6 +213,29 @@ for sentinel_fqdn in "${sentinel_fqdn_list[@]}"; do
...
+
+if [ -n "${expected_sentinel_count}" ] && [ "${#reachable_sentinel_fqdn_list[@]}" -lt "${expected_sentinel_count}" ]; then
+  echo "ERROR: reached ${#reachable_sentinel_fqdn_list[@]}/${expected_sentinel_count} expected Sentinel pods." >&2
+  exit 1
+fi
+
+if [ -n "${expected_sentinel_count}" ]; then
+  sentinel_count="${expected_sentinel_count}"
+else
+  sentinel_count="${#reachable_sentinel_fqdn_list[@]}"
```
leon-ape, line 232: When SENTINEL_POD_FQDN_LIST and SENTINEL_REPLICA_COUNT are absent, sentinel_count falls back to only reachable pods. A partial probe can register quorum=1 for a larger Sentinel set and still let postReady succeed.

weicao:
## Leon review comments — response

### Comment 2: `reload-config.sh:10` — marker at `/data` could be stale after physical backup/restore

**Analysis**: The marker mechanism uses `hostname:path:cksum(config_file)` where `cksum` is computed from the ConfigMap-mounted config file (`/etc/conf/valkey.conf`), not from persistent data. After physical restore:

1. **Different config**: If the ConfigMap has changed since backup, `cksum` differs — marker mismatch at line 127 — marker is deleted — handled correctly.
2. **Same config**: If the ConfigMap has not changed, the marker matches — the script enters the **observation window** (5s) + **content polling** (15s) before exit 0. This is actually MORE conservative than the no-marker path (which can exit 0 immediately via the mtime-fresh shortcut). The engine runtime config will also match since Valkey loads config from the ConfigMap at startup.

The Phase 1 pre-check (comparing config file values against live `CONFIG GET`) is the real safety net — any runtime vs config mismatch is caught regardless of marker state.

**Conclusion**: The stale marker from backup does not cause a false positive that the no-marker path would not also produce. It actually triggers a more conservative code path. No code change needed.

### Comment 3: `post-restore-sentinel.sh:232` — partial probe could register quorum=1

**Analysis**: When both `SENTINEL_POD_FQDN_LIST` and `SENTINEL_REPLICA_COUNT` are absent, the fallback path sets `sentinel_count` from `reachable_sentinel_fqdn_list`. A partial probe (e.g. only 1 sentinel reachable out of 3) yields `quorum = (1/2)+1 = 1`, weakening failover safety. This is a regression from the dynamic quorum cherry-pick — before that change, quorum was hardcoded at 2.

**Fix**: Commit `42f8a92c` adds a quorum floor of 2 when `expected_sentinel_count` is unknown (the fallback path only). This preserves dynamic quorum for known-count scenarios while preventing quorum=1 from partial probes. Contract test added.

**Verification**: `bash -n` clean, shellspec 40/0 full suite green.

leon-ape (approved):
No finding.

PR #2797, opened 2026-06-09
fix(valkey): fail closed on dirty restore data dir

weicao:
## Problem

Valkey restore runs in the DataProtection `prepareData` phase and should only restore into an empty data directory, or a directory that contains only the `.kb-data-protection` placeholder. The current check treats the presence of the placeholder as enough to allow restore, even if real data files are also present.

## Observed behavior and reproduction

I added `addons/valkey/scripts-ut-spec/restore_contract_spec.sh` before changing `restore.sh`. The new spec creates a data directory with both `.kb-data-protection` and `appendonly.aof`, then runs the real `dataprotection/restore.sh` through a fake `datasafed` that emits a tar stream.

Before the fix, that mixed-data case failed the spec: `restore.sh` exited 0 and extracted `restored.txt` into a directory that already contained real data.

## Why this supports the judgment

The test exercises the script behavior, not just a grep/static check. It verifies the two allowed states still work:

- empty data directory
- placeholder-only data directory

It also verifies the unsafe state fails before extraction:

- placeholder plus real data

That is the exact contract boundary of this PR.

## Solution

`restore.sh` now checks for entries in `DATA_DIR` other than `.kb-data-protection` and exits non-zero if any are present. It also rejects a `.kb-data-protection` path that exists but is not a file before touching the placeholder.

## Validation

- Targeted TDD red before fix: mixed-data case restored into dirty directory.
- `shellspec --load-path ./shellspec --shell /opt/homebrew/bin/bash addons/valkey/scripts-ut-spec/restore_contract_spec.sh` => `3 examples, 0 failures`
- `shellspec --load-path ./shellspec --shell /opt/homebrew/bin/bash addons/valkey/scripts-ut-spec/*_spec.sh` => `224 examples, 0 failures`
- `bash -n addons/valkey/dataprotection/restore.sh addons/valkey/scripts-ut-spec/restore_contract_spec.sh`
- `git diff --check`

## Boundary

This PR is a focused script-contract fix for Valkey restore data-dir safety. It is not a runtime release-ready claim. Emily will run the agreed focused restore validation after PR A and PR B are merged.

### DIFF
--- a/addons/valkey/dataprotection/restore.sh
+++ b/addons/valkey/dataprotection/restore.sh
@@ -14,11 +14,15 @@ export DATASAFED_BACKEND_BASE_PATH="${DP_BACKUP_BASE_PATH}"
 mkdir -p "${DATA_DIR}"
 
 # Safety check: refuse to restore into a non-empty data directory.
-# Use -maxdepth 1 to check for any entry (file or directory) directly inside DATA_DIR.
+# Use -maxdepth 1 to check for any real data entry directly inside DATA_DIR.
 placeholder="${DATA_DIR}/.kb-data-protection"
-existing_entries=$(find "${DATA_DIR}" -mindepth 1 -maxdepth 1)
-if [ -n "${existing_entries}" ] && [ ! -f "${placeholder}" ]; then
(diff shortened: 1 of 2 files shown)


weicao:
Merged into #2796 per westonnnn request to combine both PRs. All changes cherry-picked to fix/valkey-review-cleanup branch.

PR #2809, opened 2026-06-11
fix(redis): ensure priority recovery on switchover failure path

weicao:
## Summary

Scoped down per wangyelei review + Leon approval to focus on one root cause only (Occam's razor).

### Root cause

`switchover_with_candidate()` used fail-fast (`|| return 1`) which skips `recover_redis_priorities` when `set_priorities`/`failover`/`result-check` fails, leaving replica-priority skewed for subsequent failover attempts.

### Changes

1. **redis-switchover.sh** — Refactor `switchover_with_candidate()` to capture rc at each step and always call `recover_redis_priorities` before returning, regardless of failure.
2. **redis-switchover.sh** — Guard `recover_redis_priorities` against unset `ORIGINAL_PRIORITIES` entries for pods not previously tracked.
3. **redis-cluster-switchover.sh / redis-cluster6-switchover.sh** — Fix missing space before `||` in `get_current_shard_primary()` (shell format fix).
4. **redis_switchover_spec.sh** — 4 shellspec tests covering priority recovery on failover failure, priority recovery failure, set_priorities failure, and result check failure.

### Not included (split to separate PRs per Occam's razor)

- Timeout tuning (300→120 + SWITCHOVER_WAIT_TIMEOUT env)
- Template timeoutSeconds / kbagent timeout override
- Uncomment check_switchover_result in switchover_without_candidate
- Redis6 cluster grep -v fail filter
- Cluster no-candidate false→true parameter
- reset-master.sh variable fix (covered by #2854)

## Test plan

- [ ] CI green (shellspec + chart checks)
- [ ] Focused switchover runtime test (new head, to be scheduled with Max)

### DIFF
--- a/addons/redis/redis-cluster-scripts/redis-cluster-switchover.sh
+++ b/addons/redis/redis-cluster-scripts/redis-cluster-switchover.sh
@@ -168,7 +168,8 @@ do_switchover() {
   fi
 
   # check switchover result
-  max_attempts=60
+  local wait_timeout=${SWITCHOVER_WAIT_TIMEOUT:-120}
+  local max_attempts=$((wait_timeout / 2))
   attempt=0
   while [ $attempt -lt $max_attempts ]; do
     role=$(check_redis_role "$candidate_pod_fqdn" $service_port)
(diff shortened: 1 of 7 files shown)


wangyelei:
these issue always fixed. only update the shell format, revert other codes

leon-ape (dismissed):

leon-ape (approved):
No finding.

PR #2810, opened 2026-06-11
chore(redis): move twemproxy startup command to ConfigMap-mounted script

weicao:
## Summary
- Twemproxy was the only Redis container with an inline startup command (`nutcracker -c ... -v 4 -m 16384`) in the cmpd runtime spec
- Move it to a dedicated ConfigMap-mounted script `redis-twemproxy-start.sh` for consistency with all other Redis containers (standalone, sentinel, cluster)
- This follows the best practice: main container startup scripts should be mounted via ConfigMap, not written directly in runtime commands/args

## Test plan
- [x] `bash -n` syntax check PASS
- [x] `git diff --check` PASS
- [x] helm template renders correctly with new script reference
- [x] Runtime: deploy replication-twemproxy topology, verify twemproxy container starts and proxies correctly
- [x] CI: check-chart, shell-check, shellspec-test all SUCCESS
- [x] Runtime: smoketest #44 PASS (8 pods, proxy roundtrip, pod restart resilience)
- [x] Runtime: parameters #46 PASS 33/0/0 (dynamic + static reconfigure through twemproxy)

### DIFF
--- a/addons/redis/scripts/redis-twemproxy-start.sh
+++ b/addons/redis/scripts/redis-twemproxy-start.sh
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec nutcracker -c /etc/proxy/nutcracker.conf -v 4 -m 16384
--- a/addons/redis/templates/cmpd-redis-twemproxy.yaml
+++ b/addons/redis/templates/cmpd-redis-twemproxy.yaml
@@ -90,10 +90,7 @@ spec:
     containers:
     - name: redis-twemproxy
       imagePullPolicy: {{ default .Values.redisTwemproxyImage.pullPolicy "IfNotPresent" }}
-      command:
-        - sh
-        - -c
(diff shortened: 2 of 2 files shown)


weicao (commented):
**Maya review: APPROVE** ✅

Clean and minimal. Two things I verified:

1. **Auto-glob**: `redis.extend.scripts` uses `$.Files.Glob "scripts/**"`, so `redis-twemproxy-start.sh` is automatically included in the ConfigMap — no template changes needed.

2. **`exec` is correct**: `exec nutcracker` replaces the shell with nutcracker as PID 1, so SIGTERM goes directly to nutcracker for graceful shutdown. This is actually an improvement over the original `sh -c nutcracker ...` which left a zombie-prone shell parent.

No issues found.

leon-ape (approved):

PR #2811, opened 2026-06-11
fix(elasticsearch): make memberLeave non-blocking

weicao:
## Summary
- Rewrite member-leave.sh to be non-blocking: sets shard allocation exclusion + voting config exclusion, then returns immediately (< 5s). Previously blocked synchronously for up to 35 minutes (shard migration wait + cluster green wait), violating the KB LFA < 1 minute contract.
- Add clear_stale_allocation_exclusion_for_self() to post-start-hook.sh so that when a pod restarts or scales back, it clears any leftover exclusion for its own name.
- Add 300s timeout to post-start-hook.sh wait_for_cluster_health() which previously had an unbounded while-true loop.

## Motivation
KubeBlocks lifecycle actions (LFA) are called synchronously -- the controller blocks until the action completes. The KB contract requires LFAs to be:
1. Idempotent
2. Re-entrant
3. Non-blocking (< 1 minute)

The old member-leave.sh violated requirement 3 with MAX_WAIT_TIME=1800 (30 min shard migration wait loop) + 300s green wait.

## Design
- **memberLeave**: health check -> voting config exclusion (if master-eligible) -> shard allocation exclusion -> return immediately. ES handles shard migration asynchronously.
- **Stale cleanup**: post-start-hook.sh runs on every pod start. If the pod name appears in cluster.routing.allocation.exclude._name, it removes itself from the exclusion list (handles scale-in -> scale-out and restart cases).
- **Safety**: exclusion prevents new shards from being allocated to the leaving node. When the pod is deleted, ES reassigns orphaned shards to remaining nodes.

## Test plan
- [ ] Static syntax: bash -n passes for all 10 ES scripts
- [ ] Runtime: scale-in triggers memberLeave, returns within seconds
- [ ] Runtime: scale-out after scale-in -- new pod starts, stale exclusion cleared by post-start-hook, shards allocated normally
- [ ] Runtime: restart scenario -- exclusion cleaned on pod restart

### DIFF
--- a/addons/elasticsearch/scripts/member-leave.sh
+++ b/addons/elasticsearch/scripts/member-leave.sh
@@ -1,30 +1,26 @@
 #!/bin/sh
-# https://www.elastic.co/guide/en/elasticsearch/reference/7.7/add-elasticsearch-nodes.html
-# Enhanced with safe scale-down process following Elasticsearch best practices
+# Non-blocking memberLeave lifecycle action.
+# Sets shard allocation exclusion and voting config exclusion, then returns
+# immediately. Shard migration happens asynchronously; memberJoin clears
+# stale exclusions when the same pod name rejoins.
 
 set -eu
 
-# Configuration
-MAX_WAIT_TIME=${MAX_WAIT_TIME:-1800}  # 30 minutes max wait time
 RETRY_COUNT=${RETRY_COUNT:-3}
-RETRY_DELAY=${RETRY_DELAY:-5}
-HEALTH_CHECK_INTERVAL=${HEALTH_CHECK_INTERVAL:-10}
 
-# Setup basic auth if credentials are available
 if [ -n "${ELASTIC_USER_PASSWORD:-}" ]; then
   BASIC_AUTH="-u elastic:${ELASTIC_USER_PASSWORD}"
 else
   BASIC_AUTH=''
 fi
 
-# Check if we are using IPv6
 if echo "${POD_IP:-}" | grep -q ':'; then
   LOOPBACK="[::1]"
 else
   LOOPBACK=127.0.0.1
 fi
 
-if [ "${TLS_ENABLED:-false}" == "true" ]; then
+if [ "${TLS_ENABLED:-false}" = "true" ]; then
   READINESS_PROBE_PROTOCOL=https
 else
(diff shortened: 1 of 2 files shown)


weicao:
## Review feedback fixes (commit a29eb4e8)

Addressing @Jade review (REQUEST_CHANGES):

### Fix 1: Node-not-found jq crash (BLOCKER)
**Problem**: `_nodes/${node_name}` returns `{"nodes":{}}` when target node already removed. jq expression `.nodes | to_entries[0].value.roles | contains(["master"])` crashes on empty nodes. With `set -e`, script exits before reaching shard exclusion logic, breaking LFA re-entry.
**Fix**: Check `node_count` first. If 0, log and skip voting exclusion, set `is_master=false`, continue to shard exclusion.

### Fix 2: Curl budget exceeding action timeout (BLOCKER)
**Problem**: `--max-time 30 --retry 3` means single curl worst-case ~90s. With 5+ API calls, total could reach ~7.5 minutes, exceeding kbagent action timeout.
**Fix**: Changed to `--connect-timeout 5 --max-time 10 --retry 1`. Worst-case per curl = 20s, 5 calls = ~100s. Removed unused `RETRY_COUNT` variable.

### Issue 3: Concurrent memberLeave race (addressed with contract evidence)
KB controller calls memberLeave sequentially in a for-loop (`transformer_component_workload_ops.go:156`), not concurrently. The read-append-write pattern is safe under this sequential execution guarantee.

weicao:
## Review feedback fixes round 2 (commit 1da1ba5c)

Addressing @Jade re-review on curl budget (Issue 3):

**Problem**: `--max-time 10 --retry 1` means single curl worst-case ~20s. With 6-9 API calls, total could exceed kbagent 30s default action timeout.
**Fix**: Changed to `--connect-timeout 2 --max-time 3` with no retry. Worst-case: data node 6 calls × 3s = 18s, master node with voting retry 9 calls × 3s = 27s — both within 30s default.

Also separated all `curl | jq` pipes into two steps (curl into variable, then jq on variable) to prevent pipeline masking curl errors under `set -eu` (POSIX sh has no pipefail).

All CI checks green. @Jade APPROVE on `1da1ba5c`.

PR #37, opened 2023-11-21
chore: update etcd follower default AccessMode

Y-Rookie:
fix: https://github.com/apecloud/kubeblocks/issues/5880

PR #2818, opened 2026-06-11
fix(redis): revert reconfigure action to env-var iteration

weicao:
## Summary
- PR #2794 (`fcdf51a46`) changed Redis reconfigure exec from env-var iteration to `$1`/`$2` argv pattern with `-- ` separator
- Live kbagent controller (KB main/1.2+) does **not** pass parameter name/value as positional args after `--` — it sets file-level change env vars per addon-api contract (`docs/addon-api/08-parameters-and-config.md`)
- Under `set -eu`, `$1` is unbound → every dynamic reconfigure OpsRequest fails with `$1: unbound variable`
- Reverts to env-var iteration: read param names from env, convert `_` to `-` for Redis hyphenated param names, call `reload-parameter.sh` per param

## Root cause evidence
- Task #46 runner-r2 first-red: A02 `maxmemory-policy` OpsRequest timeout
- Controller stderr: `udf-reconfigure-cmpd-redis-replication-config: -- : line 3: $1: unbound variable`
- runner-r2.log sha256: `f9ae7b81c477b6c3d62b38a673a97aa09790c820730356674f9e1f3b89834192`

## Affects
All Redis topologies using `redis.config.reconfigureAction` template: standalone, replication, repl-twemproxy, cluster.

## Test plan
- [ ] Task #46 rerun (parameters × repl-twemproxy) with this fix patched — target PASS
- [ ] Regression check: existing parameters PASS results (task #37 replication, task #40 cluster) remain valid since env-var approach is the original contract-compliant method

### DIFF
--- a/addons/redis/templates/_helpers.tpl
+++ b/addons/redis/templates/_helpers.tpl
@@ -206,9 +206,10 @@ reconfigure:
       - -c
       - |
         set -eu
-
-        /scripts/reload-parameter.sh "$1" "$2"
-      - --
+        env | cut -d= -f1 | grep -E '^[a-zA-Z0-9_.-]+$' | sort -u | while IFS= read -r param; do
+          [ -n "${param}" ] || continue
+          /scripts/reload-parameter.sh "${param//_/-}" "$(printenv "${param}")"
+        done
 {{- end -}}
 
 {{- define "apeDts.reshard.image" -}}


weicao:
**Focused review: APPROVE (Slock-side, cannot GitHub-approve own account)**

Review checklist:
1. ✅ Single file change: `addons/redis/templates/_helpers.tpl`
2. ✅ Exact revert of PR #2794 Redis changes — file index restores to `162578ccd`
3. ✅ Env-var iteration preserved: `env | cut | grep | sort -u | while read param`
4. ✅ Hyphen conversion preserved: `${param//_/-}` (e.g., MAXMEMORY_POLICY → maxmemory-policy)
5. ✅ `targetPodSelector: All` unchanged
6. ✅ `set -eu` preserved
7. ✅ `--` arg separator correctly removed
8. ✅ `reload-parameter.sh` interface $1=paramName $2=paramValue matches wrapper call

Known limitation: env-var scan covers all vars including system ones; Redis silently rejects unknown config keys. Same pre-#2794 production behavior. wangyelei's formal fix is separate.

PR #2808, opened 2026-06-10
chore(redis): add error handling to account provision scripts

weicao:
## Summary

- **Bug fix**: `redis-account-provision.sh` and `redis-sentinel-account-provision.sh` had no error handling. If the `redis-cli` ACL statement failed (connection error or Redis `ERR` response), the script continued to `acl save` and could exit 0, silently reporting the account was created when it was not.
- **Cleanup**: Remove duplicate `CLUSTER_NAMESPACE` env entry in `redis-shard-account-ops` OpsDefinition (copy-paste artifact, harmless but messy).
- Both provision scripts now check `redis-cli` exit code and scan output for `ERR` prefix before proceeding to `acl save`.
- Scripts converted from `sh` to `bash` for consistency with testing framework; cmpd commands updated accordingly.
- 8 new shellspec tests covering: success path, ERR output, connection failure, and acl save failure.

## Validation

- `bash -n` all changed scripts: PASS
- `git diff --check`: PASS
- Focused shellspec `8/0/0`
- Full Redis shellspec suite `206/0/0`

## Test plan

- [ ] Runtime acceptance: accounts OpsRequest create/delete on standalone topology
- [ ] Runtime acceptance: accounts OpsRequest create/delete on replication topology
- [ ] Verify error path: intentionally wrong password triggers provision failure (not silent success)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2819, opened 2026-06-11
fix(redis): rework reconfigure action to file-based diff with freshness guard

weicao:
## Summary

Rework Redis reconfigure action to use the rendered config file per addon-api contract, with a freshness guard and retry tolerance.

**Problem**: The original reconfigure script reads rendered \`/etc/conf/redis.conf\`, diffs against \`CONFIG GET *\`, and applies changed dynamic params. This is vulnerable to kubelet ConfigMap projected-volume sync delay (up to ~60s), where stale file = engine state → 0 diff → exit 0 → OpsRequest false green. Additionally, kbagent \`targetPodSelector: All\` retries the action; after a successful first apply, the retry sees 0-diff + stale \`..data\` mtime → freshness check blocks indefinitely.

**Fix (3 layers)**:
1. **File-based diff with dynamic allowlist** — read rendered config, intersect with \`DYNAMIC_ALLOWLIST\`, diff against engine \`CONFIG GET *\`, apply only changed dynamic keys, readback-verify each.
2. **Freshness guard** — on 0-diff with checkable params, check \`..data\` symlink mtime; if stale, bounded wait (15s default) for kubelet sync; fail-closed on timeout.
3. **Applied-marker retry tolerance** — after successful apply (≥1 key, rc=0), write pod-local \`/tmp\` marker bound to \`<config fingerprint>|<..data target>\`; on 0-diff retry for same mounted generation, consume marker once and return 0; marker is TTL-bounded (120s) and deleted after consumption.

## Changes

- \`redis-reconfigure-config.sh\`: file-based reconfigure with freshness guard and applied-marker retry tolerance (252 lines)
- \`_helpers.tpl\`: reconfigure action template uses file-based approach (no argv)
- \`redis_reconfigure_config_spec.sh\`: ShellSpec tests (44 examples, 0 failures)

## Evidence

- ShellSpec: 44/0 (bash 5.3.9)
- CI: 7/7 SUCCESS
- Runtime gate: Parameters × repl-twemproxy focused gate PASS 34/0/0 on exact head \`7a3e70ea\`
  - A02 \`maxmemory-policy -> allkeys-lru\`: OpsRequest Succeed, kbagent retry tolerant
  - A03 \`hz -> 50\`: OpsRequest Succeed, pod UID unchanged
  - A04 \`databases -> 32\` (static): OpsRequest Succeed, pod UID changed (restart)
  - Cleanup: namespace NotFound, PV residue 0
- Evidence package sha256: \`9d22f05af1c515df8621fee6d71c2f434d6ff991ac00ce2a9b08e2f5fcd9e619\`
- Controller baseline: \`kubeblocks:pr-10362-bccd658-stella\`


addons/redis/scripts/redis-reconfigure-config.sh
```diff
@@ -0,0 +1,131 @@
...
+reconfigure_from_config_file() {
+  if [ ! -f "$config_file" ]; then
+    echo "ERROR: rendered config not found: $config_file" >&2
+    return 1
+  fi
+
+  _rcf_hash=$(config_file_hash)
+  echo "INFO: reconfigure start, config hash=$_rcf_hash" >&2
+
+  apply_config_diff
```
leon-ape, line 248: `reconfigure_from_config_file` calls `apply_config_diff` before proving the projected ConfigMap file is fresh; if the mounted file is stale and stale values are applied successfully, the action can return 0 after timeout without applying the current desired config.

weicao:
**回复 Leon review comment (3432782443)：**

已修复：`af10338a0` 重排 `reconfigure_from_config_file` 的执行顺序。

**问题**：`apply_config_diff` 在等待 ConfigMap projected volume 刷新之前就执行，如果挂载文件是旧的，旧值被成功 apply 后，timeout 期间文件不变，action 返回 0 但实际没有应用最新配置。

**修复**：先等待 config file hash 变化（证明 kubelet 已投递新内容），然后才 `apply_config_diff`。如果 timeout 未观察到 hash 变化，则在 timeout 后才执行 apply（此时文件要么本来就是最新的，要么 kubelet 已在等待期间同步完成）。false green 防护（0 params applied + timeout > 0）保持不变。

本地 shellspec 37 examples / 0 failures (bash 5.3)。

weicao:
**Leon comment 修复 + 运行验证更新：**

Leon comment (stale config apply order) 已修复：`af10338a0` 重排为 wait-then-apply。

运行验证：
- Aria focused review: PASS (37 shellspec / 0 failures)
- Max focused gate: PASS 34/0/0
  - Pin: #2819 `af10338a0` + #2810 `29ad77f1f` + controller `kubeblocks:pr-10401-5c5c2a0-stella`
  - 3 OpsRequests all `Succeed`；generation/configHash converged；dynamic maxmemory-policy/hz kept pod UID；static databases rolled；CONFIG GET + twemproxy SET/GET passed；cleanup clean
  - Evidence sha: `9958f9802c4f2be62081176dad281250f3df68d1684709e51c7c6321b29ef1e6`

Controller 依赖：#10401 已 merge (`e296fcd435fa`)。

addons/redis/scripts/redis-reconfigure-config.sh
```diff
@@ -0,0 +1,132 @@
...
+      return $?
+    fi
+  done
+
+  if [ "$_rcf_timeout" -gt 0 ]; then
+    echo "INFO: config file unchanged after ${_rcf_timeout}s" >&2
+  fi
+
+  apply_config_diff
+  _rcf_rc=$?
```
leon-ape: After the watch times out, this still applies the unchanged rendered file; if that projection is stale and contains changed values, the action can return success after applying stale config.

addons/redis/config/redis-config-effect-scope.yaml
```diff
@@ -4,3 +4,63 @@ staticParameters:
...
+  - replica-read-only
+  - replica-serve-stale-data
+  - set-max-intset-entries
+  - set-proc-title
+  - slowlog-log-slower-than
+  - slowlog-max-len
+  - stop-writes-on-bgsave-error
+  - stream-node-max-bytes
+  - stream-node-max-entries
+  - tcp-backlog
```
leon-ape: `tcp-backlog` is immutable in Redis, but this declares it dynamic; KB will route changes through reconfigure/CONFIG SET instead of a restart.

weicao:
**回复 Leon review comments (3434232803, 3434240081)：**

**1. comment 3434232803 — timeout 后仍 apply 未变的 rendered file**
已在 `af10338a0` 修复。修改后的逻辑：先等 config file hash 变化，hash 变了才 apply；如果 timeout 内 hash 没变，则 timeout 后 apply，但有 false green 防护：如果 timeout > 0 且 applied count = 0，返回错误（说明 kubelet 没投递新配置，且旧配置和当前 Redis 状态一致）。这样不会在文件陈旧时静默返回成功。

**2. comment 3434240081 — tcp-backlog 误标为 dynamic**
已修复：`6b1837c1` 将 `tcp-backlog` 从 `dynamicParameters` 移到 `staticParameters`。tcp-backlog 是 Redis 启动时设置的 TCP listen() backlog，运行时无法通过 CONFIG SET 修改。标为 dynamic 会导致 KB 走 CONFIG SET 路径失败。

当前 HEAD：`6b1837c1`

addons/redis/scripts/redis-reconfigure-config.sh
leon-ape: This path still applies the current mounted file after the wait timeout. If that file is a stale projection and differs from Redis current state, the action can still apply stale values and return success.

addons/redis/scripts/redis-reconfigure-config.sh
weicao: 已修复：`af10338a0` → 现在是 `53e37d44`（rebase 后）。重排为 wait-then-apply：先等 config file hash 变化，hash 变了才 apply diff。如果 timeout 内 hash 没变，timeout 后 apply 但有 false green 防护（0 params + timeout > 0 = fail）。

addons/redis/scripts/redis-reconfigure-config.sh
weicao: 同上修复。wait-then-apply 逻辑保证：timeout 后如果 rendered file 没变且 apply 了 0 个参数，返回错误而不是成功。不会静默 apply 陈旧配置后返回 0。

addons/redis/config/redis-config-effect-scope.yaml
weicao: 已修复：`6b1837c1` → 现在是 `b2c8a5a3`（rebase 后）。tcp-backlog 从 dynamicParameters 移到 staticParameters。Redis 的 tcp-backlog 只能在启动时设置，运行时 CONFIG SET 会失败。

addons/redis/scripts/redis-reconfigure-config.sh
leon-ape: Still unresolved: the timeout path still applies the unchanged mounted file. If that projection is stale and differs from Redis current state, _rcf_applied_count becomes greater than 0 and the action returns success after applying stale config.

weicao:
**Rebase + Leon comment fixes + exact-head runtime update：**

HEAD：`b2c8a5a3`（rebase on main after #2808/#2854 merge + tcp-backlog fix）

Leon comments 修复：
1. (3432782443) stale config apply → wait-then-apply（`53e37d44`）
2. (3434232803) timeout false green → 0 params + timeout > 0 = fail
3. (3434240081) tcp-backlog → 从 dynamicParameters 移到 staticParameters（`b2c8a5a3`）

3 条 review threads 全部 resolved。

运行验证（exact-head `b2c8a5a3`）：
- Aria focused review: PASS（37 shellspec / 0 failures，diff/CI/boundary check）
- Max runtime focused gate: PASS 34/0/0
  - Pin: #2819 `b2c8a5a3` + #2810 `29ad77f1f` + combo `7b79b0ab` + controller #10401 `5c5c2a060`
  - 3 OpsRequests `Succeed`；dynamic maxmemory-policy/hz pod UID 不变；static databases 滚动重建；CONFIG GET readback 收敛；cleanup clean
  - Evidence sha: `e7b535aa1d1713078985f36a4292e232102c9b5dcea54d8a2f5f1a5dbde007c0`
- Baseline restored: controller `pr-10362-bccd658-stella`, addon #2808 `ce7522199`

CI：all checks SUCCESS（shellspec-test, shell-check, chart checks, pr-label-check）
Controller 依赖：#10401 已 merge（`e296fcd435fa`）

leon-ape:
Still unresolved: the timeout path returns success when the unchanged mounted file applies at least one parameter, so a stale projection can still be applied successfully.

addons/redis/scripts/redis-reconfigure-config.sh
weicao: This concern is now moot — the entire bounded wait / timeout / config_file_hash() logic has been removed in `253b5bf7`. The script no longer has a timeout path that could apply a stale file.

What remains: read rendered config → diff with CONFIG GET → CONFIG SET changed dynamic params → post-set verify. No wait loop, no hash polling, no false-green guard.

If kubelet projected-volume stale reads need to be addressed, that will be a separate follow-up PR with runtime evidence of actual ConfigMap sync delay causing reconfigure failures.

weicao: Same resolution as above — the entire timeout / bounded wait path has been removed in `253b5bf7`. There is no longer any code that applies an unchanged mounted file after a wait timeout.

The stale projection concern was valid for the old code, but the fix is to remove that code path entirely (no runtime evidence it was needed), not to patch it.

addons/redis/scripts/redis-reconfigure-config.sh
```diff
@@ -0,0 +1,115 @@
...
+  return "$_acd_rc"
+}
+
+reconfigure_from_config_file() {
+  if [ ! -f "$config_file" ]; then
+    echo "ERROR: rendered config not found: $config_file" >&2
+    return 1
+  fi
+
+  apply_config_diff
```
leon-ape, line 248: This still applies the mounted config immediately without proving the projected ConfigMap is fresh. kbagent can invoke reconfigure before kubelet updates the projection, so a stale redis.conf that differs from engine state can still be applied and return success.

weicao: The freshness guard is inside `apply_config_diff()` (lines 228–242), but it only triggers on the **0-diff** path. Your concern is about the **has-diff** path — let me clarify why that path is safe.

**Normal flow (no prior engine drift):**
Before OpsRequest, engine state matches the old ConfigMap. If kubelet has not yet synced the new ConfigMap, the stale file also matches old ConfigMap → file values == engine values → 0-diff → freshness guard catches it. So the stale-file-with-diffs scenario does not arise in normal flow.

**Abnormal flow (engine state previously diverged from ConfigMap):**
If something externally changed a parameter so engine state ≠ old ConfigMap, then a stale file could produce diffs. In that case, applying the stale file values converges the engine toward the last known ConfigMap state — a transient intermediate state. When kubelet syncs the new ConfigMap, the controller retries, and the correct values get applied.

**Why not check freshness before all diffs:**
`..data` mtime only proves "kubelet synced recently", not "this specific ConfigMap revision has arrived." After a long idle period, mtime is always stale even though the file content is correct. Adding a freshness check before diffs would add 15s latency to every normal reconfigure without improving correctness.

**Safety net:** every `CONFIG SET` is followed by a readback (`verify_engine_state`, lines 62–71) that confirms the engine actually accepted the new value.

weicao: You're right — my "controller retries after kubelet syncs" argument was wrong. If the script returns success, the OpsRequest is marked done and there is no automatic retry. A stale projection that happens to diff against engine state (e.g. from manual CONFIG SET) would silently apply stale values and report success.

Fixed in `8a13a6d`: moved `ensure_projected_config_fresh()` from a post-diff 0-diff-only check inside `apply_config_diff()` to a pre-check in `reconfigure_from_config_file()`. The script now verifies projection freshness **before** reading the file for diffs.

Normal case cost: one `stat` + one `readlink` (~<1ms, mtime within `projection_fresh_age_seconds`). Stale case: waits up to `projection_wait_seconds` for kubelet to sync, then proceeds with fresh file or returns error (retry-safe). Applied-marker retry tolerance is preserved as fallback when freshness times out.

45/0 shellspec PASS, including new test "stale projection with diffs is blocked" and "fresh projection with diffs applies normally".

weicao:
Fixed in `8a13a6d`. Moved freshness check to a pre-check in `reconfigure_from_config_file()` — the script now verifies projection freshness before reading the file for diffs. A stale projection that differs from engine state (your scenario) is now blocked, not applied. See review thread reply for details.

addons/redis/scripts/redis-reconfigure-config.sh
```diff
@@ -0,0 +1,255 @@
...
+    _epcf_link=$(readlink_target "$_epcf_data_link")
+    _epcf_after=$(config_fingerprint "$config_file") || _epcf_after=""
+    if [ "$_epcf_link" != "$_epcf_initial_link" ] || [ "$_epcf_after" != "$_epcf_current" ]; then
+      projection_changed=true
+      echo "INFO: projected config changed after ${_epcf_waited}s; proceeding with reconfigure" >&2
+      return 0
+    fi
+  done
+
+  echo "ERROR: projected config did not refresh after ${projection_wait_seconds}s: dataLink='${_epcf_initial_link}', age='${_epcf_age}', mounted='${_epcf_current}', retry-safe: yes" >&2
```
leon-ape, line 187: This rejects any projected config whose ..data mtime is older than REDIS_RECONFIGURE_PROJECTION_FRESH_AGE_SECONDS unless the symlink/content changes during the wait. A delayed or retried reconfigure can have the correct mounted file with an old mtime, so the action can fail without applying a valid config.

addons/redis/config/redis-config-effect-scope.yaml
```diff
@@ -4,3 +4,63 @@ staticParameters:
   - databases
   - io-threads
   - io-threads-do-reads
+  - tcp-backlog
+
+## Parameters that can be changed at runtime via CONFIG SET
+dynamicParameters:
```
wangyelei, line 10: If only the static parameters are set, the rest will default to dynamic. Since I don't want to list out every single parameter, I'm going to close this PR.

PR #2842, opened 2026-06-17
test(redis): add shellspec tests for reload-parameter.sh

weicao:
## Summary
- Wrap top-level logic in `reload_redis_parameter()` function with ut_mode guard for shellspec testability
- Add 9 shellspec test examples covering:
  - Parameter parsing: simple key-value, multi-word value, empty-string value, split positional args
  - Redis-cli integration: with/without password, custom port, TLS flags, failure propagation

## Test plan
- [ ] CI shellspec-test workflow passes
- [ ] CI shellcheck passes
- [ ] No regression in existing Redis shellspec tests


weicao:
Superseded by #2853 (combined shellspec test PR per Leon's review guidance)

PR #2856, opened 2026-06-18
chore(mongodb): read PBM path style from datasafed config

weicao:
## Summary

- Parse `force_path_style` from `/etc/datasafed/datasafed.conf` and use it for PBM `storage.s3.forcePathStyle`.
- Keep the existing Minio provider fallback when `force_path_style` is absent, so old datasafed configs keep current behavior.
- Add MongoDB shellspec coverage for explicit `force_path_style=true`, explicit `force_path_style=false`, and legacy Minio fallback.

## Why

MongoDB PBM backup config previously inferred S3 path-style behavior from `provider == Minio`. That is too rigid for S3-compatible providers because datasafed is rclone-based, and rclone exposes this as the `force_path_style` config option. Prefer the explicit datasafed/rclone config when it is present; fall back to provider inference only for backwards compatibility.

## Verification

- `bash -n addons/mongodb/dataprotection/common-scripts.sh addons/mongodb/scripts-ut-spec/dataprotection_common_spec.sh`
- `shellspec --load-path ./shellspec --shell bash addons/mongodb/scripts-ut-spec/dataprotection_common_spec.sh`
- `shellspec --load-path ./shellspec --shell bash addons/mongodb/scripts-ut-spec`
- `git diff --cached --check`

## Boundary

Offline script/unit coverage only. No Kubernetes backup runtime was launched from this PR branch.

### DIFF
--- a/addons/mongodb/dataprotection/common-scripts.sh
+++ b/addons/mongodb/dataprotection/common-scripts.sh
@@ -62,8 +62,24 @@ function getToolConfigValue() {
   cat "$toolConfig" | grep "$var" | awk '{print $NF}'
 }
 
+function normalizeBoolValue() {
+  local value
+  value=$(echo "$1" | tr '[:upper:]' '[:lower:]')
+  case "$value" in
+    true | 1 | yes | y | on)
+      echo "true"
+      ;;
+    false | 0 | no | n | off)
+      echo "false"
+      ;;
+    *)
+      echo "$value"
+      ;;
+  esac
+}
+
 function set_backup_config_env() {
-  toolConfig=/etc/datasafed/datasafed.conf
+  toolConfig=${DATASAFED_CONFIG_FILE:-/etc/datasafed/datasafed.conf}
   if [ ! -f ${toolConfig} ]; then
     DP_error_log "Config file not found: ${toolConfig}"
     exit 1
@@ -75,6 +91,7 @@ function set_backup_config_env() {
   local region=""
   local endpoint=""
   local bucket=""
+  local force_path_style=""
 
   IFS=$'\n'
   for line in $(cat ${toolConfig}); do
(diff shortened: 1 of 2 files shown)


addons/mongodb/dataprotection/common-scripts.sh
```diff
@@ -119,6 +138,9 @@ function set_backup_config_env() {
   else
     echo "Unsupported provider: $provider"
   fi
+  if [ -n "$force_path_style" ]; then
+    export S3_FORCE_PATH_STYLE="$(normalizeBoolValue "$force_path_style")"
```
leon-ape, line 142: This parses force_path_style, but sync_pbm_storage_config still skips existing PBM configs by comparing only endpoint/region/bucket/prefix. If datasafed force_path_style changes while those fields stay the same, PBM keeps the old forcePathStyle value.

weicao: Fixed in `1e9dff52`. `sync_pbm_storage_config` now reads current `.storage.s3.forcePathStyle` and only skips PBM reconfiguration when endpoint/region/bucket/prefix and forcePathStyle all match. Added a shellspec regression where only forcePathStyle differs and verified the generated PBM config is rewritten with the new value.

Verification:
- `bash -n addons/mongodb/dataprotection/common-scripts.sh addons/mongodb/scripts-ut-spec/dataprotection_common_spec.sh`
- `shellspec --load-path ./shellspec --shell bash addons/mongodb/scripts-ut-spec/dataprotection_common_spec.sh` (4/0)
- `shellspec --load-path ./shellspec --shell bash addons/mongodb/scripts-ut-spec` (7/0)
- `git diff --check`

leon-ape (approved):

PR #2864, opened 2026-06-19
fix(redis): fail closed on stale reconfigure projection

weicao:
## Summary
- narrow Redis rendered-config reconfigure freshness proof to the no-op success path
- apply mounted-file dynamic diffs first and verify Redis readback, so old-but-correct stable projections are not blocked by mtime age
- fail closed before `INFO: applied 0` when the mounted file is a stale no-op and current projected-volume freshness cannot be proven
- remove the persistent marker shortcut because it proves only that an old mounted file was applied before, not that the current request target is that file
- rerun diff/apply when `/etc/conf/..data` or mounted content moves during the bounded wait
- add focused ShellSpec coverage for target-diff apply, stale no-op fail-closed, stale-marker reuse fail-closed, fresh projection, missing `..data`, and projection movement

## Evidence
- `sh -n addons/redis/scripts/redis-reconfigure-config.sh`
- `git diff --check -- addons/redis`
- `git diff --check origin/main...HEAD -- addons/redis`
- `/opt/homebrew/bin/shellspec --load-path ./shellspec --shell /opt/homebrew/bin/bash addons/redis/scripts-ut-spec/redis_reconfigure_config_spec.sh` => 43 examples, 0 failures
- `helm dependency build addons/redis`
- `helm lint addons/redis`
- `helm template redis addons/redis`
- rendered output has no `timeoutSeconds:-1`, no `timeoutSeconds:180`, no `RECONFIGURE_WAIT_TIMEOUT`
- direct `/bin/sh` target-diff probe: old `..data`, mounted target `allkeys-lru`, live Redis `volatile-lru` -> applies `maxmemory-policy=allkeys-lru`
- direct `/bin/sh` stale-marker-reuse probe: old `..data`, mounted/live Redis `volatile-lru`, stale marker file present -> rc=1 before `INFO: applied 0`
- GitHub checks all green on head `5061380ca0ade20d8533d2d7e22f7951e460bfc2`, including `shellspec-test`
- Slock-side focused reviews: Jade APPROVE and Blake APPROVE on head `5061380ca0ade20d8533d2d7e22f7951e460bfc2`

## Runtime evidence
Focused runtime gate passed on Saiyin:
- scope: Redis Parameters × replication-twemproxy
- head: `5061380ca0ade20d8533d2d7e22f7951e460bfc2`
- controller image: `docker.io/library/kubeblocks:pr-10401-5c5c2a0-stella`
- controller imageID: `sha256:f4584d503aba16f3cc28d4057dbba319e6c73a71607d73f94e0314853bdb51d1`
- result: PASS 34 / FAIL 0 / SKIP 0
- evidence package: `redis-pr2864-5061380ca-parameters-twemproxy-20260620T0700.tgz`
- sha256: `2dab47966b06d987e88759750724bdec3d43f4d71b1a07343d10d0dd0bf31fdd`
- cleanup: namespace NotFound, no PV residue

The previously observed #2819 failure was: OpsRequest Succeed, rendered target `allkeys-lru`, live Redis stayed `volatile-lru` until manual rerun after mount sync. This PR's exact-head focused runtime shows the dynamic `maxmemory-policy -> allkeys-lru` case reaches live Redis values without Redis pod UID changes and without kbagent `timedOut`.

## Boundary
This is the exact-head focused gate for task #60 / PR #2864, not full Redis release acceptance. Base is #2819 branch `redis-reconfigure-config-file-contract`, so this PR only shows the freshness delta.

### DIFF
--- a/addons/redis/scripts-ut-spec/redis_reconfigure_config_spec.sh
+++ b/addons/redis/scripts-ut-spec/redis_reconfigure_config_spec.sh
@@ -323,6 +323,8 @@ Describe "Redis Reconfigure Config Script Tests"
       freshness_check=false
       projection_fresh_age_seconds=10
       projection_wait_seconds=0
+      applied_marker_file=""
+      applied_marker_max_age_seconds=600
       tmp_config=$(mktemp)
       config_file="$tmp_config"
     }
@@ -362,6 +364,7 @@ Describe "Redis Reconfigure Config Script Tests"
         tmp_dir=$(mktemp -d)
         tmp_config="$tmp_dir/redis.conf"
         config_file="$tmp_config"
+        applied_marker_file="$tmp_dir/marker"
         ln -s "..2026_06_20_00_00_00.000000000" "$tmp_dir/..data"
 
         set_called_with=""
@@ -421,6 +424,7 @@ Describe "Redis Reconfigure Config Script Tests"
         tmp_dir=$(mktemp -d)
         tmp_config="$tmp_dir/redis.conf"
         config_file="$tmp_config"
+        applied_marker_file="$tmp_dir/marker"
         ln -s "..2026_06_20_00_00_00.000000000" "$tmp_dir/..data"
 
         set_called_with=""
@@ -466,10 +470,11 @@ Describe "Redis Reconfigure Config Script Tests"
         The variable set_called_with should equal "maxmemory-policy=allkeys-lru;"
         The stderr should include "INFO: applied 1 parameter"
         The stderr should not include "ERROR: projected config did not refresh"
+        The path "$tmp_dir/marker" should be file
       End
     End
 
-    Context "old stable already-applied file has stale marker"
+    Context "old stable already-applied file has an expired marker"
       set_called_with=""
       tmp_dir=""
       mock_now=100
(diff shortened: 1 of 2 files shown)


weicao:
## Maya focused review — rebased head aa7799fc on #2819 37586a41

**Result: PASS — no blockers.**

**Scope**: 2 files, +457 lines (script + tests). No template or chart changes.

**Freshness guard logic (ensure_projected_config_fresh)**:
- Post-diff lazy check — only triggers when 0 diffs found with checkable params. Does not block when file already differs from engine. Correct design for kubelet sync delay.
- Fail-closed on stale: if ..data symlink has not changed within projection_wait_seconds, returns 1 with retry-safe: yes. kbagent retries.
- Dual detection: kubelet atomic ..data symlink target + cksum fingerprint — two independent signals.
- Bounded recursion: max depth 2. Second apply_config_diff call either finds actual diffs or sees fresh mtime (returns 0, no further recursion).

**Edge cases verified**:
- Error early-exit (line 169-172): CONFIG SET failure returns immediately without entering freshness check. Does not mask real errors.
- _rcf_checkable_count == 0: skips freshness check. Correct.
- file_mtime(): stat -c %Y (Linux) then stat -f %m (macOS) fallback. POSIX portable.
- config_fingerprint() uses cksum — available everywhere.

**Tests**: 43/0 PASS (local bash 5.3.9). 8 new tests: fresh mtime, stale+timeout, stale+changes-during-wait, no-symlink, stale-equal-engine, different-from-engine, already-applied-marker, projection-changes-during-wait.

weicao:
## Focused Review: `7a3e70ea` — marker-based retry tolerance

**Verdict: PASS, no blocker.**

### Checklist

| Item | Result |
|---|---|
| Marker write only after verified apply (rc=0, applied>0) | PASS |
| Token = `<fingerprint>\|\<..data target>` — both dimensions bound | PASS |
| 0-diff path: marker checked before freshness wait | PASS |
| Does not reopen 'old projected = engine' false-green | PASS |
| Consume-once + 120s TTL + pod-local `/tmp` | PASS |
| Atomic write (tmp+mv), failure swallowed (advisory) | PASS |
| ShellSpec 44/0 (bash 5.3.9) | PASS |
| `sh -n` / `git diff --check` / `helm template` | PASS |
| CI 7/7 SUCCESS | PASS |
| Diff scope: 2 files only (script + tests) | PASS |

### Minor nit (non-blocking)

`consume_applied_marker` fallback `${applied_marker_max_age_seconds:-600}` differs from init default (120). Variable is always set by `init_reconfigure_env()` so the fallback never fires; just noting the asymmetry.

Ready for runtime gate.

Issue #331, opened 2024-02-27
[BUG] failed to reconfigure log variables of target cluster for mysql addon


sophon-zt:
```
# kbcli cluster update mysqltest --enable-all-logs=true -n default
error: failed to reconfigure log variables of target cluster: template: template-logs-block:88:47: executing "logsBlock" at <$.podSpec.containers>: map has no entry for key "podSpec"


```

PR #2865, opened 2026-06-20
chore(valkey): add targetPodSelector to accountProvision

weicao:
## Summary
- accountProvision was missing `targetPodSelector: Role` / `matchingKey: primary`, which could cause account provisioning to execute on a replica where ACL changes are local-only and won't propagate to the primary.
- postProvision already correctly uses these fields. This aligns accountProvision with the same pattern.

## Test plan

### DIFF
--- a/addons/valkey/templates/cmpd.yaml
+++ b/addons/valkey/templates/cmpd.yaml
@@ -446,6 +446,8 @@ spec:
           - sh
           - -c
           - /scripts/account-provision.sh
+        targetPodSelector: Role
+        matchingKey: primary
 
     # switchover: when Sentinel is present, delegates to SENTINEL FAILOVER.
     # When absent (standalone), performs a direct REPLICAOF NO ONE.


weicao:
Code-contract review for W1:

No code blocker found for the selector change. The patch places `targetPodSelector: Role` and `matchingKey: primary` under `accountProvision.exec`, matching the existing `postProvision.exec` selector. This aligns with `docs/addon-api/07-accounts-and-tls.md`: accountProvision should run on a pod that can execute authorization, and primary/leader is the normal primary/replica target. It also aligns with `docs/addon-api/05b-lifecycle-action-fields.md`: `Role` uses `matchingKey`.

XP design-contract walk:
- Class 1: no silent fallback added; this removes the previous implicit/random target fallback.
- Class 2: target role contract is now explicitly written at the action site.
- Classes 3-8: no state-continuity, sentinel-value, cleanup, NotFound, terminating/absent, or precedence path changed by this YAML-only patch.

Checks:
- `helm dependency build addons/valkey`: PASS
- `helm template valkey addons/valkey`: rendered data ComponentDefinition contains `accountProvision.exec.targetPodSelector: Role` and `matchingKey: primary`
- GitHub checks at review time: shell-check/check-addons/shellspec all SUCCESS; label-chart-release SKIPPED.

Before undraft/merge: current commit author/committer metadata uses a local machine email. Please amend it to the accepted public identity before marking the PR ready.

leon-ape (approved):
LGTM

PR #2866, opened 2026-06-20
fix(valkey): memberLeave exits non-zero when primary leaves with no sentinel

weicao:
## Summary
- When all sentinels are unreachable and the leaving pod is the primary, `memberLeave` now exits 1 instead of silently exiting 0.
- This lets KubeBlocks retry or timeout, giving sentinels a chance to recover before the primary pod is removed without failover.
- For non-primary pods, the behavior is unchanged (exit 0).

## Test plan

### DIFF
--- a/addons/valkey/scripts-ut-spec/valkey_member_leave_spec.sh
+++ b/addons/valkey/scripts-ut-spec/valkey_member_leave_spec.sh
@@ -230,17 +230,17 @@ Describe "Valkey Member-Leave Bash Script Tests"
     End
   End
 
-  Describe "no-sentinel primary-leave safety — exits non-zero"
+  Describe "no-sentinel fail-closed safety — exits non-zero unless confirmed replica"
     member_leave_script="../scripts/valkey-member-leave.sh"
 
-    It "has an exit 1 path for primary leaving with no sentinel"
-      When call grep -c "no reachable Sentinel and the leaving pod is the primary" "${member_leave_script}"
+    It "exits 0 only when confirmed replica (slave) and no sentinel reachable"
+      When call grep -c "leaving pod is a confirmed replica" "${member_leave_script}"
       The status should be success
       The stdout should not eq "0"
     End
 
-    It "only exits 0 for non-primary when no sentinel is reachable"
-      When call grep -c "leaving pod is not the primary" "${member_leave_script}"
+    It "exits 1 for master or unknown role when no sentinel reachable"
+      When call grep -c "cannot ensure safe failover" "${member_leave_script}"
       The status should be success
       The stdout should not eq "0"
     End
--- a/addons/valkey/scripts-ut-spec/valkey_member_leave_spec.sh
+++ b/addons/valkey/scripts-ut-spec/valkey_member_leave_spec.sh
@@ -230,19 +230,29 @@ Describe "Valkey Member-Leave Bash Script Tests"
     End
   End
 
-  Describe "no-sentinel fail-closed safety — exits non-zero unless confirmed replica"
-    member_leave_script="../scripts/valkey-member-leave.sh"
-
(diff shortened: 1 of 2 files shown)


weicao:
Focused review for W2:

I agree with the intended direction, but I would not undraft this yet.

Blocker: the no-reachable-Sentinel branch now fails only when `leaving_role` is exactly `master`. If role detection returns empty/unknown, the code still takes the non-primary path and exits 0. In the exact risk window this PR is trying to close, unknown role is not proof that the leaving pod is safe to remove. This is XP class 4: sentinel/empty value confusion, with class 1 silent-success risk.

Please either:
- fail closed when no Sentinel is reachable unless `leaving_role=slave` is explicitly proven; or
- add a concrete argument and unit test proving unknown role should remain success.

Test gap: targeted bash 5 spec currently passes because it does not cover the new branch. Please add ShellSpec cases for at least:
- no reachable Sentinel + leaving role master => exit non-zero
- no reachable Sentinel + leaving role slave => exit 0
- no reachable Sentinel + leaving role unknown/empty => chosen behavior, explicitly tested

Checks run:
- `shellspec --load-path ./shellspec --default-path addons/valkey/scripts-ut-spec/valkey_member_leave_spec.sh --shell /opt/homebrew/bin/bash`: 9 examples, 0 failures
- GitHub checks at review time: shellspec/check-addons/shell-check SUCCESS; label-chart-release SKIPPED.

Before undraft/merge: commit author/committer metadata uses a local machine email; please amend to the accepted public identity.

weicao:
Re-review of head `572ec384`:

The code-side blocker is fixed. The no-reachable-Sentinel branch now succeeds only for confirmed replica role (`leaving_role=slave`) and fail-closes for master/unknown. That resolves the class 4 unknown-role concern.

Remaining blocker before undraft: the new tests are static grep assertions. They prove the strings exist in the script, but they do not execute the no-Sentinel branch or prove the exit code contract. For this safety fix, please add behavior-level ShellSpec coverage, for example by extracting a small helper for the no-Sentinel decision or by running the script with mocked `valkey-cli` / `SENTINEL_POD_FQDN_LIST` inputs.

Required behavior cases remain:
- no reachable Sentinel + role `slave` => exit 0
- no reachable Sentinel + role `master` => exit non-zero
- no reachable Sentinel + empty/unknown role => exit non-zero

Local recheck:
- targeted bash5 memberLeave spec: 11 examples, 0 failures

CI caveat at re-review time: `pr-label-check / check` is FAILURE and shellspec/check-addons are still running. Commit author/committer metadata still uses a local machine email and should be amended before formal path.

weicao:
Re-review head `be52971e` after behavioral-test update: code/test blocker is closed from my side.

What I checked:
- Contract: no reachable Sentinel now succeeds only for confirmed replica (`role=slave`). `master`, `unknown`, and empty role all fail closed via `no_sentinel_safety_check()`.
- Tests: static grep assertions were replaced with executable behavior specs for `slave=>0`, `master=>1`, `unknown=>1`, `empty=>1`.
- Local verification: `bash -n addons/valkey/scripts/valkey-member-leave.sh` PASS; `git diff --check origin/main...HEAD` PASS; targeted ShellSpec with bash 5.3.12: `13 examples, 0 failures`.
- XP contract pass: class 1 silent fallback and class 4 sentinel/unknown-value confusion are now covered by code + behavior spec. Classes 2/3/5/6/7/8 have no new concern in this patch scope.

Remaining non-code caveats before undraft/merge:
- GitHub `shellspec-test` is still pending at my last check; wait for it to finish green.
- `pr-label-check / check` is failing.
- Commits still show local machine email (`wei@black-cat.local`); please amend to the accepted public identity before formalizing.

leon-ape (approved):
LGTM

PR #2867, opened 2026-06-20
fix(valkey): sync-acl exits non-zero on partial ACL failures

weicao:
## Summary
- Individual `ACL SETUSER` failures and `ACL SAVE` failures now cause `sync_acl_to_replica` to return non-zero.
- Previously, partial failures were logged as WARNING but the script exited 0, causing `memberJoin` to report success even though the replica's ACL state was incomplete.
- On pod restart, the replica would lose any in-memory-only ACL changes.

## Test plan

### DIFF
--- a/addons/valkey/scripts-ut-spec/valkey_sync_acl_spec.sh
+++ b/addons/valkey/scripts-ut-spec/valkey_sync_acl_spec.sh
@@ -161,6 +161,7 @@ Describe "Valkey Sync-ACL Bash Script Tests"
           "valkey-0.headless.default.svc.cluster.local" \
           "valkey-1.headless.default.svc.cluster.local"
         The status should be failure
+        The stdout should include "Syncing ACL"
         The stderr should include "ERROR: ACL SAVE failed"
       End
     End
@@ -187,6 +188,7 @@ Describe "Valkey Sync-ACL Bash Script Tests"
           "valkey-0.headless.default.svc.cluster.local" \
           "valkey-1.headless.default.svc.cluster.local"
         The status should be failure
+        The stdout should include "Syncing ACL"
         The stderr should include "ERROR: failed to set ACL for monitor"
         The stderr should include "ERROR: ACL sync completed with 1 failure(s)"
       End
--- a/addons/valkey/scripts-ut-spec/valkey_sync_acl_spec.sh
+++ b/addons/valkey/scripts-ut-spec/valkey_sync_acl_spec.sh
@@ -107,7 +107,7 @@ Describe "Valkey Sync-ACL Bash Script Tests"
     End
 
     Context "when ACL LIST fails"
-      It "returns failure and logs a warning"
+      It "returns failure and logs an error"
         valkey-cli() {
           case "$@" in
             *"ACL LIST"*)
@@ -120,7 +120,7 @@ Describe "Valkey Sync-ACL Bash Script Tests"
           "valkey-1.headless.default.svc.cluster.local"
         The status should be failure
         The stdout should include "Syncing ACL"
-        The stderr should include "WARNING"
+        The stderr should include "ERROR"
       End
     End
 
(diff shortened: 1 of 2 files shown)


weicao:
Focused review for W3:

The fail-closed direction is right, but this PR is currently blocked by tests.

Concrete blocker: the existing Valkey sync-acl ShellSpec fails under bash 5, and GitHub `shellspec-test` is also FAILURE at review time. The PR changed expected behavior from warning/success to error/non-zero, but did not update the specs.

Local targeted check:
`shellspec --load-path ./shellspec --default-path addons/valkey/scripts-ut-spec/valkey_sync_acl_spec.sh --shell /opt/homebrew/bin/bash`

Result: 6 examples, 2 failures:
- `ACL LIST fails returns failure and logs a warning` now emits `ERROR`, while the spec still expects `WARNING`.
- `ACL SAVE fails prints a warning but does not fail the sync` now exits non-zero and emits `ERROR`, while the spec still expects success/warning.

Please update/add specs for the new contract:
- ACL LIST command/server error => return non-zero and `ERROR`
- ACL SETUSER per-user partial failure => return non-zero after attempting the batch
- ACL SAVE failure => return non-zero and `ERROR`
- success path still exits 0 and prints `ACL sync complete`

XP design-contract note: this is a class 1 fix direction, removing silent success on partial ACL failure. It just needs tests to lock the new failure contract before undraft.

Also before undraft/merge: commit author/committer metadata uses a local machine email; please amend to the accepted public identity.

weicao:
Re-review of head `25474c1f`:

The spec direction is fixed, but the targeted bash5 ShellSpec still exits non-zero due ShellSpec warnings.

Local command:
`git diff --check origin/main...HEAD && shellspec --load-path ./shellspec --default-path addons/valkey/scripts-ut-spec/valkey_sync_acl_spec.sh --shell /opt/homebrew/bin/bash`

Result:
- `git diff --check`: PASS
- ShellSpec: 7 examples, 0 failures, 2 warnings, process exit 101

Warnings:
- ACL SAVE failure case emits stdout `Syncing ACL ...` but the example has no stdout expectation.
- partial ACL SETUSER failure case emits stdout for `Syncing ACL ...` and `ACL SETUSER ...` lines but the example has no stdout expectation.

Please add explicit stdout expectations or otherwise suppress/expect that output so the targeted spec exits 0. After that, the code contract looks aligned: ACL LIST error, ACL SETUSER partial failure, and ACL SAVE failure all return non-zero.

CI caveat at re-review time: `pr-label-check / check` is FAILURE and GitHub shellspec is still running. Commit author/committer metadata still uses a local machine email and should be amended before formal path.

weicao:
Re-review head `ac8e7c55` after stdout expectation update: code/test blocker is closed from my side.

What I checked:
- Contract: `ACL LIST` errors, partial `ACL SETUSER` errors, and `ACL SAVE` errors now return non-zero from `sync_acl_to_replica()`, so the memberJoin action no longer silently reports success for incomplete ACL sync.
- Tests: failure-path specs now assert status, stderr, and stdout, so ShellSpec exits clean instead of warning-only exit 101.
- Local verification: `bash -n addons/valkey/scripts/sync-acl.sh` PASS; `git diff --check origin/main...HEAD` PASS; targeted ShellSpec with bash 5.3.12: `7 examples, 0 failures`.
- GitHub checks at re-review time: shellspec-test, shell-check, and chart checks are SUCCESS. `pr-label-check / check` is still FAILURE.
- XP contract pass: class 1 silent-success fallback is closed for ACL LIST / SETUSER / SAVE errors. Class 3 commit-state discontinuity is still intentionally handled as retryable memberJoin failure, not silent success. Classes 2/4/5/6/7/8 have no new concern in this patch scope.

Remaining non-code caveats before undraft/merge:
- Fix the failing label check path.
- Commits still show local machine email (`wei@black-cat.local`); please amend to the accepted public identity before formalizing.

leon-ape (approved):
LGTM

PR #2869, opened 2026-06-20
chore(valkey): add io-threads to staticParameters in ParametersDefinition

weicao:
## Summary
- `io-threads` and `io-threads-do-reads` require a restart to take effect but were missing from `staticParameters`.
- The config template renders them, so they should be explicitly classified as static.
- Currently works by accident (unlisted params default to static when `dynamicParameters` is non-empty), but this is fragile.

## Test plan

### DIFF
--- a/addons/valkey/templates/paramsdef.yaml
+++ b/addons/valkey/templates/paramsdef.yaml
@@ -24,6 +24,8 @@ spec:
     - logfile
     - aclfile
     - tcp-backlog
+    - io-threads
+    - io-threads-do-reads
   dynamicParameters:
     - acllog-max-len
     - activedefrag


weicao:
Focused review for W8:

No code blocker found for classifying `io-threads` and `io-threads-do-reads` as static. The config template renders both keys in `valkey.conf`, and `docs/addon-api/08-parameters-and-config.md` says `staticParameters` is for parameters that require restart after change. This patch makes that classification explicit instead of relying on default/static fallback behavior.

Checks:
- `helm dependency build addons/valkey`: PASS
- `helm template valkey addons/valkey`: rendered ParametersDefinition includes `io-threads` and `io-threads-do-reads` under `staticParameters` and not under `dynamicParameters`.
- GitHub checks at review time: chart/shell-check SUCCESS; shellspec still IN_PROGRESS; label-chart-release SKIPPED.

XP design-contract walk: class 1 improves explicitness; class 2 writes the restart-required contract at the policy site; classes 3-8 are not touched by this ParametersDefinition-only patch.

Before undraft/merge: wait for CI/IDC acceptance, and amend commit author/committer metadata from the local machine email to the accepted public identity.

leon-ape (approved):
LGTM

PR #2872, opened 2026-06-21
fix(valkey): push ACL to all data pods in accountProvision

weicao:
## Summary
- Valkey ACL rules are not replicated via native replication — each node has its own users.acl
- account-provision.sh previously only ran ACL SETUSER + ACL SAVE on the local primary pod
- After applying locally, discovers all connected replicas via INFO REPLICATION and pushes the same ACL to each
- Uses COMPONENT_REPLICAS to validate expected replica count; bounded retry (30s) waits for enough online replicas
- Fail-closed: INFO REPLICATION failure, insufficient online replicas, or any replica ACL SAVE failure causes exit non-zero
- Standalone (COMPONENT_REPLICAS=1) exits immediately after local apply

INFO REPLICATION replaces VALKEY_POD_FQDN_LIST enumeration because the FQDN list is fixed at pod creation time and may be stale after scale-out — new pods would not appear in the primary's list.

Companion to #2865 (targetPodSelector). Together they ensure: accountProvision runs on primary (#2865), and the ACL propagates to all existing replicas (this PR).

**Merge order**: merge #2865 first, then rebase this PR to drop the shared commit before merging.

## Test plan
- T08-A: static CMPD assertion for targetPodSelector (guards #2865)
- T08-B: create account via account-provision.sh on primary, verify ACL LIST + --user auth on all data pods (guards this PR)
- kubeblocks-tests PR #270 contains both assertions

### DIFF
--- a/addons/valkey/scripts/account-provision.sh
+++ b/addons/valkey/scripts/account-provision.sh
@@ -75,14 +75,14 @@ while IFS= read -r line; do
       if [ "${state}" = "online" ]; then
         online_ips+=("${ip}")
       else
-        echo "WARNING: replica ${ip} state=${state}, skipping ACL push" >&2
+        echo "WARNING: replica ${ip} state=${state}, not online" >&2
       fi
       ;;
   esac
 done <<< "${repl_info}"
 
-if [ "${slave_count}" -gt 0 ] && [ "${#online_ips[@]}" -eq 0 ]; then
-  echo "ERROR: ${slave_count} replica(s) connected but none in online state — cannot propagate ACL" >&2
+if [ "${slave_count}" -gt 0 ] && [ "${#online_ips[@]}" -ne "${slave_count}" ]; then
+  echo "ERROR: ${slave_count} replica(s) connected but only ${#online_ips[@]} online — cannot guarantee all replicas receive ACL" >&2
   exit 1
 fi
 
--- a/addons/valkey/scripts/account-provision.sh
+++ b/addons/valkey/scripts/account-provision.sh
@@ -65,17 +65,26 @@ repl_info=$("${_cli[@]}" INFO REPLICATION 2>&1) || {
 repl_info="${repl_info//$'\r'/}"
 
 online_ips=()
+slave_count=0
 while IFS= read -r line; do
   case "${line}" in
     slave[0-9]*:*)
+      slave_count=$((slave_count + 1))
       ip=$(echo "${line}" | sed 's/.*ip=\([^,]*\).*/\1/')
       state=$(echo "${line}" | sed 's/.*state=\([^,]*\).*/\1/')
       if [ "${state}" = "online" ]; then
         online_ips+=("${ip}")
+      else
+        echo "WARNING: replica ${ip} state=${state}, skipping ACL push" >&2
       fi
       ;;
   esac
 done <<< "${repl_info}"
+
+if [ "${slave_count}" -gt 0 ] && [ "${#online_ips[@]}" -eq 0 ]; then
+  echo "ERROR: ${slave_count} replica(s) connected but none in online state — cannot propagate ACL" >&2
+  exit 1
+fi
 
 for ip in "${online_ips[@]}"; do
   apply_acl "${ip}" "replica(${ip})"
--- a/addons/valkey/scripts/account-provision.sh
+++ b/addons/valkey/scripts/account-provision.sh
@@ -55,51 +55,27 @@ apply_acl() {
 # Apply on the local primary first.
 apply_acl "127.0.0.1" "primary(local)"
 
-# For multi-node deployments, discover connected replicas via INFO
-# REPLICATION and push the ACL to each.  Fail-closed: if we cannot confirm
-# that every expected replica received the ACL, the action fails so
-# KubeBlocks can surface the inconsistency.
-expected_replicas=$(( ${COMPONENT_REPLICAS:-1} - 1 ))
-if [ "${expected_replicas}" -le 0 ]; then
-  exit 0
-fi
-
-deadline=$((SECONDS + 30))
-online_ips=()
-
-while [ $SECONDS -lt $deadline ]; do
-  build_cli "127.0.0.1"
-  repl_info=$("${_cli[@]}" INFO REPLICATION 2>&1) || {
-    echo "INFO: INFO REPLICATION failed, retrying..." >&2
-    sleep 3
-    continue
-  }
-  repl_info="${repl_info//$'\r'/}"
-
-  online_ips=()
-  while IFS= read -r line; do
-    case "${line}" in
-      slave[0-9]*:*)
-        ip=$(echo "${line}" | sed 's/.*ip=\([^,]*\).*/\1/')
-        state=$(echo "${line}" | sed 's/.*state=\([^,]*\).*/\1/')
-        if [ "${state}" = "online" ]; then
-          online_ips+=("${ip}")
-        fi
-        ;;
-    esac
-  done <<< "${repl_info}"
-
-  if [ ${#online_ips[@]} -ge ${expected_replicas} ]; then
-    break
-  fi
-  echo "INFO: waiting for replicas (online=${#online_ips[@]} expected=${expected_replicas})..." >&2
-  sleep 3
-done
-
-if [ ${#online_ips[@]} -lt ${expected_replicas} ]; then
-  echo "ERROR: not enough online replicas after 30s (online=${#online_ips[@]} expected=${expected_replicas})" >&2
+# Discover connected replicas via INFO REPLICATION and push the ACL to each.
+# Uses only real-time data from the engine — no stale env vars.
+build_cli "127.0.0.1"
+repl_info=$("${_cli[@]}" INFO REPLICATION 2>&1) || {
+  echo "ERROR: INFO REPLICATION failed" >&2
   exit 1
-fi
+}
+repl_info="${repl_info//$'\r'/}"
+
+online_ips=()
+while IFS= read -r line; do
+  case "${line}" in
+    slave[0-9]*:*)
+      ip=$(echo "${line}" | sed 's/.*ip=\([^,]*\).*/\1/')
+      state=$(echo "${line}" | sed 's/.*state=\([^,]*\).*/\1/')
+      if [ "${state}" = "online" ]; then
+        online_ips+=("${ip}")
+      fi
+      ;;
+  esac
+done <<< "${repl_info}"
 
 for ip in "${online_ips[@]}"; do
   apply_acl "${ip}" "replica(${ip})"
--- a/addons/valkey/scripts/account-provision.sh
+++ b/addons/valkey/scripts/account-provision.sh
@@ -2,53 +2,105 @@
 set -e
 # account-provision.sh — called by KubeBlocks to create/update a database account.
 #
-# Learning note:
-#   KubeBlocks injects three special variables before calling this action:
-#     KB_ACCOUNT_NAME      - logical account name (e.g., "default")
-#     KB_ACCOUNT_PASSWORD  - generated password (from passwordGenerationPolicy)
-#     KB_ACCOUNT_STATEMENT - the command to execute, built by KubeBlocks
+# KubeBlocks injects:
+#   KB_ACCOUNT_NAME      - logical account name (e.g., "default")
+#   KB_ACCOUNT_PASSWORD  - generated password (from passwordGenerationPolicy)
+#   KB_ACCOUNT_STATEMENT - the command to execute (ACL SETUSER ...)
 #
-#   For Redis/Valkey, KB_ACCOUNT_STATEMENT is an ACL SETUSER command:
-#     e.g. "ACL SETUSER myuser on >password ~* +@all"
-#
-#   Because Valkey's auth model is ACL-based (not SQL), the provision script
-#   simply passes KB_ACCOUNT_STATEMENT verbatim to valkey-cli and then
-#   calls ACL SAVE to persist the ACL file to disk.
-#
-#   The `initAccount: true` flag on the systemAccount means KubeBlocks
-#   calls this action once during cluster initialisation, using the
-#   "default" user.  Subsequent ACL changes go through OpsRequest.
+# With targetPodSelector: Role / matchingKey: primary, this runs on the
+# primary pod.  After applying the ACL locally, the script pushes the same
+# ACL rule to all connected replicas (discovered via INFO REPLICATION).
+# Valkey ACL is not replicated via the native replication stream — each
+# node maintains its own users.acl.
 
 port="${SERVICE_PORT:-6379}"
 
-base_cmd=(valkey-cli --no-auth-warning -p "${port}")
-if [ -n "${VALKEY_DEFAULT_PASSWORD}" ]; then
-  base_cmd+=(-a "${VALKEY_DEFAULT_PASSWORD}")
-fi
-if [ -n "${VALKEY_CLI_TLS_ARGS}" ]; then
-  # shellcheck disable=SC2206
-  base_cmd+=(${VALKEY_CLI_TLS_ARGS})
+build_cli() {
+  local host="${1:-127.0.0.1}"
+  _cli=(valkey-cli --no-auth-warning -h "${host}" -p "${port}")
+  if [ -n "${VALKEY_DEFAULT_PASSWORD}" ]; then
+    _cli+=(-a "${VALKEY_DEFAULT_PASSWORD}")
+  fi
+  if [ -n "${VALKEY_CLI_TLS_ARGS}" ]; then
+    # shellcheck disable=SC2206
+    _cli+=(${VALKEY_CLI_TLS_ARGS})
+  fi
+}
+
+apply_acl() {
+  local host="${1}" label="${2}"
+  build_cli "${host}"
+
+  local output
+  output=$(echo "${KB_ACCOUNT_STATEMENT}" | "${_cli[@]}" 2>&1) || {
+    echo "ERROR: account statement failed on ${label} (${host}): ${output}" >&2
+    return 1
+  }
+  if [ "${output}" != "OK" ]; then
+    echo "ERROR: account statement returned unexpected response on ${label} (${host}): ${output}" >&2
+    return 1
+  fi
+
+  local acl_save_out
+  acl_save_out=$("${_cli[@]}" ACL SAVE 2>&1) || {
+    echo "ERROR: ACL SAVE failed on ${label} (${host}): ${acl_save_out}" >&2
+    return 1
+  }
+  if [ "${acl_save_out}" != "OK" ]; then
+    echo "ERROR: ACL SAVE returned unexpected response on ${label} (${host}): ${acl_save_out}" >&2
+    return 1
+  fi
+}
+
+# Apply on the local primary first.
+apply_acl "127.0.0.1" "primary(local)"
+
+# For multi-node deployments, discover connected replicas via INFO
+# REPLICATION and push the ACL to each.  Fail-closed: if we cannot confirm
+# that every expected replica received the ACL, the action fails so
+# KubeBlocks can surface the inconsistency.
+expected_replicas=$(( ${COMPONENT_REPLICAS:-1} - 1 ))
+if [ "${expected_replicas}" -le 0 ]; then
+  exit 0
 fi
 
-# Execute the statement provided by KubeBlocks.
-# Pipe via stdin so that '>' in ACL password syntax (e.g. >mypassword) is not
-# misinterpreted as a shell output-redirect operator.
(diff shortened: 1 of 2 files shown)


weicao:
Focused review result for PR #2872 (`45eac560`): blocker before undraft/merge.

What is good:
- The patch is fail-closed for each host it actually visits: both `ACL SETUSER` output and `ACL SAVE` output must be `OK`.
- Local fake `valkey-cli` probe confirmed it applies local first, skips the current pod from the FQDN list, applies to the other listed pods, and exits non-zero when a replica `ACL SAVE` returns a non-OK response.
- Local checks passed: `bash -n addons/valkey/scripts/account-provision.sh`, `git diff --check origin/main...HEAD`, and Valkey ShellSpec with bash 5: `287 examples, 0 failures`.

Blocker 1 — current membership source can be stale after scale-out.

The PR summary says this pushes ACL to every other data pod, but the implementation iterates `VALKEY_POD_FQDN_LIST` from the running primary pod. Existing Valkey script code already documents the issue in `switchover.sh`: `VALKEY_POD_FQDN_LIST is rendered into pod environment at pod creation time. After scale-out, old primary pods can still have a stale list...`.

That means the fix is correct only for pods present in the primary's current env snapshot. If the cluster scales out, memberJoin can copy existing ACLs to the new pod, but a later accountProvision executed on the old primary can still miss that newer pod because the old primary's `VALKEY_POD_FQDN_LIST` does not include it. Under the W1 contract we accepted in the Slock thread — "new account must be available on all existing data pods" — this is still a design-contract gap.

Please either:
- use a current membership source for accountProvision, not the pod-creation-time FQDN env snapshot; or
- explicitly narrow the contract and test scope to "pods present in the primary env snapshot", which I do not recommend for W1 acceptance.

Design-contract classification: Class 3, treating a one-time state snapshot as continuous current state.

Blocker 2 — commit metadata must be amended before undraft/merge.

Commit `45eac560` has author/committer `Wei <wei@black-cat.local>`. Please amend to the accepted human identity before undrafting/merging. The `Co-authored-by: Alice <alice@apecloud.com>` trailer is fine.

Additional note:
- PR #2872 is stacked on #2865 (`ad363dbb`) while targeting `main`. That is fine if the intended acceptance unit is "#2865 + #2872 together", but the merge/review wording should keep that explicit.
- The broad full ShellSpec suite gave `1122 examples, 0 failures, 1 warning`; the warning is unrelated MongoDB dataprotection regex stderr, so I used the Valkey-only green result as the PR-relevant signal.

weicao:
Re-review for head `becf2958`: the stale `VALKEY_POD_FQDN_LIST` blocker is addressed in direction, but there is a new fail-closed blocker before undraft/merge.

What improved:
- Moving replica discovery from pod-creation-time `VALKEY_POD_FQDN_LIST` to live `INFO REPLICATION` is the right direction for the scale-out-then-accountProvision counterexample.
- Local fake `valkey-cli` probe confirms the happy path applies to online replicas reported by `INFO REPLICATION`.
- Static checks still pass: `bash -n addons/valkey/scripts/account-provision.sh`; `git diff --check origin/main...HEAD`.

Blocker — `INFO REPLICATION` discovery failure / incomplete topology still exits 0 after updating only the primary.

Current code:

```bash
repl_info=$("${_cli[@]}" INFO REPLICATION 2>&1) || repl_info=""
...
if [ "${state}" = "online" ]; then
  apply_acl "${ip}" "replica(${ip})"
else
  echo "WARNING: skipping replica ${ip} (state=${state})" >&2
fi
```

This means these cases all succeed from KB's point of view even though W1/all-current-data-pod propagation was not proven:
- `INFO REPLICATION` command fails: script exits 0 after local primary only.
- `INFO REPLICATION` returns no `slaveN:` rows in a multi-replica component: script exits 0 after local primary only.
- a replica row exists but `state != online`: script logs WARNING and exits 0.

I reproduced those paths with a fake `valkey-cli`: modes `fail`, `none`, and `loading` all returned `rc=0`; `loading` only printed `WARNING: skipping replica 10.0.0.4 (state=wait_bgsave)`.

Under the accepted W1 contract, accountProvision must not report success unless it either:
1. applies ACL + ACL SAVE to every expected current data pod; or
2. has a formally narrowed contract, which we have rejected for W1.

Recommended fix:
- Do not swallow `INFO REPLICATION` failure. Treat it as `ERROR` + non-zero.
- Use `COMPONENT_REPLICAS` to compute expected replica count (`COMPONENT_REPLICAS - 1`) when present, and require exactly that many online replicas before returning success. If the component is single-node, expected replica count is 0 and local-only success is valid.
- For replica rows with `state != online`, either bounded-retry until online or fail closed. Warning-only success keeps the original false-success class.
- Keep per-replica `ACL SETUSER` and `ACL SAVE` fail-closed as this PR already does.

Design-contract classification: Class 1 silent fallback / warning-only success, plus Class 2 non-empty expected-replica contract not enforced at the write site.

Separate pre-undraft item still open:
- Commits `45eac560` and `becf2958` still have author/committer `Wei <wei@black-cat.local>`. Please amend to the accepted human identity before undraft/merge.

Also update the PR body after the next push: it still says the implementation iterates `VALKEY_POD_FQDN_LIST`, which is now outdated.

weicao:
Re-reviewed head `d4619aec17e558a4d24cc229b32d0a3fa4b20b6f`.

Code-design blocker from the previous review is closed for this head:

- `COMPONENT_REPLICAS` is used to compute the expected replica count.
- Standalone (`COMPONENT_REPLICAS=1`) exits after local primary apply.
- Multi-replica mode now bounded-retries `INFO REPLICATION` and requires enough `state=online` replicas before continuing.
- Replica `ACL SETUSER` and `ACL SAVE` remain fail-closed.

Local checks:

- `bash -n addons/valkey/scripts/account-provision.sh` PASS
- `git diff --check origin/main...HEAD` PASS
- fake `valkey-cli` matrix PASS for the review cases:
  - 2 expected replicas online -> rc=0, local + both replicas updated
  - `INFO REPLICATION` failure -> retry then rc=1
  - no slave rows -> retry then rc=1
  - non-online replica -> retry then rc=1
  - replica `ACL SAVE` failure -> rc=1
- GitHub checks are green at the time of review, including `shellspec-test` and `shell-check`.

XP design-contract walk:

- Class 1 silent fallback: previous partial-success path is closed for the multi-replica path because insufficient online replicas now exits nonzero. `COMPONENT_REPLICAS` itself is defined as required in the Valkey CMPD env and existing Valkey scripts also treat it as required.
- Class 2 non-empty contract write: `targetPodSelector: Role` / `matchingKey: primary` is present in CMPD; account ACL writes check `OK` for both `ACL SETUSER` and `ACL SAVE`.
- Class 3 stale state: previous `VALKEY_POD_FQDN_LIST` snapshot issue is removed from the runtime discovery path; current connected replicas are discovered from `INFO REPLICATION` at action time.
- Class 4 sentinel values: no empty-list success in multi-replica mode; empty/failed discovery becomes timeout + rc=1.
- Class 5 conditional cleanup: not applicable in this script path.
- Class 7 terminating-vs-absent: not applicable.
- Class 8 precedence: no complex boolean chain introduced.

Remaining before undraft/merge:

1. Update PR body. It still says the implementation iterates `VALKEY_POD_FQDN_LIST`; current code uses `INFO REPLICATION` + `COMPONENT_REPLICAS`.
2. Amend/squash metadata before undraft: commits `45eac560`, `becf2958`, and `d4619aec` still have local author/committer identity `Wei <wei@black-cat.local>`, and the older commit messages still describe the stale FQDN-list implementation.
3. Runtime acceptance is still required: #2865 + #2872 should pass PR #270 T08-A/T08-B, especially the all-data-pod `ACL LIST` + `--user` auth check, before W1 is called accepted.

addons/valkey/scripts/account-provision.sh
```diff
@@ -2,53 +2,105 @@
...
-  exit 1
+
+# Apply on the local primary first.
+apply_acl "127.0.0.1" "primary(local)"
+
+# For multi-node deployments, discover connected replicas via INFO
+# REPLICATION and push the ACL to each.  Fail-closed: if we cannot confirm
+# that every expected replica received the ACL, the action fails so
+# KubeBlocks can surface the inconsistency.
+expected_replicas=$(( ${COMPONENT_REPLICAS:-1} - 1 ))
```
leon-ape: COMPONENT_REPLICAS comes from component vars on the existing primary and can be stale after scale-out. If the primary was created at replicas=1, later accountProvision computes expected_replicas=0 and exits after updating only the primary, leaving the new replicas without the account.

weicao: Good catch. COMPONENT_REPLICAS is injected as a pod env var at creation time, so after scale-out the primary's value stays stale (e.g. still 1 even when actual replicas=3). This would make expected_replicas=0 and skip all replica ACL push.

Fix approach: instead of using COMPONENT_REPLICAS to compute expected count, push ACL to ALL replicas discovered via `INFO REPLICATION` connected_slaves. No count gate — just discover and push.

Note: we're discussing the overall account-provision design direction with the team (whether to keep this primary-push approach vs targetPodSelector:All vs Redis-style OpsDefinition). Will address this as part of that resolution.

leon-ape: Current head `35600e74` still computes `expected_replicas` from `COMPONENT_REPLICAS` and exits 0 when it is `<= 0`. Since that env value can be stale after scale-out, accountProvision can still succeed after updating only the primary.

weicao:
Fixed the `COMPONENT_REPLICAS` staleness issue Leon identified.

**Change**: Removed the `COMPONENT_REPLICAS`-gated polling loop. Now the script does a single `INFO REPLICATION` call and pushes ACL to all currently online replicas — no stale env var dependency.

**Result**: 107 → 82 lines. The script works correctly after scale-out because `INFO REPLICATION` always reflects the real-time connected replica list.

Commit: d9b4d87e

weicao:
**Bob focused review of d9b4d87e — PASS, 0 blocker.**

Checked:
1. **COMPONENT_REPLICAS fully removed** — no stale env var dependency left
2. **INFO REPLICATION failure → exit 1** — fail-closed, no silent skip
3. **0 online replicas → for loop naturally skips** — single-node scenario correct
4. **Per-replica apply_acl failure → script exits non-zero** — set -e propagates return 1
5. **apply_acl refactoring**: clean, per-host error messages include label for diagnosis
6. **bash -n static check**: PASS

Minor observations (not blockers):
- sed spawns 2 subprocesses per replica line — acceptable here since this is not a probe script (no SIGKILL/zombie risk)
- Uses global port for all replicas (same SERVICE_PORT assumption) — valid for standard addon deployment

107 → 82 lines, simpler and correct. Ready for Emily runtime validation.

addons/valkey/scripts/account-provision.sh
```diff
@@ -2,53 +2,81 @@
...
+      ip=$(echo "${line}" | sed 's/.*ip=\([^,]*\).*/\1/')
+      state=$(echo "${line}" | sed 's/.*state=\([^,]*\).*/\1/')
+      if [ "${state}" = "online" ]; then
+        online_ips+=("${ip}")
+      fi
+      ;;
+  esac
+done <<< "${repl_info}"
+
+for ip in "${online_ips[@]}"; do
```
leon-ape: This still exits 0 when INFO REPLICATION returns no slaveN rows or no online replicas, after updating only the primary. A multi-node accountProvision can still be reported successful without proving that any replica received the ACL.

weicao:
Fixed in a4ed313b. Changes:

- Track `slave_count` (all slave rows from INFO REPLICATION, any state)
- If `slave_count > 0` but no replica has `state=online`: exit 1 with error
- Non-online replicas now log a WARNING with their state before being skipped
- `bash -n` static check PASS

Behavior per scenario:
- **Standalone (0 slaves)**: exits 0 after primary — correct
- **Multi-replica, all online**: pushes to all replicas — correct
- **Multi-replica, some online**: pushes to online ones, warns about non-online — correct (non-online replicas will get ACL from sync-acl on memberJoin when they reconnect)
- **Multi-replica, none online**: **exits 1** — no longer falsely reports success

weicao:
Focused review on head `a4ed313b4826c224ad78137dd2339ed8b8abb4b4`: blocker remains.

The new guard closes the exact `slave_count > 0 && online_ips == 0` false-success case, but the mixed case still exits 0:

- `INFO REPLICATION` has `slave0 state=online` and `slave1 state=wait_bgsave`
- script applies ACL to primary and the online replica
- script logs a warning for the non-online replica
- script returns 0

That still lets accountProvision report success while one currently reported replica did not receive the ACL. For this contract, zero exit should mean every replica reported by `INFO REPLICATION` either received the ACL, or there were no replica rows. Otherwise a scale-out window can leave the new replica without the account after it later becomes online.

Suggested fix: fail/retry when `slave_count != ${#online_ips[@]}` (or bounded-wait until all reported replicas are online, then push), instead of only failing when `${#online_ips[@]} == 0`. Please also add an executable shell/spec case for the mixed online + non-online replica path; the current commit only changes `account-provision.sh`, and the existing local ShellSpec suite does not cover this script.

Checks I ran locally:
- `bash -n addons/valkey/scripts/account-provision.sh` PASS
- `git diff --check origin/main...HEAD -- addons/valkey/scripts/account-provision.sh` PASS
- local mock: no slaves => rc 0; one online replica => rc 0; one connected non-online replica => rc 1; mixed online + non-online replicas => rc 0, which is the remaining false-success path
- `shellspec --load-path ./shellspec --default-path "addons/valkey/scripts-ut-spec" --shell bash` on macOS bash 3.2: 40 examples / 0 failures, but no account-provision coverage observed

weicao:
Follow-up fix in fc84c7d7: tightened the guard per Bob cross-review.

**Previous** (a4ed313b): fail only when `slave_count > 0 && online_count == 0`
**Now** (fc84c7d7): fail when `slave_count > 0 && online_count != slave_count`

This closes the mixed scenario Bob identified: if INFO REPLICATION reports 2 slaves but only 1 is online (the other is in wait_bgsave), the script now exits 1 instead of reporting success after pushing ACL to only the online replica.

KB will retry the lifecycle action. By the retry, transitioning replicas should be online.

Behavior matrix:
- **0 slaves (standalone)**: exit 0 after primary — correct
- **2 slaves, 2 online**: push to all, exit 0 — correct
- **2 slaves, 1 online + 1 wait_bgsave**: **exit 1** — correct, was false-success before
- **2 slaves, 0 online**: **exit 1** — correct
- **INFO REPLICATION fails**: **exit 1** — correct (already handled)

`bash -n` PASS.

weicao:
Focused re-review on head `fc84c7d7c8006ae1d930d482000b1b2bfdb9cc6c`: PASS / 0 code-contract blocker for the false-success path I flagged.

I re-ran the same executable local mock matrix against `account-provision.sh`:

- no slave rows => rc 0
- one online replica => rc 0
- one connected non-online replica (`wait_bgsave`) => rc 1
- mixed replicas (`1 online + 1 wait_bgsave`) => rc 1
- two online replicas => rc 0

The mixed case now fails with:

```text
ERROR: 2 replica(s) connected but only 1 online — cannot guarantee all replicas receive ACL
```

Design-contract pass:
- Class 1 silent fallback: closed for reported-but-not-online replicas; no partial success on mixed state.
- Class 2 non-empty/complete success contract: zero exit now means every replica row reported by `INFO REPLICATION` was online before replica ACL push proceeds.
- Class 3 state boundary: the script still uses a single `INFO REPLICATION` snapshot, but failure lets KB retry rather than silently accepting a transitioning state.
- Class 4 sentinel values: `slave_count` vs `online_ips` is explicit; no `online_count == 0` special-case hole remains.
- Classes 5-8: no cleanup/NotFound/terminating/precedence concern in this patch.

Checks run locally:
- `bash -n addons/valkey/scripts/account-provision.sh` PASS
- `git diff --check origin/main...HEAD -- addons/valkey/scripts/account-provision.sh` PASS
- `shellspec --load-path ./shellspec --default-path "addons/valkey/scripts-ut-spec" --shell bash` on macOS bash 3.2: 40 examples / 0 failures

Caveat: I still do not see checked-in ShellSpec coverage for `account-provision.sh`; the local mock above proves the branch for this review, but the repository suite will not catch a future regression unless a spec is added.

leon-ape:
Finding on current head fc84c7d7: accountProvision can still report success for a multi-pod Valkey component when the primary INFO REPLICATION snapshot contains no slaveN rows. After the COMPONENT_REPLICAS guard was removed, slave_count=0 is treated as standalone, so the script exits 0 after updating only the primary. If existing replicas are temporarily disconnected or otherwise absent from that single primary snapshot, KubeBlocks can mark the account provisioned while those data pods did not receive the ACL.

weicao:
Follow-up after Leon's latest comment: my `fc84c7d7` PASS was scoped to the mixed-state false-success path (`1 online + 1 non-online`). Leon's new case is different and still needs owner handling.

Current code behavior is still:

- `INFO REPLICATION` has no `slaveN:` rows => `slave_count=0`, `online_count=0`, script exits 0 after applying ACL only on the local primary

That is correct for true standalone, but unsafe if the component is expected to have replicas and the primary's single `INFO REPLICATION` snapshot temporarily reports no slaves. In that case accountProvision can still be marked successful while replicas did not receive the ACL.

So I agree this is a remaining #2872 blocker unless the PR adds a reliable way to distinguish true standalone from "replicas expected but absent from this snapshot", or otherwise fails/retries the ambiguous multi-pod case. The earlier local mock matrix should be extended with this no-slave-row + expected-replicas case once the script has a stable expected-topology source.

weicao:
Design pivot in e35696e9: switched to `targetPodSelector: All` (same approach as sentinel B3 fix in PR #2932).

**Why**: Cross-pod fan-out from the primary is inherently fragile for Valkey ACL because:
1. `COMPONENT_REPLICAS` is stale after scale-out (Leon, first review)
2. `INFO REPLICATION` is a point-in-time snapshot — 0 slave rows cannot distinguish standalone from all-replicas-disconnected (Leon, latest review)
3. Mixed online/non-online states require complex retry logic

**What changed**:
- `cmpd.yaml`: `targetPodSelector: Role / matchingKey: primary` -> `targetPodSelector: All`
- `account-provision.sh`: Simplified to local-only ACL apply (92 -> 43 lines). Each pod applies `ACL SETUSER + ACL SAVE` on itself. No cross-pod discovery, no INFO REPLICATION, no stale env vars.

**Why this works**: KB runs the script on every pod in the component. Since Valkey ACL is per-node (not replicated), each pod must receive the ACL independently. `targetPodSelector: All` is the Occam's razor solution — it eliminates the entire class of false-success bugs.

`bash -n` PASS.

weicao:
Focused review on head `e35696e9d09618e37dba68afa011720b6421b1c9`: PASS / 0 code-contract blocker for the design switch to `targetPodSelector: All`.

What I checked:

- PR diff is limited to:
  - `addons/valkey/templates/cmpd.yaml`
  - `addons/valkey/scripts/account-provision.sh`
- `accountProvision.exec.targetPodSelector: All` is set for the Valkey data component and `matchingKey` is removed.
- The data `account-provision.sh` is local-only: 43 rendered lines, no `INFO REPLICATION`, no `COMPONENT_REPLICAS`, no `slave_count`, no `online_ips`, no cross-pod fan-out.
- Synthetic merge onto current `origin/main` applied cleanly; rendered result keeps both data and Sentinel accountProvision on `targetPodSelector: All`:
  - `valkey-8`: `container=valkey`, `targetPodSelector=All`
  - `valkey-9`: `container=valkey`, `targetPodSelector=All`
  - `valkey-sentinel-8/9`: still `targetPodSelector=All` from #2932

Contract review:

- Class 1 silent fallback: closed. The script no longer treats incomplete `INFO REPLICATION` snapshots as success because it no longer uses snapshots.
- Class 2 non-empty/complete-success contract: success is now per-pod local `ACL SETUSER + ACL SAVE`; KB fan-out owns all-pod execution.
- Class 3 state boundary: removes point-in-time primary replication state from the contract.
- Class 4 sentinel values: removes `slave_count=0` ambiguity between standalone and disconnected replicas.
- Classes 5-8: no cleanup / NotFound / terminating / precedence issue in this patch.

Checks run locally:

- `bash -n addons/valkey/scripts/account-provision.sh` PASS
- `git diff --check` PASS on the synthetic merge tree
- `helm template valkey addons/valkey --namespace kb-system` PASS
- `shellspec --load-path ./shellspec --default-path "addons/valkey/scripts-ut-spec" --shell bash` on macOS bash 3.2: 40 examples / 0 failures
- local command mock for the simplified script: success => rc 0; statement command failure/unexpected output => rc 1; `ACL SAVE` failure/unexpected output => rc 1

Boundary: this is a code/render contract PASS. Runtime closeout should still validate a fresh accountProvision on a 3-data-pod cluster and prove the generated user exists/authenticates on all 3 data pods after KB runs the action with `targetPodSelector: All`.

leon-ape (approved):
LGTM

PR #2876, opened 2026-06-21
fix(tidb): PD roleProbe peer fallback + remove externalManaged

weicao:
## Summary

- **PD roleProbe chicken-and-egg fix**: When PD-0 local API is unresponsive during initial bootstrap (stuck in region sync), the roleProbe now falls back to querying peer endpoints via PD_POD_FQDN_LIST. This lets PD-0 get its follower role even when its own server has not fully started, unblocking KB controller component Running judgment and allowing TiKV creation to proceed.
- **Remove externalManaged: true** from PD, TiKV, and TiDB ComponentDefinitions. Without an external provisioner, the controller clears the config template and precheck fails.

## Root cause evidence

Kate IDC vcluster testing found:
1. 3 PD pods all Running (2/2), InstanceSet 3/3 Ready
2. PD-1=leader, PD-2=follower, PD-0=no role (probe timeout)
3. KB controller: status conditions, creating: true, its running: false
4. PD-0 logs: region sync with leader meet error: server not started
5. Topology ordering blocks TiKV until PD Running - chicken-and-egg deadlock

## Test plan
- [ ] Kate redeploys TiDB addon with this chart fix on IDC vcluster
- [ ] Verify PD component reaches Running with all 3 pods having roles
- [ ] Verify TiKV and TiDB pods are created after PD Running
- [ ] Run full test suite

### DIFF
--- a/addons/tidb/scripts-ut-spec/pd_role_probe_spec.sh
+++ b/addons/tidb/scripts-ut-spec/pd_role_probe_spec.sh
@@ -1,0 +1,123 @@
+# shellcheck shell=bash
+# shellcheck disable=SC2034,SC2317
+
+Describe "pd_role_probe tests"
+    Include ../scripts/pd_role_probe.sh
+
+    setup() {
+        HOSTNAME="tidb-cluster-tidb-pd-0"
+        PD_POD_FQDN_LIST="tidb-cluster-tidb-pd-0.tidb-cluster-tidb-pd-headless.default.svc.cluster.local,tidb-cluster-tidb-pd-1.tidb-cluster-tidb-pd-headless.default.svc.cluster.local,tidb-cluster-tidb-pd-2.tidb-cluster-tidb-pd-headless.default.svc.cluster.local"
+    }
+
+    BeforeEach 'setup'
+
+    Describe "local API success — leader"
+        timeout() {
+            shift
+            echo '{"leader":{"name":"tidb-cluster-tidb-pd-0"},"members":[{"name":"tidb-cluster-tidb-pd-0"},{"name":"tidb-cluster-tidb-pd-1"},{"name":"tidb-cluster-tidb-pd-2"}]}'
+        }
+
+        It "returns leader when HOSTNAME matches leader.name"
+            When call pd_role_probe
+            The status should be success
+            The stdout should equal "leader"
+        End
+    End
+
+    Describe "local API success — follower"
+        timeout() {
+            shift
+            echo '{"leader":{"name":"tidb-cluster-tidb-pd-1"},"members":[{"name":"tidb-cluster-tidb-pd-0"},{"name":"tidb-cluster-tidb-pd-1"},{"name":"tidb-cluster-tidb-pd-2"}]}'
+        }
+
+        It "returns follower when HOSTNAME is not leader but is member"
+            When call pd_role_probe
+            The status should be success
+            The stdout should equal "follower"
+        End
+    End
+
+    Describe "peer fallback success"
+        peer_call_count=0
+
+        timeout() {
+            shift
+            if [ "$1" = "/pd-ctl" ] && [ "${2:-}" != "-u" ]; then
+                return 1
+            fi
+            echo '{"leader":{"name":"tidb-cluster-tidb-pd-1"},"members":[{"name":"tidb-cluster-tidb-pd-0"},{"name":"tidb-cluster-tidb-pd-1"},{"name":"tidb-cluster-tidb-pd-2"}]}'
+        }
+
+        It "falls back to peer and returns follower"
+            When call pd_role_probe
+            The status should be success
+            The stdout should equal "follower"
+        End
+    End
+
+    Describe "current pod absent from member list"
+        timeout() {
+            shift
+            echo '{"leader":{"name":"tidb-cluster-tidb-pd-1"},"members":[{"name":"tidb-cluster-tidb-pd-1"},{"name":"tidb-cluster-tidb-pd-2"}]}'
+        }
+
+        It "returns failure when HOSTNAME not in members"
+            When call pd_role_probe
+            The status should be failure
+            The stdout should equal ""
+        End
+    End
+
+    Describe "no leader in response"
+        timeout() {
+            shift
+            echo '{"members":[{"name":"tidb-cluster-tidb-pd-0"},{"name":"tidb-cluster-tidb-pd-1"}]}'
+        }
+
+        It "returns failure when no leader field"
+            When call pd_role_probe
+            The status should be failure
+            The stdout should equal ""
+        End
+    End
+
+    Describe "invalid JSON"
+        timeout() {
+            shift
+            echo 'not valid json at all'
+        }
+
+        It "returns failure on invalid JSON"
+            When call pd_role_probe
+            The status should be failure
+            The stdout should equal ""
+        End
+    End
+
+    Describe "all endpoints unreachable"
+        timeout() {
+            shift
+            return 1
+        }
+
+        It "returns failure when all endpoints fail"
+            When call pd_role_probe
+            The status should be failure
+            The stdout should equal ""
+        End
+    End
+
+    Describe "empty PD_POD_FQDN_LIST with local failure"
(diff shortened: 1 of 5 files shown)


weicao:
Blocking focused review at head `c96c565f`.

Pre-flight:
- Scope matches the reported first blocker: PD roleProbe deadlock plus the earlier `externalManaged: true` config-template mismatch.
- Contract checked against `kubeblocks-addon-docs/docs/addon-api/05a-lifecycle-roles-and-probe.md`: roleProbe output must be stable and controller-parseable because it drives service/update/switchover/backup role semantics.
- Static checks passed locally: `git diff --check origin/main...HEAD` and `bash -n addons/tidb/scripts/*.sh`.
- No local Kubernetes runtime was run.

Blockers:

1. Class 1 / Class 4: the fallback still invents `follower` without proving the current pod is a PD member.

The new code accepts the first endpoint that returns `.leader.name`, then emits `follower` for every pod whose `$HOSTNAME` is not the leader. It never checks that `$HOSTNAME` appears in `.members[].name`.

That hides the exact sentinel state we are trying to eliminate: "role unknown / current pod not yet a member" becomes `follower`. In the failure Kate saw, this can make KB mark PD-0 as role-probed even if PD-0 has not actually joined the PD member list. That may unblock TiKV, but it weakens the roleProbe contract and can mask real membership/DNS/bootstrap defects.

Required fix: after getting a member response with leader, verify `any(.members[]; .name == $HOSTNAME)` before outputting any role. If not present, exit non-zero and let KB retry. Only emit:
- `leader` when `leader.name == $HOSTNAME`
- `follower` when `$HOSTNAME` is present in members and is not leader
- no stdout + non-zero when leader is empty or current pod is absent from members

2. Production-script runtime contract: the local `/pd-ctl member` call has no short timeout, so the peer fallback may never execute for the reported failure mode.

The old failure is specifically that local PD API may be unresponsive during region sync / `server not started`. In this PR, line 94 still runs `/pd-ctl member` first without a bounded timeout. If that command hangs until kbagent kills the action, the script never reaches the peer loop, so the PR does not reliably close the root cause.

Required fix: make each `pd-ctl member` attempt bounded, either by setting a roleProbe `timeoutSeconds` plus a shorter per-endpoint `timeout` wrapper, or by skipping local-first and iterating `PD_POD_FQDN_LIST` with bounded per-endpoint attempts. The total action must fit under the kbagent/lifecycle probe budget.

3. Missing negative-path test for the contract above.

This is a runtime lifecycle script change, and the bug is a silent fallback class. Please add at least shellspec or equivalent script tests for:
- local endpoint fails/hangs but peer endpoint proves current pod is a member -> returns follower/leader correctly;
- peer endpoint returns leader but current pod is absent from `.members[]` -> exits non-zero and emits no role;
- no leader / invalid member response -> exits non-zero and emits no role.

ExternalManaged removal looks correct in this diff: PD/TiKV/TiDB no longer carry `externalManaged: true`.

8-class walk:
- Class 1 silent fallback: blocking, unknown/non-member becomes follower.
- Class 2 non-empty contract: role output is non-empty on success, but success predicate is incomplete; covered by blocker 1.
- Class 3 commit-state continuity: no separate mid-commit object state issue in this inline script.
- Class 4 sentinel value: blocking, empty/unknown role state is collapsed into follower.
- Class 5 cleanup pre-state: not applicable.
- Class 7 terminating vs absent: not applicable.
- Class 8 precedence: no complex boolean chain issue seen.

weicao:
Re-review at head `1ed7486e`.

What is fixed:
- Previous blocker 1 is closed in code: after reading `.leader.name`, the script now checks `$HOSTNAME` against `.members[]?.name` before emitting any role. Unknown/non-member no longer becomes `follower`.
- Previous blocker 2 is closed in code: local and peer `pd-ctl member` calls are wrapped with `timeout 2`, so local API hang can fall through to peer fallback before the probe budget is exhausted.
- `externalManaged: true` removal remains correct for PD/TiKV/TiDB.

Direct checks I ran:
- `git diff --check origin/main...HEAD` passed.
- `bash -n addons/tidb/scripts/*.sh` passed.
- Production-image tool check passed for both images relevant to current tests:
  - `docker.io/pingcap/pd:v8.4.0`: `/usr/bin/timeout`, `/usr/bin/bash`, `/usr/bin/jq`, `/pd-ctl` exist.
  - `docker.io/pingcap/pd:v7.5.2`: `/usr/bin/timeout`, `/usr/bin/bash`, `/usr/bin/jq`, `/pd-ctl` exist.
- I also ran a local mocked harness for the inline roleProbe logic covering:
  - local failure + peer proves current pod member -> returns `follower`;
  - peer returns leader but current pod absent from members -> non-zero and no stdout;
  - no leader in response -> non-zero and no stdout.

Remaining blocker: the PR still has no committed negative-path test for this roleProbe contract.

This is a production lifecycle probe and the defect class was exactly silent fallback / sentinel confusion. The current code is small, but without a committed test the next edit can reintroduce `unknown -> follower` and still pass static checks. Please add shellspec or equivalent script-level coverage for at least these cases:

1. local endpoint fails or times out, peer endpoint proves current pod is in `.members[]` and not leader -> outputs `follower`;
2. endpoint returns `.leader.name`, but `$HOSTNAME` is absent from `.members[]` -> exits non-zero and emits no role;
3. response has no leader or invalid JSON -> exits non-zero and emits no role.

A clean way is to move the roleProbe body into `addons/tidb/scripts/pd_role_probe.sh`, include it from `cmpd-pd.yaml` via `.Files.Get`, and add `scripts-ut-spec/pd_role_probe_spec.sh`. That keeps the YAML small and gives us durable regression coverage.

8-class walk after the code fix:
- Class 1 silent fallback: code path now OK; needs regression test.
- Class 2 non-empty contract: success emits `leader|follower`; failure emits no role + non-zero. OK.
- Class 3 commit-state continuity: no stale state reuse issue in this single probe invocation.
- Class 4 sentinel value: code now distinguishes absent leader/member from follower; needs regression test.
- Class 5 cleanup pre-state: not applicable.
- Class 7 terminating vs absent: not applicable.
- Class 8 precedence: no complex boolean chain issue seen.

No local Kubernetes runtime was run. After the test is added, I expect this to be ready for Kate's patched-chart IDC validation.

weicao:
Re-review at head `017c9db5`: no remaining blockers from my focused review.

What changed since the last review:
- `roleProbe` body moved to `addons/tidb/scripts/pd_role_probe.sh` and `cmpd-pd.yaml` includes it via `.Files.Get`.
- `roleProbe.timeoutSeconds: 5` is explicit.
- `scripts-ut-spec/pd_role_probe_spec.sh` now covers the previously-blocking negative paths.

Checks I ran locally:
- `git diff --check origin/main...HEAD` passed.
- `bash -n addons/tidb/scripts/*.sh addons/tidb/scripts-ut-spec/*.sh` passed.
- `shellspec --load-path ./shellspec --default-path addons/tidb/scripts-ut-spec --shell bash` passed: `14 examples, 0 failures`.
- `rg "externalManaged: true|externalManaged" addons/tidb` returned no matches.
- From the previous review of the same code path: `timeout`, `bash`, `jq`, and `/pd-ctl` exist in both `docker.io/pingcap/pd:v8.4.0` and `v7.5.2`.

Contract review:
- Class 1 silent fallback: closed. Unknown/no-member/no-leader paths exit non-zero and emit no role.
- Class 2 non-empty contract: closed. Success emits only `leader` or `follower`; failure emits no stdout.
- Class 3 commit-state continuity: no cross-call state reuse issue in this probe invocation.
- Class 4 sentinel value: closed by member-list validation and negative tests.
- Class 5 cleanup pre-state: not applicable.
- Class 7 terminating vs absent: not applicable.
- Class 8 operator precedence: no complex boolean chain issue seen.

No local Kubernetes runtime was run. Next step is Kate's patched-chart IDC validation pinned to PR head `017c9db5`: verify PD component reaches Running with all PD pods role-labeled, TiKV/TiDB are created, runtime pod `containerStatuses.imageID` captures 7.5.2/8.4.0 image evidence, then run the full TiDB suite.

leon-ape (approved):
LGTM

PR #2878, opened 2026-06-22
chore(clickhouse): guard logConfigs block to prevent YAML null rendering

weicao:
## Summary
- When `logConfigs: {}` in `values.yaml`, the template renders `logConfigs:\n` which YAML parses as null, causing CRD validation errors on ComponentDefinition
- Guard the `logConfigs` block in both `cmpd-ch.yaml` and `cmpd-keeper.yaml` with `{{- if .Values.logConfigs }}`
- Set a sensible default: `error: /bitnami/clickhouse/log/*.err.log` (glob covers both `clickhouse-server.err.log` and `keeper-server.err.log`)

## Changes
- `addons/clickhouse/values.yaml`: `logConfigs: {}` → `logConfigs: {error: /bitnami/clickhouse/log/*.err.log}`
- `addons/clickhouse/templates/cmpd-ch.yaml`: guard logConfigs with `{{- if .Values.logConfigs }}`
- `addons/clickhouse/templates/cmpd-keeper.yaml`: same guard

## Acceptance criteria
- Default install: renders `logConfigs: [{name: error, filePathPattern: /bitnami/clickhouse/log/*.err.log}]`
- `--set logConfigs=null` or `logConfigs: {}`: block not rendered, no YAML null
- Explicit custom path: renders custom pattern correctly

## Test plan
- [ ] `helm template` with default values renders valid logConfigs
- [ ] `helm template` with `--set logConfigs=null` omits logConfigs block
- [ ] `helm template` with explicit custom logConfigs renders correctly
- [ ] vcluster install gate: addon installs cleanly with new defaults

### DIFF
--- a/addons/clickhouse/templates/cmpd-ch.yaml
+++ b/addons/clickhouse/templates/cmpd-ch.yaml
@@ -94,11 +94,13 @@ spec:
       - name: shared-tools
         emptyDir: {}
   updateStrategy: BestEffortParallel
+  {{- if .Values.logConfigs }}
   logConfigs:
     {{- range $name, $pattern := .Values.logConfigs }}
     - name: {{ $name }}
       filePathPattern: {{ $pattern }}
     {{- end }}
+  {{- end }}
   exporter:
     scrapePath: /metrics
     scrapePort: "8001"
--- a/addons/clickhouse/templates/cmpd-ch.yaml
+++ b/addons/clickhouse/templates/cmpd-ch.yaml
@@ -94,7 +94,7 @@ spec:
       - name: shared-tools
         emptyDir: {}
   updateStrategy: BestEffortParallel
-  {{- if .Values.logConfigs }}
+  {{- if not (empty .Values.logConfigs) }}
   logConfigs:
     {{- range $name, $pattern := .Values.logConfigs }}
     - name: {{ $name }}
--- a/addons/clickhouse/templates/cmpd-keeper.yaml
+++ b/addons/clickhouse/templates/cmpd-keeper.yaml
@@ -116,7 +116,7 @@ spec:
           - name: tcp-secure
             targetPort: tcp-secure
             port: 9440
-  {{- if .Values.logConfigs }}
+  {{- if not (empty .Values.logConfigs) }}
   logConfigs:
     {{- range $name, $pattern := .Values.logConfigs }}
     - name: {{ $name }}
--- a/addons/clickhouse/templates/cmpd-keeper.yaml
+++ b/addons/clickhouse/templates/cmpd-keeper.yaml
@@ -116,11 +116,13 @@ spec:
           - name: tcp-secure
             targetPort: tcp-secure
             port: 9440
+  {{- if .Values.logConfigs }}
   logConfigs:
     {{- range $name, $pattern := .Values.logConfigs }}
     - name: {{ $name }}
       filePathPattern: {{ $pattern }}
     {{- end }}
+  {{- end }}
   exporter:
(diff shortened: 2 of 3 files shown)


weicao:
Focused review for head `89aaaad0f13cf94267642a1480d64b2fc4f112e3`.

Findings: no static blockers found.

Scope check:
- Changed files are limited to ClickHouse chart values and the two ComponentDefinition templates.
- The diff directly targets the observed bug: when `logConfigs` is empty/null, the chart must not render a `logConfigs: null` block that the CRD rejects.
- No KB controller/debug-image changes are mixed into this PR.

Contract/design checklist:
- Class 1 silent fallback: acceptable. Default now emits a real error-log pattern instead of hiding the empty default as `null`.
- Class 2 non-empty field: no new required field is introduced; when rendered, each item still has `name` and `filePathPattern` from the map entry.
- Class 3/5/7 state/cleanup: not applicable; chart-only change.
- Class 4 sentinel/null: addressed by guarding the whole block with `if .Values.logConfigs`; empty map/null should omit the field instead of serializing a null value.
- Class 6 NotFound and class 8 precedence: not applicable.

Evidence boundary: I verified this statically by reviewing the current diff and ClickHouse log paths in the chart (`clickhouse-server.err.log` and `keeper-server.err.log` both live under `/bitnami/clickhouse/log/`). I could not run `helm template` locally because this machine has no helm CLI.

Before merge, please attach the IDC `helm template` evidence for these three cases:
1. default values: `logConfigs` renders as an array with `error: /bitnami/clickhouse/log/*.err.log`;
2. empty/null override: no `logConfigs:` block is rendered;
3. explicit custom override: custom pattern renders correctly.

After that, one minimal vcluster install/upgrade gate is enough for this PR.

weicao:
Focused re-review for head `8048f4b9e88fdbfac13c38a30d992e971e883bfe`.

Verdict: static review has no blocker for the logConfigs null-rendering fix. Merge gate still needs the file-based helm template evidence that Lisa is running.

Scope/current-head check:
- Delta from previously reviewed `89aaaad0` is only the two guard changes in `cmpd-ch.yaml` and `cmpd-keeper.yaml`.
- Full PR scope remains limited to ClickHouse chart values plus the two ComponentDefinition templates.
- No KB debug patch or unrelated cleanup is mixed in.

Contract check:
- The guard now wraps the whole `logConfigs:` key block with `not (empty .Values.logConfigs)`, so nil and empty-map values should omit the key rather than rendering YAML null.
- Default values still provide `logConfigs.error: /bitnami/clickhouse/log/*.err.log`, preserving the intended default one-item render.
- Custom non-empty maps still enter the range and render list items.

XP design-contract checklist:
- Class 1 silent fallback: no fallback hides the bug; empty values omit the optional field.
- Class 2 non-empty field: when rendered, each item still comes from map key/value; file-based template gate must validate non-empty real output.
- Class 3 commit-state discontinuity: not applicable; static template render only.
- Class 4 sentinel nil/empty confusion: addressed by `empty`, covering nil and empty map/slice/string before emitting the key.
- Class 5 cleanup pre-state: not applicable.
- Class 7 terminating-vs-absent: not applicable.
- Class 8 precedence: boolean expression is simple and parenthesized by template function call shape.

Evidence boundary: I do not have local `helm` in this environment, so this is static contract review only. Required before merge: helm template gates with real values files: default values => one item; `logConfigs: null` => no `logConfigs:` block; `logConfigs: {}` => no `logConfigs:` block; custom values file => expected custom/default items.

leon-ape (approved):
LGTM

PR #2882, opened 2026-06-22
chore(qdrant): use KB_LEAVE_MEMBER_POD_FQDN for memberLeave cluster check

weicao:
## Summary

Fix `memberLeave` lifecycle action false-success bug in Qdrant addon. The script silently exits 0 without removing the departing Raft peer during horizontal scale-in.

## Root Cause

**Direct cause**: `member-leave.sh` uses bare `curl` and `jq` commands, but the Qdrant container only has these tools at `/qdrant/tools/curl` and `/qdrant/tools/jq`. When bare `curl` fails (exit 127, command not found), the pipeline `curl ... | grep` returns non-zero, and the script incorrectly treats this as "member already absent" → `exit 0`. KB controller reports the action as successful, but no shard migration or peer removal actually occurs.

**Contract issue**: The script conflates "command execution failed" with "member not in cluster", allowing a false-success exit 0 path. Additionally, `curl` without `--fail` does not return non-zero on HTTP 4xx/5xx, so Qdrant-side rejections (e.g., DELETE a peer that still has shards) would also be silently swallowed.

### Evidence Chain

| Step | Evidence |
|------|----------|
| PATH check on qdrant pod | `curl` and `jq` not in PATH; exist at `/qdrant/tools/` |
| kbagent action log | `memberLeave` result = `"member qdr-smoke-2604-qdrant-3 is not in the cluster\n"` |
| Post scale-in Raft state | 4 peers remain (qdrant-3 still present) |
| Action execution pod | qdrant-0 (targetPodSelector: Any) |

## Fix

1. **Tool paths**: Replace all bare `curl` → `/qdrant/tools/curl` and `jq` → `/qdrant/tools/jq`
2. **Control plane separation**: Use a live peer from `QDRANT_POD_FQDN_LIST` as the control endpoint; the leaving pod may be shutting down. Admin operations (move_shard, DELETE peer) go through the live control endpoint.
3. **Structured membership check**: Find leaving peer ID via `jq` on `/cluster` peer URIs — curl failures can no longer be mistaken for "member absent". Only when the peer is structurally confirmed absent from the cluster JSON does the script exit 0.
4. **HTTP error handling**: `curl -sf` (--silent --fail) catches HTTP 4xx/5xx errors, preventing false-success on Qdrant-side rejections.
5. **Error path discipline**: All error conditions exit/return non-zero with diagnostic messages.
6. **Shard transfer fallback**: Primary shard query goes to the leaving peer (alive during memberLeave); falls back to control endpoint if leaving peer becomes unreachable during transfer.

## Test Plan

- [ ] Manually verify DELETE peer API on frozen vcluster cluster
- [ ] Deploy patched addon chart to vcluster
- [ ] Rerun full smoke suite (S01-S20) — S13 scale-in must show 3 Raft peers after completion

### DIFF
--- a/addons/qdrant/scripts/qdrant-member-leave.sh
+++ b/addons/qdrant/scripts/qdrant-member-leave.sh
@@ -13,7 +13,9 @@ load_common_library() {
 }
 
 load_common_library
-current_pod_fqdn=$(get_target_pod_fqdn_from_pod_fqdn_vars "$QDRANT_POD_FQDN_LIST" "$CURRENT_POD_NAME")
+
+CURL=/qdrant/tools/curl
+JQ=/qdrant/tools/jq
 
 if [ "${TLS_ENABLED:-}" = "true" ]; then
   SCHEME="https"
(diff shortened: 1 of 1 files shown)


leon-ape (dismissed):
LGTM

weicao:
Focused review on current head `d9cd71c3116237069267007ab8cf704c28278129`.

Static checks I ran:
- `bash -n addons/qdrant/scripts/qdrant-member-leave.sh` passed
- `git diff --check origin/main...HEAD` passed
- no bare `curl` / `jq` remain in this script; the `/qdrant/tools/*` direction fixes the direct PATH part of the S13 root cause

Blockers for the runtime contract:

1. `KB_LEAVE_MEMBER_POD_FQDN` is being used as the control endpoint for every operation (`/cluster`, collection shard queries, shard move POST, and `DELETE /cluster/peer/{id}`). In the S13 evidence, kbagent ran `memberLeave` on qdrant-0 while the leaving peer qdrant-3 had already been deleted / was refusing connections. This patch would turn the previous false-success into `exit 1`, but it still cannot remove the stale Raft peer in that observed state.

   Please separate the two concepts:
   - live control endpoint: a reachable Qdrant pod, e.g. current action pod / another live peer from `QDRANT_POD_FQDN_LIST`
   - leaving peer identity: `KB_LEAVE_MEMBER_POD_NAME` / `KB_LEAVE_MEMBER_POD_FQDN`, used to resolve the peer id from a successful live `/cluster` response

   Important follow-on: once `/cluster` is queried from a live peer, `.result.peer_id` is the live peer id, not the leaving peer id. The leaving peer id needs to be parsed from `.result.peers` by matching the peer URI/name. Only if that parsed peer is absent should the script take the idempotent `exit 0` path.

2. The false-success contract is not fully closed yet. `curl` without `-f` can return rc=0 for HTTP 4xx/5xx, and the script does not verify Qdrant's JSON `status/result` for the mutation paths. In particular, Qdrant documents `DELETE /cluster/peer/{peer_id}` as returning an error if the peer still has shards: https://api.qdrant.tech/api-reference/distributed/remove-peer . With line 86 as written, that application error can still be reported to KB as action success.

   Please make mutation/probe helpers fail closed on both transport errors and Qdrant error responses. At minimum, `move_shard` and `remove_peer` should capture response + rc and only continue / exit 0 after a positive success predicate, not just a process rc.

3. `grep -q "$KB_LEAVE_MEMBER_POD_NAME"` on raw `/cluster` JSON is still a weak membership predicate for the idempotent-success branch. Since the script now depends on `/qdrant/tools/jq`, use structured JSON parsing of `.result.peers` and treat malformed/empty JSON as `exit 1`, not member absent.

Non-blocking but should be tracked: the existing `while true; sleep 1` shard-transfer wait is still outside the single-shot lifecycle-action contract under the kbagent 60s clamp. It is not the direct S13 root cause, so I would split it if we want to keep this PR narrow, but we should not call the memberLeave action fully bounded until that loop is redesigned or deliberately accepted.

XP contract walk-through:
- Pre-flight/current head: reviewed `d9cd71c3`; one changed production lifecycle script; PR body matches the observed PATH + false-success root cause, but the endpoint change broadens behavior.
- Class 1 silent fallback: still present via weak member-absent check and unchecked Qdrant error responses.
- Class 3 state discontinuity: using the leaving pod endpoint assumes the leaving pod is still reachable after KB has already selected another pod for the action; S13 evidence contradicts that assumption.
- Class 4 sentinel/error meaning: process rc alone is being used as success while HTTP/application errors can be encoded in the response body.
- Other classes: no NotFound/terminating cleanup/boolean precedence issue found in this diff.

addons/qdrant/scripts/qdrant-member-leave.sh
```diff
@@ -23,45 +25,53 @@ else
...
+# Check if the leaving member is still in the Raft cluster.
+# Use KB_LEAVE_MEMBER_POD_FQDN directly — QDRANT_POD_FQDN_LIST may already
+# be updated to exclude the leaving pod during scale-in.
+cluster_check=$($CURL $CURL_TLS -s --max-time 10 ${leave_peer_uri}/cluster)
+if [ $? -ne 0 ] || [ -z "$cluster_check" ]; then
+  echo "ERROR: failed to query cluster info from ${leave_peer_uri}/cluster"
+  exit 1
+fi
+
+if ! echo "$cluster_check" | grep -q "$KB_LEAVE_MEMBER_POD_NAME"; then
```
leon-ape: If /cluster returns a non-empty HTTP/API error response, this check falls through to "member not in the cluster" and exits 0, masking the memberLeave failure.

addons/qdrant/scripts/qdrant-member-leave.sh
```diff
@@ -23,48 +25,111 @@ else
...
+# Query cluster state from a live peer.
+cluster_info=$($CURL $CURL_TLS -sf --max-time 10 ${control_uri}/cluster)
+if [ $? -ne 0 ] || [ -z "$cluster_info" ]; then
+  echo "ERROR: failed to query cluster info from ${control_uri}"
+  exit 1
+fi
+
+# Find the leaving peer's ID by matching KB_LEAVE_MEMBER_POD_NAME in peer URIs.
+# Structured jq check — curl/pipeline failures cannot be mistaken for "member absent".
+leave_peer_id=$(echo "$cluster_info" | $JQ -r \
```
leon-ape, line 55: If jq fails to parse /cluster or .result.peers is missing, leave_peer_id becomes empty and the next branch exits 0, masking the memberLeave failure as member absent.

addons/qdrant/scripts/qdrant-member-leave.sh
```diff
@@ -23,48 +25,111 @@ else
...
+
+    col_names=$(echo ${cols} | $JQ -r '.result.collections[].name')
     for col_name in ${col_names}; do
-        col_cluster_info=`curl $CURL_TLS -s ${leave_peer_uri}/collections/${col_name}/cluster`
-        col_shard_count=`echo ${col_cluster_info} | jq -r '.result.local_shards[] | length'`
-        if [[ ${col_shard_count} -eq 0 ]]; then
-            echo "no shards found in collection ${col_name}"
+        # Query the leaving peer for its local shards (still alive during memberLeave).
+        local col_cluster_info
+        col_cluster_info=$($CURL $CURL_TLS -sf ${leave_peer_uri}/collections/${col_name}/cluster)
```
leon-ape, line 108: This initial shard query still requires the leaving pod endpoint to be reachable; if the leaving pod is already unreachable while its Raft peer remains, move_shards returns 1 before the control-endpoint verification path runs.

weicao:
Re-reviewed current head `02aa702415f957714deddc05135e9389f2b76cd3`.

Static checks I ran:
- `bash -n addons/qdrant/scripts/qdrant-member-leave.sh` passed
- `git diff --check origin/main...HEAD` passed
- confirmed no bare `curl` / `jq` remain in this script

Good changes since the previous review:
- control endpoint is now separated from the leaving peer endpoint
- `move_shard` and `DELETE /cluster/peer/{id}` now go through the live control endpoint
- transport / HTTP 4xx/5xx errors are handled with `/qdrant/tools/curl -sf`
- raw `grep` membership check was replaced with jq-based parsing

Remaining blocker: the top-level jq membership parse can still false-succeed as “member absent”. Lines 47-53 assign `leave_peer_id` from jq output, but do not check jq's rc. Because `set -o errexit` is only enabled later inside the lock subshell, a malformed `/cluster` response or missing `.result.peers` makes jq return non-zero, leaves `leave_peer_id` empty, and then line 53 exits 0 as idempotent success. I reproduced the jq side locally: invalid JSON and `{ "result": {} }` both return jq rc=5. This is the same contract class as the original bug: inspection failure is still conflated with confirmed member absence.

Please make this path fail closed. Shape I would expect:

```bash
if ! leave_peer_id=$(echo "$cluster_info" | "$JQ" -er \
  --arg name "$KB_LEAVE_MEMBER_POD_NAME" \
  '.result.peers | to_entries[] | select(.value.uri | contains($name)) | .key'); then
  # Decide separately: valid JSON + peers present + no matching peer => absent/exit 0;
  # malformed JSON / missing peers / jq error => exit 1.
fi
```

Better is to parse in two explicit steps: first validate `.result.peers` is an object from a successful `/cluster` response; then query for matches. Only a valid peers object with zero matches should produce the idempotent `exit 0`. Also validate exactly one numeric `leave_peer_id` and a non-empty numeric `leader_peer_id` before mutations.

Nonblocking but important for the frozen S13 cluster / manual verification plan: the fallback to control endpoint only exists inside the post-move wait loop. The initial local-shard discovery at lines 81-87 still hard-fails if the leaving pod is already unreachable. That is fail-closed, so I am not treating it as the same false-success blocker, but it means this script may not repair an already-deleted leaving pod with existing collections. If the PR's scope includes manual repair on Wendy's frozen cluster, add the same control-endpoint remote_shards fallback before returning from line 86; otherwise document that runtime validation must be a fresh scale-in rerun.

XP contract walk-through on this head:
- Pre-flight/current head: reviewed `02aa7024`; root-cause scope is still one production lifecycle script; endpoint separation now matches the S13 evidence better.
- Class 1 silent fallback: blocker remains at lines 47-53 because jq inspection failure can become idempotent success.
- Class 3 state discontinuity: previous leaving-pod-as-control-endpoint blocker is fixed for admin operations.
- Class 4 sentinel/error meaning: empty `leave_peer_id` still carries both “absent” and “parse failed”; split those states.
- Lifecycle single-shot: existing `while true` remains a tracked follow-up under the kbagent 60s contract; not blocking this narrow root-cause PR unless we decide to claim full bounded-action compliance.

weicao:
Re-reviewed current head `f0199e9adad5e81fbe71d07cb828acfdb7c58203`.

Checks I ran:
- `bash -n addons/qdrant/scripts/qdrant-member-leave.sh` passed
- `git diff --check origin/main...HEAD` passed
- no bare `curl` / `jq` remain in this script
- local jq boundary probe for membership parsing:
  - valid peers + one match => continue with numeric id
  - valid peers + zero matches => idempotent absent / `exit 0`
  - missing `.result.peers` => fail closed
  - malformed JSON => fail closed
  - multiple matches => fail closed

The previous blocker is closed. The membership path now separates parse/schema failure from valid member absence, and guards multiple/non-numeric peer IDs. The earlier control-endpoint issue is also addressed: move/delete go through the live control endpoint, not the leaving pod endpoint.

XP contract walk-through:
- Pre-flight/current head: reviewed `f0199e9a`; changed file remains scoped to the S13 memberLeave root cause and false-success contract.
- Class 1 silent fallback: fixed for the original curl/jq failure and for jq parse/schema failure in membership detection.
- Class 3 state discontinuity: fixed for admin operations by separating live control endpoint from leaving peer identity.
- Class 4 sentinel/error meaning: fixed for empty `leave_peer_id`; empty now only means valid peers object + zero matches.
- Other classes: no new NotFound / terminating-vs-absent / boolean-precedence blocker found in this diff.

Remaining risk is runtime-only: this focused review does not prove the fresh scale-in path succeeds under KB/kbagent timing. Please validate with the patched chart on the preserved workflow: first a narrow/manual API check if needed, then a fresh smoke rerun where S13 must end with 3 Qdrant Raft peers and no stale deleted peer.

weicao:
Zara focused review: approved.

GitHub would not let this account submit a formal Approve review (`Can not approve your own pull request`), so recording the approval as a PR comment.

Evidence reviewed:
- current head `f0199e9adad5e81fbe71d07cb828acfdb7c58203`
- static checks passed: `bash -n addons/qdrant/scripts/qdrant-member-leave.sh`, `git diff --check origin/main...HEAD`, no bare `curl`/`jq`
- membership parsing boundary checked locally: valid match continues, valid absent exits 0, malformed/missing peers/multiple matches fail closed
- runtime smoke evidence from Wendy: 35 PASS / 0 FAIL / 5 SKIP on patched addon, S13 HScale-in passed, Raft peers=3 after scale-in, deleted qdrant-3 absent, baseline data intact

Scope boundary: runtime validation used IDC image-supply override `--set image.repository=apecloud/qdrant`; this validates the memberLeave fix under a working image supply path. The default ACR path `qdrant/qdrant:v1.17.1` remains a separate Layer 1 image-supply follow-up.

leon-ape (approved):
LGTM

PR #2883, opened 2026-06-22
chore(milvus): remove null serviceName from proxy ComponentDefinition

weicao:
## Summary

- Removes explicit `serviceName:  # use default name` from `cmpd-proxy.yaml`
- Helm renders empty YAML value as `null`, but CRD schema requires `type: string`
- Server-side apply rejects: `spec.services[0].serviceName: Invalid value: "null": must be of type string`
- Omitting the field lets KubeBlocks use the default service name as documented

## Root cause

Found during live Milvus addon smoke test (T01 addon install) on vcluster with KB 1.x main branch.

## Test plan

- [ ] `helm template milvus addons/milvus` renders without `serviceName: null`
- [ ] Addon install succeeds on KB 1.x vcluster
- [ ] Milvus standalone cluster creates and reaches Running

### DIFF
--- a/addons/milvus/templates/cmpd-proxy.yaml
+++ b/addons/milvus/templates/cmpd-proxy.yaml
@@ -20,7 +20,6 @@ spec:
         letterCase: MixedCases
   services:
     - name: milvus
-      serviceName:  # use default name
       spec:
         type: ClusterIP
         ports:


weicao:
Focused review on head `d85f20204d73a37ed836c4f1caef5192ae0a4bb9`: no blockers from my side.

Evidence checked:
- Diff is scoped to the observed T01 failure: removes only the explicit empty `serviceName:` from `addons/milvus/templates/cmpd-proxy.yaml`.
- PR head no longer has an empty `serviceName:` in the Milvus templates.
- `git diff --check origin/main...origin/fix-milvus-proxy-servicename` passes.

Contract review:
- This matches the CRD contract: omit optional `serviceName` to use the default; do not render `serviceName: null` because the schema requires a string if the field is present.
- Class 1/2 checked: no silent fallback or unenforced non-empty field remains in this template path.
- Class 5/7 not applicable; this is install-time render/apply, not cleanup.

I could not run `helm template` locally because this machine does not have `helm` installed. The required next evidence is Diana's vcluster rerun: render/install should pass T01, then standalone create should proceed to the next gate.

leon-ape (approved):
LGTM

PR #2884, opened 2026-06-22
fix(etcd): remove null serviceName from peer service definition

weicao:
## Summary
- Same pattern as milvus proxy fix (#2883): empty `serviceName:` in `cmpd.yaml` renders as `null`, which CRD schema rejects (`type: string` required)
- Discovered during live testing of milvus standalone topology on IDC vcluster — etcd addon install fails with server-side apply validation error

## Test plan
- [ ] Diana re-runs smoke suite with etcd addon install on IDC vcluster
- [ ] Verify etcd ComponentDefinition registers without CRD validation error

### DIFF
--- a/addons/etcd/templates/cmpd.yaml
+++ b/addons/etcd/templates/cmpd.yaml
@@ -132,7 +132,6 @@ spec:
             targetPort: client
       disableAutoProvision: true
     - name: peer
-      serviceName:
       spec:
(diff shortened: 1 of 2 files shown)


weicao:
Duplicate — etcd peer serviceName already fixed in PR #2881 (commit 8b49f39). Closing.

PR #2885, opened 2026-06-22
feat(etcd): add ParametersDefinition for Reconfiguring

weicao:
## Summary
- Add `paramsdef.yaml` for etcd addon, enabling KB Reconfiguring OpsRequest
- All parameters are static (etcd reads config at startup only, `restartOnFileChange: true`)
- User-configurable parameters: `auto-compaction-retention`, `snapshot-count`, `heartbeat-interval`, `election-timeout`, `log-level`, etc.
- Immutable parameters (name, data-dir, listen-urls, initial-cluster, TLS) are excluded
- Uses existing `etcd3.paramsDefinition` helper name (`etcd3-pd`)
- CUE schema validates parameter ranges

## Test plan
- [ ] Install updated etcd addon in vcluster
- [ ] Verify `ParametersDefinition` CR created: `kubectl get parametersdefinition etcd3-pd`
- [ ] Create etcd cluster, verify `ComponentParameter` exists
- [ ] Submit `Reconfiguring` OpsRequest to change `auto-compaction-retention` from "1" to "2"
- [ ] Verify OpsRequest Succeed, pods restarted, config updated at `/var/run/etcd/etcd.conf`
- [ ] Run `kubeblocks-tests/etcd/run-tests.sh -t reconfigure` (explicit suite)

### DIFF
--- a/addons/etcd/templates/paramsdef.yaml
+++ b/addons/etcd/templates/paramsdef.yaml
@@ -13,50 +13,24 @@ spec:
   fileFormatConfig:
     format: yaml
   staticParameters:
-    - snapshot-count
-    - heartbeat-interval
-    - election-timeout
-    - quota-backend-bytes
-    - strict-reconfig-check
-    - max-snapshots
-    - max-wals
     - auto-compaction-mode
     - auto-compaction-retention
-    - tls-min-version
-    - tls-max-version
-    - proxy
-    - proxy-failure-wait
-    - proxy-refresh-interval
-    - proxy-dial-timeout
-    - proxy-write-timeout
-    - proxy-read-timeout
-    - enable-pprof
-    - self-signed-cert-validity
     - log-level
     - logger
+    - enable-pprof
+    - snapshot-count
+    - max-snapshots
+    - max-wals
   parametersSchema:
     topLevelKey: EtcdParameter
     cue: |-
       #EtcdParameter: {
-        "snapshot-count"?: int & >=0
-        "heartbeat-interval"?: int & >=50
-        "election-timeout"?: int & >=500
-        "quota-backend-bytes"?: int & >=0
-        "strict-reconfig-check"?: bool
-        "max-snapshots"?: int & >=0
-        "max-wals"?: int & >=0
         "auto-compaction-mode"?: string & "periodic" | "revision"
         "auto-compaction-retention"?: string
-        "tls-min-version"?: string & "TLS1.2" | "TLS1.3"
-        "tls-max-version"?: string & "TLS1.2" | "TLS1.3"
-        "proxy"?: string & "on" | "readonly" | "off"
-        "proxy-failure-wait"?: int & >=0
-        "proxy-refresh-interval"?: int & >=0
-        "proxy-dial-timeout"?: int & >=0
-        "proxy-write-timeout"?: int & >=0
-        "proxy-read-timeout"?: int & >=0
-        "enable-pprof"?: bool
-        "self-signed-cert-validity"?: int & >=1
         "log-level"?: string & "debug" | "info" | "warn" | "error" | "panic" | "fatal"
         "logger"?: string & "zap"
+        "enable-pprof"?: bool
+        "snapshot-count"?: int & >=0
+        "max-snapshots"?: int & >=0
+        "max-wals"?: int & >=0
       }
--- a/addons/etcd/templates/paramsdef.yaml
+++ b/addons/etcd/templates/paramsdef.yaml
@@ -1,0 +1,70 @@
+apiVersion: parameters.kubeblocks.io/v1alpha1
+kind: ParametersDefinition
+metadata:
+  name: {{ include "etcd3.paramsDefinition" . }}
+  labels:
+    {{- include "etcd.labels" . | nindent 4 }}
+  annotations:
+    {{- include "etcd.annotations" . | nindent 4 }}
+spec:
+  componentDef: {{ include "etcd3.cmpdName" . }}
+  templateName: {{ include "etcd3.configTemplate" . }}
+  fileName: etcd.conf
+  fileFormatConfig:
+    format: yaml
+  staticParameters:
+    - wal-dir
+    - snapshot-count
+    - heartbeat-interval
+    - election-timeout
+    - quota-backend-bytes
+    - strict-reconfig-check
+    - max-snapshots
+    - max-wals
+    - auto-compaction-mode
+    - auto-compaction-retention
+    - cipher-suites
+    - tls-min-version
+    - tls-max-version
+    - proxy
+    - proxy-failure-wait
+    - proxy-refresh-interval
+    - proxy-dial-timeout
+    - proxy-write-timeout
+    - proxy-read-timeout
+    - enable-pprof
+    - cors
+    - discovery
+    - discovery-fallback
+    - discovery-proxy
+    - discovery-srv
+    - self-signed-cert-validity
+    - log-level
+    - logger
+    - log-outputs
+  parametersSchema:
+    topLevelKey: EtcdParameter
+    cue: |-
+      #EtcdParameter: {
+        "snapshot-count"?: int & >=0
+        "heartbeat-interval"?: int & >=50
+        "election-timeout"?: int & >=500
+        "quota-backend-bytes"?: int & >=0
+        "strict-reconfig-check"?: bool
+        "max-snapshots"?: int & >=0
+        "max-wals"?: int & >=0
+        "auto-compaction-mode"?: string & "periodic" | "revision"
+        "auto-compaction-retention"?: string
+        "tls-min-version"?: string & "TLS1.2" | "TLS1.3"
+        "tls-max-version"?: string & "TLS1.2" | "TLS1.3"
+        "proxy"?: string & "on" | "readonly" | "off"
+        "proxy-failure-wait"?: int & >=0
+        "proxy-refresh-interval"?: int & >=0
+        "proxy-dial-timeout"?: int & >=0
+        "proxy-write-timeout"?: int & >=0
+        "proxy-read-timeout"?: int & >=0
+        "enable-pprof"?: bool
+        "self-signed-cert-validity"?: int & >=1
+        "log-level"?: string & "debug" | "info" | "warn" | "error" | "panic" | "fatal"
+        "logger"?: string & "zap"
+      }
--- a/addons/etcd/templates/paramsdef.yaml
+++ b/addons/etcd/templates/paramsdef.yaml
@@ -8,12 +8,11 @@ metadata:
     {{- include "etcd.annotations" . | nindent 4 }}
 spec:
   componentDef: {{ include "etcd3.cmpdName" . }}
-  templateName: {{ include "etcd3.configTemplate" . }}
+  templateName: config
   fileName: etcd.conf
   fileFormatConfig:
     format: yaml
   staticParameters:
-    - wal-dir
     - snapshot-count
     - heartbeat-interval
     - election-timeout
@@ -23,7 +22,6 @@ spec:
     - max-wals
     - auto-compaction-mode
     - auto-compaction-retention
-    - cipher-suites
     - tls-min-version
     - tls-max-version
     - proxy
@@ -33,15 +31,9 @@ spec:
     - proxy-write-timeout
     - proxy-read-timeout
     - enable-pprof
-    - cors
-    - discovery
-    - discovery-fallback
-    - discovery-proxy
-    - discovery-srv
     - self-signed-cert-validity
     - log-level
     - logger
-    - log-outputs
   parametersSchema:
     topLevelKey: EtcdParameter
     cue: |-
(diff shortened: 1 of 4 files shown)


weicao:
Reviewed current head `377bf11ac7283d82853f49baa36346e1335ab6cd` against the KB 1.2+ parameter contract. I’m blocking this revision.

Direct checks:
- refreshed `origin/main`; PR diff is scoped to `addons/etcd/templates/paramsdef.yaml` only
- `git diff --check origin/main..HEAD` passed
- local `helm template` was not available on this machine (`helm` not installed), so this is a static/source contract review, not rendered/runtime evidence
- contract anchors: `docs/addon-api/08-parameters-and-config.md` says `templateName` must bind `ComponentDefinition.spec.configs[*].name`, and `externalManaged` is part of the parameter-management chain; `addon-reconfigure-pd-to-cmpd-migration-guide.md` repeats the KB main / 1.2+ requirement

Blocker 1, class 2 contract write-site mismatch: `addons/etcd/templates/paramsdef.yaml:11` sets:

```yaml
templateName: {{ include "etcd3.configTemplate" . }}
```

That renders the ConfigMap template name (`etcd3-config-template-...`). The actual CmpD config entry is `addons/etcd/templates/cmpd.yaml:203`:

```yaml
- name: config
```

KB validates this exact relationship in `validateTemplateName()` / `hasConfigTemplate()` by checking `cmpd.Spec.Configs[*].Name == paramsDef.Spec.TemplateName`. With the current value, the PD will reference a config template the matched ComponentDefinition does not define. Fix this to `templateName: config`.

Blocker 2, class 6 missing side effect: the PR adds a PD but does not make the etcd config entry externally managed. Current etcd config still has only:

```yaml
restartOnFileChange: true
```

For the KB 1.2+ parameter path, the config entry that should produce `ComponentParameter` needs `externalManaged: true`. Sibling addons with PDs do this on the matching CmpD config entry (for example Valkey/RabbitMQ/MongoDB). Without that, fixing `templateName` alone can still leave `ComponentParameter` empty / unindexed. Because this PR changes addon behavior, please include the `cmpd.yaml` change in the same PR, not as a later runtime surprise.

Blocker 3, class 2 / class 4 advertised field not enforced consistently: `staticParameters` lists keys that the CUE schema does not define, e.g. `wal-dir`, `cipher-suites`, `cors`, `discovery`, `discovery-fallback`, `discovery-proxy`, `discovery-srv`, and `log-outputs`. Once a PD has JSON schema, `ValidateComponentParameterAssignments()` rejects assignments whose key is not found in the generated schema (`parameter <key> not found in parameters schema`). So this revision advertises static parameters that a Reconfiguring request cannot actually set. Either add schema entries for every static parameter you expose, or remove unsupported keys from `staticParameters` until the schema covers them.

Other classes walked:
- class 1: no legacy `reloadAction` silent fallback found, good for KB 1.2+ direction
- class 3/5/7/8: no state/cleanup/boolean-chain issues in this one-file diff
- class 4: no empty dynamic-parameter fallback issue; all-static direction is conservative, but the schema/static mismatch above is the sentinel/advertisement problem to fix

After the patch, minimum runtime validation should be: PD becomes Available, new etcd Cluster gets a non-empty `ComponentParameter` for `config`/`etcd.conf`, explicit PR #329 `reconfigure` no longer SKIPs on missing parameter support, the target key updates `/var/run/etcd/etcd.conf`, and `restartOnFileChange` causes the expected pod restart/recreate. Do not put `reconfigure` into the default/release suite until that evidence exists.

addons/etcd/templates/paramsdef.yaml
```diff
@@ -0,0 +1,70 @@
...
+  annotations:
+    {{- include "etcd.annotations" . | nindent 4 }}
+spec:
+  componentDef: {{ include "etcd3.cmpdName" . }}
+  templateName: {{ include "etcd3.configTemplate" . }}
+  fileName: etcd.conf
+  fileFormatConfig:
+    format: yaml
+  staticParameters:
+    - wal-dir
```
leon-ape: staticParameters exposes keys that are absent from parametersSchema, including wal-dir, cipher-suites, cors, discovery, discovery-fallback, discovery-proxy, discovery-srv, and log-outputs. Those advertised parameters will be rejected by schema-based assignment validation.

addons/etcd/templates/paramsdef.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: parameters.kubeblocks.io/v1alpha1
+kind: ParametersDefinition
+metadata:
+  name: {{ include "etcd3.paramsDefinition" . }}
+  labels:
+    {{- include "etcd.labels" . | nindent 4 }}
+  annotations:
+    {{- include "etcd.annotations" . | nindent 4 }}
+spec:
+  componentDef: {{ include "etcd3.cmpdName" . }}
```
leon-ape, line 10: This PR adds a ParametersDefinition for etcd, but the matching ComponentDefinition config entry still lacks externalManaged: true, so the KB parameter chain will not index this config into ComponentParameter.

addons/etcd/templates/paramsdef.yaml
```diff
@@ -0,0 +1,70 @@
...
+  templateName: {{ include "etcd3.configTemplate" . }}
```
leon-ape: templateName renders to the ConfigMap template name, but KB matches ParametersDefinition.spec.templateName against ComponentDefinition.spec.configs[*].name. The matched etcd config entry is named config, so this PD will not bind to the CmpD config entry.

weicao:
Re-reviewed current head `adae55f25655ae9bb0dedd8d5def142767a85ba8`.

Direct evidence:
- PR diff against refreshed `origin/main` is scoped to `addons/etcd/templates/cmpd.yaml` and `addons/etcd/templates/paramsdef.yaml`
- `git diff --check origin/main..HEAD` passed
- `templateName` is now `config`, matching `ComponentDefinition.spec.configs[0].name`
- `cmpd.yaml` now sets `externalManaged: true` on the same `config` entry
- `fileName: etcd.conf` matches the template ConfigMap data key
- static-parameter list and CUE key list compare equal; no leftover advertised static key is missing from schema
- no legacy `reloadAction` / `shellTrigger` path found in the etcd templates
- local `helm` and `cue` CLIs are not installed here, so this is source/static contract evidence, not rendered PD status or runtime reconfigure evidence

The three previous blockers are closed in this head.

XP contract walk:
- pre-flight: PR scope matches the stated ParametersDefinition fix; no unrelated addon files in the diff
- class 1: no silent legacy reloadAction fallback found
- class 2: non-empty binding fields are now present at the write site (`componentDef`, `templateName`, `fileName`, `externalManaged`)
- class 3/5/7/8: no state/cleanup/boolean-chain logic in this patch
- class 4: previous advertised-but-unaccepted static parameter mismatch is closed by making `staticParameters` match CUE schema exactly
- class 6: missing parameter-controller side effect is addressed by adding `externalManaged: true`

Static review is clear from me. Runtime gate remains required before we count etcd reconfigure as supported: install this addon PR head in the target vcluster, verify `ParametersDefinition/etcd3-pd` Available, verify the new etcd Cluster has non-empty `ComponentParameter` for `config`/`etcd.conf`, run PR #329 explicit `reconfigure`, prove `/var/run/etcd/etcd.conf` changed, and prove `restartOnFileChange` caused the expected pod restart/recreate. Keep `reconfigure` out of default/release suite until that evidence exists.

addons/etcd/templates/paramsdef.yaml
weicao: Fixed in adae55f: `templateName` is now `config`, matching `ComponentDefinition.spec.configs[0].name`.

addons/etcd/templates/paramsdef.yaml
weicao: Fixed in adae55f: `externalManaged: true` added to `cmpd.yaml` configs[0].

addons/etcd/templates/paramsdef.yaml
weicao: Fixed in adae55f: removed 8 unschema'd keys (wal-dir, cipher-suites, cors, discovery*, log-outputs) from staticParameters. All remaining static parameters now have matching CUE schema entries.

weicao:
B-small runtime evidence update for head `356eeb6c2761f45dcc587c1fa35b78daa87205cb`.

Scoped result: **B-small reconfigure fresh-run runtime PASS, evidence-complete**.

Pinned candidate:
- Addon PR #2885: `356eeb6c` (B-small PD scope: 8 low-risk staticParameters)
- Tests PR #329: `b8cfbbc` (evidence-capture harness)
- KB/DP: `1.2.0-alpha.2`
- Evidence dir: `evidence/20260623-125924-78316`

Runtime result:
- PASS=9 FAIL=0 SKIP=0
- T01 cluster create PASS
- T02 baseline data + config snapshot PASS
- T03 Reconfiguring OpsRequest PASS / Succeed
- T04 config propagation + restart verification PASS
- T05 post-reconfigure write/read health PASS

Independent evidence verification:
- `componentparameter.yaml`: `externalManaged: true`, `restartOnFileChange: true`, `phase: Finished`, B-small scope confirmed.
- `opsrequest.yaml`: `type: Reconfiguring`, `auto-compaction-retention: "2"`, `phase: Succeed`.
- `opsrequest-events.txt`: normal lifecycle events only: WaitForProgressing, ValidateOpsRequestPassed, ReconfigureStarted, OpsRequestProcessedSuccessfully.
- `etcd-conf-before.txt`: all 3 pods have `auto-compaction-retention: "1"`.
- `etcd-conf-after.txt`: all 3 pods have `auto-compaction-retention: "2"`.
- `pod-uids-before.txt` / `pod-uids-after.txt`: all 3 pod UIDs changed, `restarted=yes`, nodes preserved across node6/node4/node5.
- `SHA256SUMS`: 15 entries; attached evidence files verified by QA.

Boundary:
- This supports the B-small `auto-compaction-retention` restartOnFileChange path on a fresh install.
- This is not release-final evidence.
- This does not prove a broad etcd parameter allowlist is safe.
- This does not cover the earlier stale-CmpD upgrade/migration compatibility failure; that remains separate follow-up evidence.

weicao:
Reviewed current head `684abfdf339493717bf7e476025db122ad891a82` for the chart/CmpD version-bump patch.

Direct checks:
- PR head on GitHub matches `684abfd`.
- Diff from previous B-small head `356eeb6c` is scoped to `addons/etcd/Chart.yaml` and `addons/etcd/releases_notes.yaml`.
- `git diff --check 356eeb6c..684abfd` passed.
- `helm dependency build addons/etcd` passed.
- `helm template etcd addons/etcd --set etcdVersion=3.6.1` passed.
- Rendered output has no stale `1.2.0-alpha.0` / `etcd-3-1.2.0-alpha.0` / old config-template / old script-template references.

Rendered contract check:
- `ComponentDefinition.metadata.name`: `etcd-3-1.2.0-alpha.1`.
- CmpD config template: `etcd3-config-template-1.2.0-alpha.1`.
- Script template: `etcd3-script-template-1.2.0-alpha.1`.
- `ParametersDefinition.spec.componentDef`: `etcd-3-1.2.0-alpha.1`.
- `ParametersDefinition.spec.templateName`: `config`.
- B-small PD scope remains exactly 8 static parameters: `auto-compaction-mode`, `auto-compaction-retention`, `log-level`, `logger`, `enable-pprof`, `snapshot-count`, `max-snapshots`, `max-wals`.

XP design-contract walk:
- Pre-flight: scope matches the confirmed root cause, the existing-install upgrade compatibility gap from changing CmpD `spec.configs.externalManaged` under the same CmpD name.
- Class 1: no silent fallback introduced; the fix creates a new versioned CmpD instead of relying on `helm uninstall && helm install` as an implicit workaround.
- Class 2: non-empty binding fields remain present and consistent (`componentDef`, `templateName`, `fileName`, `externalManaged`, `restartOnFileChange`).
- Class 3/5/7/8: no state, cleanup, terminating/absent, or boolean-chain logic in this two-file patch.
- Class 4: no new sentinel/advertised-parameter mismatch; staticParameters and CUE keys remain aligned for the B-small scope.
- Class 6: no NotFound/side-effect branch touched; the version bump addresses the missing side effect from trying to mutate an immutable same-name CmpD.

Static/render review is clear for `684abfd`.

Boundary:
- CI is still partly in progress at the time of this review (`shellspec-test` still running; chart/shell/pr-label checks seen so far are green).
- Previous fresh-run evidence at `356eeb6c` is now previous-candidate evidence only.
- New head `684abfd` still needs runtime validation: at minimum fresh-install reconfigure gate plus existing-install `helm upgrade` path proving the new `1.2.0-alpha.1` CmpD avoids the stale-CmpD immutable failure.
- This is not release-final evidence and still does not prove broad etcd parameter allowlist safety.

weicao:
Closing per Leon11's direction (#etcd:3afe48c1):

> 低风险的参数 + 配置文件，没必要引入pd
> runtime 可热加载的参数，也没有必要引入pd

etcd addon does not need ParametersDefinition — neither for static (init-only/start-time) nor runtime hot-loadable parameters. Parameter management stays on the configMap/configTemplate path.

The B-small PD implementation was validated (narrow reconfigure PASS + helm upgrade gate PASS), but the product direction doesn't require it for etcd.

weicao:
Supplementary closeout notes (per Shirley TL):

1. The `684abfd` reconfigure + helm-upgrade evidence-complete results are retained as historical validation only, not as merge justification.
2. PR #2894 (switchover fix) and kubeblocks-addon-docs PR #814 (dispatch guard guide) are unaffected by this closure and continue through their own gates.

PR #2886, opened 2026-06-22
feat(milvus): add ParametersDefinition for standalone reconfigure

weicao:
## Summary
- Add `paramdef-standalone.yaml` — ParametersDefinition for milvus-standalone enabling Reconfiguring OpsRequest
- Add `externalManaged: true` to standalone CMPD configSpecs to enable the parameters controller chain
- CUE schema with 5 static parameters: `log.level`, `proxy.maxNameLength`, `proxy.maxFieldNum`, `proxy.maxDimension`, `common.gracefulTime`
- All parameters are static (pod restart via `restartOnFileChange: true`)

## Contract compliance
- `templateName: config` matches CMPD `configSpecs[0].name`
- `fileName: user.yaml` matches template ConfigMap data key
- `externalManaged: true` required per `docs/addon-api/08-parameters-and-config.md`
- CUE ranges verified against Milvus 2.5.x official config docs

## Test plan
- [ ] Helm template render: `helm template` succeeds with PD resource
- [ ] PD registers in cluster: `kubectl get parametersdefinition` shows `milvus-standalone-pd-*`
- [ ] Reconfiguring OpsRequest works end-to-end (covered by companion PR in kubeblocks-tests)

### DIFF
--- a/addons/milvus/templates/paramdef-standalone.yaml
+++ b/addons/milvus/templates/paramdef-standalone.yaml
@@ -1,0 +1,40 @@
+apiVersion: parameters.kubeblocks.io/v1alpha1
+kind: ParametersDefinition
+metadata:
+  name: milvus-standalone-pd-{{ .Chart.Version }}
+  labels:
+    {{- include "milvus.labels" . | nindent 4 }}
+  annotations:
+    {{- include "milvus.annotations" . | nindent 4 }}
+spec:
+  componentDef: {{ include "milvus-standalone.cmpdRegexpPattern" . }}
+  templateName: config
+  fileName: user.yaml
+  fileFormatConfig:
+    format: yaml
+  parametersSchema:
+    topLevelKey: MilvusStandaloneParameter
+    cue: |-
+      #MilvusStandaloneParameter: {
+        ...
+        log?: {
+          ...
+          level?: string & ("debug" | "info" | "warn" | "error" | "panic" | "fatal") | *"info"
+        }
+        proxy?: {
+          ...
+          maxNameLength?: int & >=1 & <=32768 | *255
+          maxFieldNum?: int & >=1 & <=2048 | *64
+          maxDimension?: int & >=1 & <=32768 | *32768
+        }
+        common?: {
+          ...
+          gracefulTime?: int & >=0 & <=600000 | *5000
+        }
+      }
+  staticParameters:
+    - log.level
+    - proxy.maxNameLength
+    - proxy.maxFieldNum
+    - proxy.maxDimension
+    - common.gracefulTime
--- a/addons/milvus/templates/_helpers.tpl
+++ b/addons/milvus/templates/_helpers.tpl
@@ -347,6 +347,7 @@ Milvus user config - standalone
   volumeName: milvus-config
   namespace: {{.Release.Namespace}}
   defaultMode: 420
+  externalManaged: true
   restartOnFileChange: true
 - name: scripts
   template: {{ include "milvus-scripts.configTemplateName" . }}


weicao:
Blocking review on current head `6fb94b25`.

The `ParametersDefinition` direction is right now (`templateName: config`, `fileName: user.yaml`, static parameter list, and `externalManaged: true` on the config item are aligned with the standalone config path), but the CUE contract is still too loose for the exposed nested parameter groups:

- `paramdef-standalone.yaml` keeps `...` inside `log`, `proxy`, and `common`.
- That means a typo like `proxy.maxNamelenght` can still validate as an unknown nested field instead of being rejected by the parameter contract.
- Top-level openness is reasonable because Milvus `user.yaml` also carries other top-level sections, but the child objects for the five exposed parameter groups should be closed unless we intentionally support arbitrary nested keys.

Please remove the nested ellipses under the exposed groups, or add a negative test/evidence path proving unknown nested keys are rejected and cannot silently enter desired/runtime config.

Static evidence from my local review:
- `git diff --check origin/main...pr-2886-review`: passed earlier on the matching branch head.
- I did not run `helm template`/`helm lint` locally because this machine does not have `helm` installed.
- No live K8s reconfigure execution in this review.

weicao:
Follow-up on the nested CUE discussion.

I accept the part of Andrew's reasoning that a fully closed nested object can reject existing unexposed keys from the merged template config. So I am not insisting on the specific implementation "remove every nested `...`" as the only fix.

But the replacement claim — "typo keys are rejected because they are not in `staticParameters` / `dynamicParameters`" — is not supported by KB main source. In `apecloud/kubeblocks` main:

- `pkg/parameters/core/reconfigure_util.go` says that when `StaticParameters` is non-empty and `DynamicParameters` is empty, a non-static updated key is treated as dynamic/default reload behavior (`CheckUpdateDynamicParameters`, around lines 160-170).
- `IsDynamicParameter()` returns true for a key that is not in `StaticParameters` when the dynamic list is empty (around lines 203-210).
- The unit test `TestIsDynamicParameter` explicitly expects `param4` to be dynamic when only `StaticParameters: [param1,param2,param3]` is configured (around lines 160-168 in `reconfigure_util_test.go`).

That means with open nested CUE, a typo such as `proxy.maxNamelenght` can be schema-valid and is not obviously rejected by the static/dynamic lists. MongoDB using the same pattern is precedent, but it is not proof that Milvus should preserve the same guard gap.

Please close this with one of these evidence paths:

1. Add a negative test that submits an unknown nested key such as `proxy.maxNamelenght` and proves the OpsRequest/ComponentParameter path rejects it and does not update runtime config; or
2. Tighten the schema while still preserving the known unexposed template keys, for example by enumerating the existing keys under `log/proxy/common` that must survive full-config validation.

Until that negative path exists, this remains a blocker for the addon PR.

Evidence boundary: source review against KB main plus existing addon docs; no live K8s run in this review.

addons/milvus/templates/paramdef-standalone.yaml
```diff
@@ -0,0 +1,40 @@
...
+    topLevelKey: MilvusStandaloneParameter
+    cue: |-
+      #MilvusStandaloneParameter: {
+        ...
+        log?: {
+          ...
+          level?: string & ("debug" | "info" | "warn" | "error" | "panic" | "fatal") | *"info"
+        }
+        proxy?: {
+          ...
```
leon-ape, line 25: The nested ellipses keep the exposed groups open, so an unknown key such as proxy.maxNamelenght can still pass the CUE schema. With staticParameters set and dynamicParameters empty, KB treats keys not in staticParameters as dynamic, so the typo is not rejected by the static list.

PR #2887, opened 2026-06-22
fix(mongodb): shard pre-terminate action

### LINKED ISSUE #2860: MongoDB: replace removed KB_CLUSTER_COMPONENT_IS_SCALING_IN dependency in shard preTerminate
Parent: #2858

## Symptom

On KubeBlocks 1.1, MongoDB shard scale-in cleanup can be skipped during `preTerminate`.

`addons/mongodb/templates/cmpd-mongodb-shard.yaml` invokes:

```bash
/scripts/mongodb-shard-manage.sh --pre-terminate
```

The script then checks `KB_CLUSTER_COMPONENT_IS_SCALING_IN` before running shard removal logic.

## Problem

`KB_CLUSTER_COMPONENT_IS_SCALING_IN` was one of the legacy built-in lifecycle envs removed from KubeBlocks 1.1. The MongoDB shard `preTerminate` action does not explicitly provide this env.

As a result, `addons/mongodb/scripts/mongodb-shard-manage.sh` can treat the shard as not scaling in and exit before executing the shard draining/removal flow.

This is not a same-name addon-owned env case: the value is not assigned by the script and is not declared in the lifecycle action env block.

## Expected fix direction

Use a supported 1.1 mechanism to determine shard scale-in state, or explicitly pass the required signal through ComponentDefinition vars / lifecycle envs instead of relying on removed built-in env injection.

gnolong:
close https://github.com/apecloud/kubeblocks-addons/issues/2860

### DIFF
--- a/addons/mongodb/templates/cmpd.yaml
+++ b/addons/mongodb/templates/cmpd.yaml
@@ -39,7 +39,7 @@ spec:
       volumeName: mongodb-config
       namespace: {{ .Release.Namespace }}
       defaultMode: 0400  # for only read
-      externalManaged: false
+      externalManaged: true
   logConfigs:
     {{- range $name,$pattern := .Values.logConfigs }}
     - name: {{ $name }}
--- a/addons/mongodb/templates/cmpd.yaml
+++ b/addons/mongodb/templates/cmpd.yaml
@@ -39,7 +39,7 @@ spec:
       volumeName: mongodb-config
       namespace: {{ .Release.Namespace }}
       defaultMode: 0400  # for only read
-      externalManaged: true
+      externalManaged: false
   logConfigs:
     {{- range $name,$pattern := .Values.logConfigs }}
     - name: {{ $name }}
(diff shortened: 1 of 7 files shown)


addons/mongodb/templates/cmpd.yaml
```diff
@@ -39,7 +39,7 @@ spec:
       volumeName: mongodb-config
       namespace: {{ .Release.Namespace }}
       defaultMode: 0400  # for only read
-      externalManaged: true
+      externalManaged: false
```
leon-ape: This changes MongoDB parameter-management behavior: existing ParametersDefinition objects still target mongodb-config, but all matching CmpD config entries are changed to externalManaged: false, so the parameters chain will no longer index those configs into ComponentParameter.

addons/mongodb/templates/shardingdefinition.yaml
```diff
@@ -14,3 +14,16 @@ spec:
     maxShards: 128
   provisionStrategy: Parallel
   updateStrategy: Parallel
+  lifecycleActions:
+    shardRemove:
```
leon-ape, line 18: The shard removal command moved from CmpD preTerminate to ShardingDefinition shardRemove, but the shard ComponentVersion still only defines a preTerminate action image. The new shardRemove action has no matching image override or exec.image, so its execution surface is not proven to use the MongoDB image/tools that the script depends on.

xuriwuyun (approved):

PR #2889, opened 2026-06-22
feat(mariadb): skip user-input SQL errors in reconfigure action to avoid infinite retry

weicao:
Issue: #2968

Classify MariaDB `SET GLOBAL` errors `1231` / `1193` / `1064` as user-input errors in both base and persisted reconfigure action variants. Skip those parameters instead of exiting retry-safe, preventing controller infinite retry on invalid reconfigure values. Persisted variant skips writing `runtime-overrides.d` for skipped parameters.

## ConfigMap source-of-truth boundary

The controller's ConfigMap remains the single source of truth for desired parameter state. This PR only changes error handling in the reconfigure action script. It does not alter the ConfigMap-to-engine parameter flow.

- Skipped parameters: the engine rejects the `SET GLOBAL`, and the script continues to the next parameter instead of exiting. The ConfigMap still records the user's requested value; the engine simply does not apply it because the value is invalid for that engine version or context.
- `runtime-overrides.d`: the persisted variant skips writing the `.cnf` override file for a skipped parameter. This means a pod restart will not re-apply the invalid value; the engine starts with its defaults, and the ConfigMap can be corrected by the user.
- No silent divergence: the script logs which parameters were skipped and why, including error code and message. The OpsRequest completes without infinite retry, allowing the user to correct the ConfigMap value.

This PR does not change how valid parameters flow from ConfigMap to engine, nor does it introduce any new persistence mechanism.

## Current-head validation

Current PR head: `04ef268fff4e3bcfd089325cb6e6385db680f15d`.

CI is green on this head:

- release-chart variants: pass
- generate-files: pass
- pr-label-check: pass
- shell-check: pass
- shellspec-test: pass

Local ShellSpec on the same head:

- `shellspec --load-path ./shellspec --shell bash addons/mariadb/scripts-ut-spec/reconfigure_persisted_alpha86_spec.sh`: 59 examples, 0 failures

IDC focused validation on the same head:

| Case | Result |
|---|---|
| standalone valid reconfigure `slow_query_log=ON` | PASS: OpsRequest Succeed, variable applied |
| standalone all-invalid reconfigure `slow_query_log=nonsense` + `log_warnings=nonsense` | PASS: OpsRequest Succeed, variables unchanged, Cluster Running |
| standalone mixed reconfigure `innodb_lock_wait_timeout=60` + `slow_query_log=nonsense` | PASS: valid parameter applied, invalid parameter skipped |
| async replication persisted mixed reconfigure `sync_binlog=10` + `slow_query_log=nonsense` | PASS: valid parameter applied and persisted in `runtime-overrides.d`; rejected parameter not written; restart keeps `sync_binlog=10`; Cluster Running |

Result: 6 PASS / 0 FAIL.

Evidence: `/tmp/mariadb-2889-focused-idc2-83536.tar.gz`, sha256 `cf0b09e6d5fa021a9734fd88a43fe197310f603e6417206f9015b0ed3ce9b80f`.

Environment note: the run was on idc3, server `https://192.168.10.201:31742`, namespace `mariadb-test`; the `idc2` segment in the artifact filename is a runner naming leftover. Cleanup was clean.

Invalid-run evidence retained but not counted: `/tmp/mariadb-2889-focused-idc2-38295.tar.gz`, sha256 `a7570c1b953d26f89f63b23958abad4a8591d86924bd5d40f6cd5dd2d1840f4e`; this was a Layer 3 topology/test-scope mistake and is not counted as product PASS or FAIL.

Boundary: this is PR-scoped focused reconfigure validation for head `04ef268f`; it is not a release-ready claim and not post-merge main validation.

### DIFF
--- a/addons/mariadb/templates/_helpers.tpl
+++ b/addons/mariadb/templates/_helpers.tpl
@@ -521,8 +521,8 @@ reconfigure:
             error_code=$(printf '%s' "${output}" | grep -oE 'ERROR [0-9]+' | head -1 | awk '{print $2}')
             case "${error_code}" in
               1231|1232|1193|1064)
-                echo "[SKIP] parameter ${param_name}=${param_value} rejected by engine (error ${error_code}): ${output}"
-                echo "[SKIP] parameter ${param_name}=${param_value} rejected by engine (error ${error_code}): ${output}" >&2
+                echo "[REJECT] parameter ${param_name}=${param_value} rejected by engine (error ${error_code}): ${output}"
+                echo "[REJECT] parameter ${param_name}=${param_value} rejected by engine (error ${error_code}): ${output}" >&2
                 skipped_count=$((skipped_count + 1))
                 ;;
               *)
@@ -533,10 +533,10 @@ reconfigure:
           fi
         done < "${parameter_file}"
 
-        if [ "${applied_count}" -eq 0 ] && [ "${skipped_count}" -gt 0 ]; then
-          echo "No parameters applied; ${skipped_count} had user-input errors (engine rejected)"
-          echo "No parameters applied; ${skipped_count} had user-input errors (engine rejected)" >&2
-          exit 0
+        if [ "${skipped_count}" -gt 0 ]; then
+          echo "${skipped_count} parameter(s) were rejected by engine; failing reconfigure to avoid accepting invalid rendered config"
+          echo "${skipped_count} parameter(s) were rejected by engine; failing reconfigure to avoid accepting invalid rendered config" >&2
+          exit 1
         fi
         if [ "${applied_count}" -eq 0 ]; then
           echo "No parameters were applied during reconfigure action" >&2
@@ -977,8 +977,8 @@ reconfigure:
             error_code=$(printf '%s' "${output}" | grep -oE 'ERROR [0-9]+' | head -1 | awk '{print $2}')
             case "${error_code}" in
               1231|1232|1193|1064)
-                echo "[SKIP] parameter ${param_name}=${param_value} rejected by engine (error ${error_code}): ${output}"
-                echo "[SKIP] parameter ${param_name}=${param_value} rejected by engine (error ${error_code}): ${output}" >&2
+                echo "[REJECT] parameter ${param_name}=${param_value} rejected by engine (error ${error_code}): ${output}"
+                echo "[REJECT] parameter ${param_name}=${param_value} rejected by engine (error ${error_code}): ${output}" >&2
                 skipped_count=$((skipped_count + 1))
                 continue
                 ;;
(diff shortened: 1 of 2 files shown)


addons/mariadb/templates/_helpers.tpl
```diff
@@ -515,11 +516,26 @@ reconfigure:
...
+                ;;
+              *)
+                echo "Failed to set parameter ${param_name}=${param_value}: ${output}" >&2
+                exit 1
+                ;;
+            esac
           fi
         done < "${parameter_file}"
 
+        if [ "${applied_count}" -eq 0 ] && [ "${skipped_count}" -gt 0 ]; then
```
leon-ape: When every assignment is rejected with one of these user-input SQL errors, this still exits 1 after the loop. A Reconfiguring request that contains only an invalid value remains on the retry path, so the infinite-retry case described by this PR is not avoided.

addons/mariadb/templates/_helpers.tpl
```diff
@@ -1018,6 +1046,10 @@ reconfigure:
           #     than catching the same issue at write time.
         done < "${parameter_file}"
 
+        if [ "${applied_count}" -eq 0 ] && [ "${skipped_count}" -gt 0 ]; then
```
leon-ape: When every assignment is rejected with one of these user-input SQL errors, this persisted reconfigure path still exits 1 after the loop. A request that contains only an invalid value remains on the retry path.

weicao:
@leon-ape thanks for the review. The current HEAD (367234b2) addresses both points: in both the base and persisted reconfigure variants, the all-invalid case now exits 0 (lines ~536 and ~1052). The standalone/async/semisync validation in idc3 shows OpsRequest phase=Succeed and cluster returns to Running when all parameters are rejected as user-input errors. Please re-review.

leon-ape (commented):
Returning success after MariaDB rejects all requested parameters leaves the rejected values in the ConfigMap. On restart, the addon still treats that ConfigMap as the parameter source of truth, and the semisync startup reconcile can write an existing runtime override back to the rejected value, so the OpsRequest can succeed while persisting a config state that MariaDB already rejected.

leon-ape:
Current head still exits 1 whenever a user-input SQL error is classified: both reconfigure paths increment skipped_count and then fail when skipped_count > 0. A request containing only rejected values remains on the failing path.

weicao:
Updated the PR body with current-head evidence for `04ef268fff4e3bcfd089325cb6e6385db680f15d`.

Code path check: both base and persisted reconfigure variants now only fail when `applied_count=0` and `skipped_count=0`; an all-invalid request that produces classified user-input SQL errors increments `skipped_count`, logs the rejects, and exits 0 instead of taking the failure path.

Validation now recorded in the PR body:
- CI green on current head.
- Local ShellSpec: 59 examples / 0 failures.
- IDC focused validation: 6 PASS / 0 FAIL, covering all-invalid, mixed, persisted override skip, and restart behavior.
- Evidence: `/tmp/mariadb-2889-focused-idc2-83536.tar.gz`, sha256 `cf0b09e6d5fa021a9734fd88a43fe197310f603e6417206f9015b0ed3ce9b80f`.

Boundary is PR-scoped focused reconfigure validation, not release-ready or post-merge main validation.

leon-ape (approved):
LGTM

JashBook (approved):

PR #2891, opened 2026-06-23
fix(pulsar): move env: inside debugEnabled conditional to avoid null

weicao:
## Summary
- When `debugEnabled=false` (default), initContainers in proxy and bookies-recovery ComponentDefinitions rendered `env:` with no items, producing YAML null
- K8s API rejects: `initContainers[0].env must be of type array: "null"`
- Move `env:` inside the `{{- if .Values.debugEnabled }}` conditional so the key only appears when debug env vars exist

Affects 6 templates: `cmpd-proxy-{2,3,4}.yaml`, `cmpd-bookies-recovery-{2,3,4}.yaml`

## Test plan
- [ ] `helm template` renders without `env: null` when `debugEnabled=false`
- [ ] `helm template --set debugEnabled=true` renders `env:` with debug vars
- [ ] Diana re-run hscale suite after merge

### DIFF
--- a/addons/pulsar/templates/cmpd-bookies-recovery-2.yaml
+++ b/addons/pulsar/templates/cmpd-bookies-recovery-2.yaml
@@ -70,8 +70,8 @@ spec:
         imagePullPolicy: {{ default "IfNotPresent" .Values.image.pullPolicy }}
         command:
           - /kb-scripts/check-bookies.sh
-        env:
         {{- if .Values.debugEnabled }}
+        env:
           - name: PULSAR_LOG_ROOT_LEVEL
             value: DEBUG
           - name: PULSAR_LOG_LEVEL
--- a/addons/pulsar/templates/cmpd-bookies-recovery-3.yaml
+++ b/addons/pulsar/templates/cmpd-bookies-recovery-3.yaml
@@ -70,8 +70,8 @@ spec:
         imagePullPolicy: {{ default "IfNotPresent" .Values.image.pullPolicy }}
         command:
           - /kb-scripts/check-bookies.sh
-        env:
         {{- if .Values.debugEnabled }}
+        env:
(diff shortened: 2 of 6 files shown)


weicao (commented):
Focused review for the HS00 blocker at head `3cef744f960ecd50acaf4d09c7c603f5c143c50f`.

No remaining review blockers from my side.

What I checked:
- Root-cause/scope: diff is limited to the 6 failing Pulsar templates (`cmpd-proxy-{2,3,4}` and `cmpd-bookies-recovery-{2,3,4}`), one-line move per file. No unrelated cleanup or behavior expansion.
- Contract: the null-producing shape was `env:` outside `{{- if .Values.debugEnabled }}` for the initContainers. This PR moves `env:` inside the conditional, so with default `debugEnabled=false` the key is omitted instead of rendering YAML null.
- `debugEnabled=true` path: debug env vars still render under the initContainer `env` list.
- Existing main container env lists in the same files still keep `env:` outside the conditional because they have non-debug entries (`SERVICE_PORT`, etc.) after the conditional; those are not the same null case.
- Local static check: `git diff --check origin/main...HEAD` passes.
- CI evidence observed: chart checks are green (`check-addons-helm`, dockerhub/aliyun variants, cluster helm, generate-files, shell-check). `shellspec-test` was still in progress when I checked.

XP contract checklist: class 1 silent fallback avoided (no default/debug fallback hiding the null); class 2 non-empty/list contract fixed at writer side for the failing initContainers; classes 3-8 not materially involved because this is static template rendering only, with no stateful cleanup/NotFound/termination/precedence path.

Boundary: this is static/template review plus CI chart-check evidence. Diana still needs a fresh HS00-HS04 rerun after merge; the previous HS00 sample remains Pulsar install blocker evidence only and should not enter Milvus hscale PASS/FAIL aggregation.

PR #2895, opened 2026-06-23
chore(redis): enable switchover result check for without-candidate path

weicao:
## Problem

`switchover_without_candidate()` had the `check_switchover_result` call commented out.
After `SENTINEL FAILOVER`, it returns success immediately without verifying the switchover actually happened.

## Fix

Uncomment `check_switchover_result "" "$initial_master" || return 1` so the without-candidate
path polls for master change (up to 300s) and returns failure if the switchover did not complete.

## Test

Added ShellSpec test: when `check_switchover_result` returns failure, `switchover_without_candidate` also returns failure.

Local validation:
- Focused switchover spec: 27/0
- Full Redis scripts suite: 411/0
- `bash -n` PASS, `git diff --check` PASS

### DIFF
--- a/addons/redis/scripts/redis-switchover.sh
+++ b/addons/redis/scripts/redis-switchover.sh
@@ -356,9 +356,7 @@ switchover_without_candidate() {
   # do switchover
   execute_sentinel_failover "$CUSTOM_SENTINEL_MASTER_NAME" || return 1
 
-  # check switchover result using initial_master
-  # if no candidate specified, skip check
-  # check_switchover_result "" "$initial_master" || return 1
+  check_switchover_result "" "$initial_master" || return 1
 }
 
 # This is magic for shellspec ut framework.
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

addons/redis/scripts/redis-switchover.sh
```diff
@@ -356,9 +356,7 @@ switchover_without_candidate() {
   # do switchover
   execute_sentinel_failover "$CUSTOM_SENTINEL_MASTER_NAME" || return 1
 
-  # check switchover result using initial_master
-  # if no candidate specified, skip check
-  # check_switchover_result "" "$initial_master" || return 1
+  check_switchover_result "" "$initial_master" || return 1
```
wangyelei, line 359: When restarting Redis components, the system invokes a switchover without specifying a candidate. A failure in this switchover will block the restart, so avoid testing the switchover without a candidate in isolation.

PR #2896, opened 2026-06-23
fix(milvus): add services to standalone CmpD for Expose support

weicao:
## Summary
- Standalone ComponentDefinition was missing `services` section, causing KB Expose OpsRequest to fail with `component service is not defined, expose operation is not supported`
- Adds `services` block with `milvus:19530` + `metrics:9091`, aligned with proxy CmpD pattern
- Enables Expose OpsRequest (NodePort enable/disable) on standalone topology

## Evidence
- EX01 FAIL on live alpha.2: `component service is not defined, expose operation is not supported, cluster: mvs-exp-5093, component: milvus`
- Root cause: `cmpd-standalone.yaml` has container ports but no `spec.services`
- Proxy CmpD already has correct `services` definition

## Test plan
- [ ] Diana rerun Expose EX00-EX05 with this fix
- [ ] EX01: OpsRequest creates NodePort Service
- [ ] EX03: OpsRequest removes NodePort Service
- [ ] EX05: 0 residual resources

### DIFF
--- a/addons/milvus/templates/cmpd-standalone.yaml
+++ b/addons/milvus/templates/cmpd-standalone.yaml
@@ -10,6 +10,19 @@ spec:
   provider: ApeCloud
   description: {{ .Chart.Description }}
   serviceKind: {{ .Chart.Name }}
+  services:
+    - name: milvus
+      spec:
+        type: ClusterIP
+        ports:
+          - name: milvus
+            port: 19530
+            protocol: TCP
+            targetPort: milvus
+          - name: metrics
+            port: 9091
+            protocol: TCP
+            targetPort: metrics
   configs:
     {{- include "milvus.config.standalone" . | indent 4 }}
   vars:


weicao:
Focused review on current head `ff910c4a`.

No blockers from my side.

Evidence checked:
- Diff scope is one file only: `addons/milvus/templates/cmpd-standalone.yaml`, 13 added lines.
- Root cause match: standalone CmpD had container ports but no `spec.services`, which matches the live Expose blocker `component service is not defined`.
- `git diff --check origin/main...HEAD` passed.
- `helm dependency build addons/milvus` passed.
- `helm lint addons/milvus` passed.
- `helm template milvus addons/milvus --namespace kb-system` passed.
- Render spot-check: `milvus-standalone-1.2.0-alpha.0` now has `spec.services[0].name: milvus` with ports `milvus:19530` and `metrics:9091`; no `serviceName: null` is rendered.

XP contract walk:
- Pre-flight/root cause: scoped to the Expose ComponentService contract gap; no unrelated cleanup.
- API/runtime contract: adds a base ComponentService to the standalone CmpD. Expose Ops can still request output service name `external`.
- Class 1 silent fallback: none.
- Class 2 non-empty contract: service name and ports are explicit/non-empty.
- Class 3 commit-state discontinuity: none; static chart contract.
- Class 4/4a sentinel collision: none.
- Class 5 cleanup pre-state: none.
- Class 6 NotFound short-circuit: none.
- Class 7 terminating-vs-absent: none.
- Class 8 precedence: none.

Residual risk: local/static + CI evidence only. After merge, Diana still needs a fresh Expose EX00-EX05 rerun to prove `external` NodePort Service creation, endpoints, disable cleanup, and data path.

Merge gate: wait for `shellspec-test` to finish; the chart and shell checks I can see are already green.

addons/milvus/templates/cmpd-standalone.yaml
```diff
@@ -10,6 +10,19 @@ spec:
   provider: ApeCloud
   description: {{ .Chart.Description }}
   serviceKind: {{ .Chart.Name }}
+  services:
```
leon-ape, line 13: `spec.services` is added to the existing `milvus-standalone-1.2.0-alpha.0` ComponentDefinition because the chart version is unchanged. Milvus does not render `apps.kubeblocks.io/skip-immutable-check`, so existing installs update the same CmpD immutable hash and the ComponentDefinition becomes unavailable.

PR #2900, opened 2026-06-23
chore(valkey): add per-sentinel timeout to roleProbe check-role.sh

weicao:
## Summary

Deep review W3: the sentinel `masters` query in `check-role.sh` has no per-call timeout. If a single sentinel hangs (TCP level), the script blocks until kbagent SIGKILLs it at the roleProbe `timeoutSeconds` (3s), preventing the remaining sentinels from being queried.

**Fix:** wrap each `valkey-cli sentinel masters` call with `timeout 1` so a hung sentinel fails fast after 1s. With 3 sentinels the worst-case loop budget is 3s, fitting within the 3s probe timeout.

## Changes

- `addons/valkey/scripts/check-role.sh`: add `timeout 1` before `valkey-cli` in sentinel query loop
- `addons/valkey/scripts-ut-spec/check_role_spec.sh`: add structural test asserting the timeout wrapper

## Test plan

- [x] `bash -n` static check PASS
- [x] ShellSpec 83 examples, 0 failures (bash 5.3.12)
- [ ] Emily runtime validation on vcluster

### DIFF
--- a/addons/valkey/scripts-ut-spec/check_role_spec.sh
+++ b/addons/valkey/scripts-ut-spec/check_role_spec.sh
@@ -953,6 +953,16 @@ Describe "Valkey Check-Role Bash Script Tests"
     End
   End
 
+  Describe "per-sentinel timeout guard"
+    check_role_script="../scripts/check-role.sh"
+
+    It "wraps the sentinel masters query with timeout"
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

PR #2901, opened 2026-06-23
chore(valkey): exclude lost+found from restore safety check

weicao:
## Summary

Deep review W4: the restore `prepareData` safety check refuses to proceed if DATA_DIR is non-empty, but ext4 filesystems automatically create a `lost+found` directory on mount. This causes restore to fail on freshly mounted ext4 volumes with `ERROR: DATA_DIR is not empty`.

**Fix:** exclude `lost+found` alongside the existing `.kb-data-protection` placeholder in the `find` command.

## Changes

- `addons/valkey/dataprotection/restore.sh`: add `! -name "lost+found"` to the safety check `find`

## Test plan

- [x] `bash -n` static check PASS
- [x] ShellSpec restore_contract_spec 3 examples, 0 failures
- [ ] Emily runtime validation: restore on ext4 volume


leon-ape (approved):
LGTM

PR #2902, opened 2026-06-23
chore(redis): enable switchover result check for cluster without-candidate path

weicao:
## Summary
- `switchover_without_candidate()` in `redis-cluster-switchover.sh` passed `need_check="false"` to `do_switchover()`, causing it to return success immediately after `CLUSTER FAILOVER` without checking if the command returned OK or if the candidate actually became primary.
- Fix: change `"false"` to `"true"` so the result is verified.

## Root cause
`do_switchover()` has a `need_check` parameter (3rd arg). When `"false"`, lines 161-163 return 0 before checking `$result != "OK"` and before the role-polling loop. `switchover_with_candidate` correctly passes `"true"`; `switchover_without_candidate` incorrectly passed `"false"`.

## Test plan
- [x] New `redis_cluster_switchover_spec.sh` with 9 tests including regression test that fails if `need_check` reverts to `"false"`
- [x] Focused ShellSpec: 9/0 PASS
- [x] Full Redis ShellSpec: 419/0 PASS
- [x] Mutation check: revert to `"false"` → 9/1 FAIL
- [x] Aria focused review PASS
- [ ] Max runtime: cluster without-candidate switchover → OpsRequest completes → primary changed

### DIFF
--- a/addons/redis/redis-cluster-scripts/redis-cluster-switchover.sh
+++ b/addons/redis/redis-cluster-scripts/redis-cluster-switchover.sh
@@ -223,7 +223,7 @@ switchover_without_candidate() {
   fi
 
   # do switchover
-  do_switchover "$candidate_pod" "$candidate_pod_fqdn" "false" || return 1
+  do_switchover "$candidate_pod" "$candidate_pod_fqdn" "true" || return 1
 }
 
 switchover_with_candidate() {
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

addons/redis/redis-cluster-scripts/redis-cluster-switchover.sh
```diff
@@ -223,7 +223,7 @@ switchover_without_candidate() {
   fi
 
   # do switchover
-  do_switchover "$candidate_pod" "$candidate_pod_fqdn" "false" || return 1
+  do_switchover "$candidate_pod" "$candidate_pod_fqdn" "true" || return 1
```
wangyelei, line 226: When restarting Redis components, the system invokes a switchover without specifying a candidate. A failure in this switchover will block the restart, so avoid testing the switchover without a candidate in isolation.

PR #2903, opened 2026-06-23
fix(redis): filter failed masters in Redis6 cluster switchover

weicao:
## Problem

`redis-cluster6-switchover.sh` `get_all_shards_master()` uses `grep "master"` without `grep -v "fail"`, so failed/pfail masters are included when checking if the switchover candidate is known by all shard masters. This can cause false switchover failures when a shard master is down.

The Redis7+ version (`redis-cluster-switchover.sh`) already has this filter at the same line.

## Fix

Add `| grep -v "fail"` to `get_all_shards_master()` line 100, aligning Redis6 with Redis7+.

## Test

New `redis_cluster6_switchover_spec.sh` with 13 tests covering:
- `get_all_shards_master()`: healthy masters only, failed master filtered, pfail master filtered, regression assertion
- `switchover_without_candidate()`: success/failure propagation, no eligible secondary, pod already removed, cluster nodes info failure
- `switchover_with_candidate()`: empty FQDN, success/failure propagation

Mutation check: reverting `grep -v "fail"` → 13/3 FAIL (exactly the 3 filter tests).

Focused ShellSpec: 13/0 PASS
Full Redis scripts ShellSpec: 423/0 PASS

### DIFF
--- a/addons/redis/redis-cluster-scripts/redis-cluster6-switchover.sh
+++ b/addons/redis/redis-cluster-scripts/redis-cluster6-switchover.sh
@@ -97,7 +97,7 @@ get_all_shards_master() {
   fi
   set_xtrace_when_ut_mode_false
 
-  echo "$cluster_nodes_info" | grep "master" | while read -r line; do
+  echo "$cluster_nodes_info" | grep "master" | grep -v "fail" | while read -r line; do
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

PR #2904, opened 2026-06-23
chore(clickhouse): remove unsupported parameter definitions

weicao:
## Summary

Remove the ClickHouse ParametersDefinition path because the current KB fixed-path PD model cannot safely represent ClickHouse XML config with dynamic node names.

This PR:
- deletes the three ClickHouse PD templates
- deletes the two CUE schemas used only by those PDs
- removes unused PD/PCR helpers from `_helpers.tpl`
- replaces the README Reconfiguring example with an explicit unsupported note and points users to custom config templates

## Contract Boundary

`kubeblocks-addon-docs/docs/addon-api/08-parameters-and-config.md` says the reconfigure path must form a closed loop across config template, ParametersDefinition, schema, parameter classification, and reconfigure action. It also says addons should mark reconfigure unsupported when the current PD/ComponentParameter/reconfigure model cannot naturally express the engine configuration.

For ClickHouse, XML config contains dynamic node names such as quota names, remote server cluster names, and profile names. Keeping a PD-backed OpsRequest example exposes a misleading path.

## Scope

In scope:
- addon-side cleanup only
- ClickHouse README boundary update
- no CMPD/config-template/script changes

Out of scope:
- KB controller changes
- Cloud raw-config behavior changes
- kubeblocks-tests PR289 SKIP update; Richard will handle that downstream

## Local Checks

- `rg` found no remaining ClickHouse PD/helper/CUE/PCR-helper references in `addons/clickhouse/templates/`, including `cmpd-ch.yaml` and `cmpd-keeper.yaml`
- `git diff --check`

Not run locally:
- `helm template` because this machine does not have `helm`; rely on addon repo CI/render gate


weicao:
Closing per updated ClickHouse reconfigure direction: keep the existing ParametersDefinition path for now and move reconfigure testing to direct ConfigMap/XML edit + ClickHouse auto-reload validation instead of deleting PDs.

PR #2905, opened 2026-06-23
fix(kafka): remove containerName from resourceFieldRef to fix backup pod failure

weicao:
## Summary
- Backup job pods fail with CreateContainerConfigError: container kafka not found because resourceFieldRef.containerName=kafka references a container that does not exist in the backup pod context
- DP controller propagates env specs from source container to backup job pods; backup pods only have backupdata and manager containers
- Fix: remove containerName: kafka from all 4 ComponentDefinition templates; without containerName, Kubernetes defaults to the current container, which resolves correctly in both normal pod and backup pod contexts
- 4 files changed, 4 lines removed

## Test plan
- [ ] Verify normal kafka pod still gets correct KB_KAFKA_CONTAINER_MEMORY_LIMIT_MIB value (heap ratio unchanged)
- [ ] Verify backup job pod no longer fails with CreateContainerConfigError
- [ ] Run backup-runtime suite end-to-end

### DIFF
--- a/addons/kafka/templates/cmpd-broker-27.yaml
+++ b/addons/kafka/templates/cmpd-broker-27.yaml
@@ -233,7 +233,6 @@ spec:
           - name: KB_KAFKA_CONTAINER_MEMORY_LIMIT_MIB
             valueFrom:
               resourceFieldRef:
-                containerName: kafka
                 resource: limits.memory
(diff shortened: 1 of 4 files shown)


leon-ape (approved):
LGTM

PR #2906, opened 2026-06-23
fix(redis): use ip-port for secondary membership detection

weicao:
## Problem

`redis-cluster6-manage.sh --post-provision` fails with `[ERR] Node is not empty` when creating a Redis 6 cluster.

`check_node_in_cluster` used `$secondary_pod_name` (pod hostname like `redis-shard-b86-1`) to grep `CLUSTER NODES` output. Redis 6 has no `cluster-announce-hostname` support — `CLUSTER NODES` shows only IP addresses (e.g. `10.0.0.1:6379@16379`). The grep always failed, causing the script to re-add already-joined secondaries.

The Redis 7 version (`redis-cluster-manage.sh`) handles this correctly with a `network_mode` branch: pod name for default mode (works because Redis 7 has `cluster-announce-hostname`), IP:port for non-default. The Redis 6 version lacked this branching.

## Fix

Use `$scale_out_shard_secondary_node` (IP:port from `CONFIG GET cluster-announce-ip`) instead of `$secondary_pod_name` in the `check_node_in_cluster` call. This matches the IP-based `CLUSTER NODES` output in Redis 6.

One-line change at `redis-cluster6-manage.sh:854`.

## Tests

- 3 new ShellSpec tests in `redis_cluster6_manage_spec.sh`
- Mutation check: reverting the fix causes 2/3 tests to FAIL
- Full Redis suite: 423/0 PASS

### DIFF
--- a/addons/redis/redis-cluster-scripts/redis-cluster6-manage.sh
+++ b/addons/redis/redis-cluster-scripts/redis-cluster6-manage.sh
@@ -851,7 +851,7 @@ scale_out_redis_cluster_shard() {
   for secondary_pod_name in "${!scale_out_shard_default_other_nodes[@]}"; do
     scale_out_shard_secondary_node="${scale_out_shard_default_other_nodes[$secondary_pod_name]}"
     echo "primary_node_with_port: $primary_node_with_port, primary_node_fqdn: $primary_node_fqdn, mapping_primary_cluster_id: $mapping_primary_cluster_id"
-    if check_node_in_cluster "$primary_node_fqdn" "$primary_node_port" "$secondary_pod_name"; then
+    if check_node_in_cluster "$primary_node_fqdn" "$primary_node_port" "$scale_out_shard_secondary_node"; then
       echo "Secondary node $secondary_pod_name already joined the cluster, skip replicating to primary"
       continue
     fi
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

khalildai:
The 2026-07-31 cherry-pick to `release-1.0` produced a deterministic branch-specific test regression:

- exact `e4da66edf620619b99f1cd39b2b1d849eb2344ca`, ShellSpec run [30610228671](https://github.com/apecloud/kubeblocks-addons/actions/runs/30610228671)
- 477 examples, exactly 3 failures, all in the newly added `redis_cluster6_manage_spec.sh`
- every fixture exits at the release-1.0 guard because it supplies `CURRENT_SHARD_COMPONENT_SHORT_NAME` and `KB_CLUSTER_POD_NAME_LIST`, but omits the branch-required `KB_CLUSTER_POD_HOST_IP_LIST`, `KB_CLUSTER_COMPONENT_POD_NAME_LIST`, and `KB_CLUSTER_COMPONENT_POD_HOST_IP_LIST`
- the same cherry-pick on `release-1.1` exact `1c8380ba8aad98a1282ff176d6d622468d47c41e` passed in run [30610242434](https://github.com/apecloud/kubeblocks-addons/actions/runs/30610242434); that branch has the newer FQDN-only guard

The production one-line IP:port change is not what fails here. The release-1.0 backport needs its test fixture adapted to that branch's built-in environment contract, then a fresh ShellSpec run. Patrol did not rerun or push a follow-up commit.

PR #2918, opened 2026-06-24
chore(oceanbase-ce): harden restore postReady diagnostics

weicao:
## Problem

OBCE soak S1 i11 exposed a restore postReady retry gap: after `ALTER SYSTEM RESTORE` returned `OB_REPLICA_NUM_NOT_ENOUGH(-4672)`, the tenant remained in `status=RESTORE` with `tenant_role=NULL`. A retry then queried only `tenant_role` and failed as `unexpected role 'NULL'`, which hid the more useful restore state.

This PR does not claim to fix the primary 4672 restore root cause. Current candidates are still being investigated separately: OB CE 4.3.5 single-replica restore behavior, missing `locality` restore option, or resource pool/unit allocation state.

## Changes

- Query existing tenants by both `status` and `tenant_role`.
- Treat `status=RESTORE` with empty/NULL role as restore-in-progress/stuck, not as an unknown role.
- Add bounded wait for existing restore tenants with progress logging.
- On restore history failure, timeout, disappeared tenant, unexpected state, or `ALTER SYSTEM RESTORE` failure, print diagnostics from:
  - `DBA_OB_TENANTS`
  - `CDB_OB_RESTORE_PROGRESS`
  - `CDB_OB_RESTORE_HISTORY`
- Replace the restore command backtick execution with direct mysql exit-code checking.

## Verification

- `bash -n addons/oceanbase-ce/dataprotection/restore.sh`
- `helm template obce addons/oceanbase-ce >/tmp/obce-actionset-render.yaml`
- `git diff --check`

Not run: `make scripts-test` because this machine does not have `shellspec`; the Makefile attempted `sudo` install and failed due non-interactive password requirement.

## Boundary

This is L4 retry/idempotence/diagnostic hardening only. Runtime validation still needs an approved vcluster restore run.

### DIFF
--- a/addons/oceanbase-ce/dataprotection/restore.sh
+++ b/addons/oceanbase-ce/dataprotection/restore.sh
@@ -43,13 +43,12 @@ function restoreTenant() {
   local sql=${2:?missing restore sql}
   local time=0
   while true; do
-    if [[ "$tenant_name" == ${TENANT_NAME} ]]; then
+    if [[ "$tenant_name" == "${TENANT_NAME:-}" ]]; then
       echo "INFO: drop init tenant ${tenant_name}"
       ${mysql_cmd} "SET SESSION ob_query_timeout=1000000000; DROP TENANT IF EXISTS ${TENANT_NAME} FORCE;"
     fi
     echo "INFO: $sql"
-    `${mysql_cmd} "${sql}"`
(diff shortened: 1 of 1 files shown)


leon-ape (approved):
LGTM

PR #2919, opened 2026-06-24
fix(redis): build cross-shard CA bundle for TLS cluster bus trust

weicao:
## Problem

KubeBlocks creates per-Component (per-shard) self-signed CAs for Redis Cluster TLS. Each shard gets a different CA fingerprint. The cluster bus (port 16379) performs server-to-server TLS handshake using `tls-ca-cert-file` for certificate validation. Since each shard's CA cannot validate another shard's certificate, the handshake fails:

```
Verify error: self-signed certificate in certificate chain
SSL routines:tls_post_process_server_certificate:certificate verify failed
Verify return code: 19 (self-signed certificate in certificate chain)
```

Gossip messages are never exchanged (`cluster_stats_messages_sent:0`) and the cluster never forms, despite all 6 pods being Running.

Evidence: three shards with three completely different CA sha256 fingerprints + cross-shard cluster bus TLS handshake failure (verify code 19).

## Fix

Addon-level CA bundle workaround:

1. **Post-provision CA exchange** (`redis-cluster-common.sh`): Each shard-0 pod publishes its CA via Redis key `__TLS_PEER_CA__` (using `--insecure` client connection). The first pod to complete collection concatenates all unique CAs into `/data/ca-bundle.crt`, CONFIG SET `tls-ca-cert-file` to reload TLS context, readback verification via CONFIG GET, distributes encoded bundle to ALL pods via `__TLS_CA_BUNDLE_BASE64__`, cleans up exchange keys.

2. **Secondary pod handling** (`redis-cluster-server-start.sh`): Background process waits for `__TLS_CA_BUNDLE_BASE64__` Redis key, writes to `/data/ca-bundle.crt`, CONFIG SET, then DEL key.

3. **Restart handling** (`redis-cluster-server-start.sh`): `build_redis_tls_config()` checks for existing `/data/ca-bundle.crt` on startup and uses it instead of per-shard CA.

4. **Post-provision log** (`redis-cluster-manage.sh`): `exec > >(tee -a /data/post-provision.log) 2>&1` with timestamps and exit codes.

## Boundary

- This PR fixes only cross-shard CA trust (the confirmed root cause)
- SAN/FQDN/cluster-announce-hostname is a separate layer, needs Phase 1 probe evidence
- Concurrent initialization race is a known secondary issue, separate PR
- No private keys are exchanged — only CA public certificates
- Exchange keys are cleaned up after use

### DIFF
--- a/addons/redis/redis-cluster-scripts/redis-cluster-common.sh
+++ b/addons/redis/redis-cluster-scripts/redis-cluster-common.sh
@@ -565,7 +565,21 @@ build_cross_shard_ca_bundle() {
   fi
   set_xtrace_when_ut_mode_false
   if echo "$addslot_output" | grep -qi "already busy"; then
-    echo "Pre-init slot allocation: exchange=$exchange_slot (already owned, reusing from previous attempt)"
+    local myself_slots
+    unset_xtrace_when_ut_mode_false
+    if is_empty "$REDIS_DEFAULT_PASSWORD"; then
+      myself_slots=$($redis_cli -h 127.0.0.1 -p "$service_port" CLUSTER NODES 2>/dev/null | grep "myself" | awk '{for(i=9;i<=NF;i++) print $i}')
+    else
+      myself_slots=$($redis_cli -h 127.0.0.1 -p "$service_port" -a "$REDIS_DEFAULT_PASSWORD" CLUSTER NODES 2>/dev/null | grep "myself" | awk '{for(i=9;i<=NF;i++) print $i}')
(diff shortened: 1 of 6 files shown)


wangyelei:
This issue does not actually exist. You can refer to the code under addons-cluster/redis — it already supports using `user-provided ``tls.crt, ca.crt, and tls.key `instead of relying on the certificates generated by `KubeBlocks`. This way, a shared TLS configuration can be reused across components.

Issue #338, opened 2024-02-29
[BUG] failed to collect metrics for racle-mysql: Error pinging mysqld


sophon-zt:

sophon-zt:
```
ts=2024-02-29T07:10:55.656567222Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.136:0: connect: connection refused"
ts=2024-02-29T07:10:55.656575847Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.137:0: connect: connection refused"
ts=2024-02-29T07:11:25.653445514Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.137:0: connect: connection refused"
ts=2024-02-29T07:11:25.653445555Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.136:0: connect: connection refused"
ts=2024-02-29T07:11:55.656319264Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.137:0: connect: connection refused"
ts=2024-02-29T07:11:55.657286347Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.136:0: connect: connection refused"
ts=2024-02-29T07:12:25.650668097Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.136:0: connect: connection refused"
ts=2024-02-29T07:12:25.650668305Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.137:0: connect: connection refused"
ts=2024-02-29T07:12:55.659321555Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.137:0: connect: connection refused"
ts=2024-02-29T07:12:55.659070472Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.136:0: connect: connection refused"
ts=2024-02-29T07:13:25.652382139Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.136:0: connect: connection refused"
ts=2024-02-29T07:13:25.652332014Z caller=level.go:71 level=error msg="Error pinging mysqld" err="dial tcp 10.42.0.137:0: connect: connection refused"
[...]

PR #2924, opened 2026-06-25
chore(valkey): protect data-path parameters as immutable

weicao:
## Summary

- Add `immutableParameters` section to Valkey ParametersDefinition with `dir`, `dbfilename`, `appendfilename`, `appenddirname`
- Move `dir` from `staticParameters` to `immutableParameters`

These 4 parameters control where Valkey stores its data files. If changed via reconfigure OpsRequest, Valkey would restart and look for data in the wrong location, causing silent data loss. Marking them as `immutableParameters` makes KB reject such changes upfront.

Found during deep review (task #338, review criterion: addon-api 1.2 08-parameters-and-config.md contract).

## Verification

- helm template renders correctly for both valkey8-pd and valkey9-pd
- immutableParameters section appears in both PD objects
- staticParameters retains all other entries unchanged
- No changes to dynamicParameters or CUE schema

### DIFF
--- a/addons/valkey/templates/paramsdef.yaml
+++ b/addons/valkey/templates/paramsdef.yaml
@@ -14,13 +14,17 @@ spec:
   fileName: valkey.conf
   fileFormatConfig:
     format: redis
+  immutableParameters:
+    - dir
+    - dbfilename
+    - appendfilename
+    - appenddirname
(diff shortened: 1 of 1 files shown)


leon-ape (approved):
LGTM

PR #2925, opened 2026-06-25
chore(mariadb): move max_connections to dynamicParameters

weicao:
## Summary
- Move `max_connections` from `staticParameters` to `dynamicParameters` in effect scope
- Move `back_log` from `staticParameters` to `dynamicParameters`

Both parameters support `SET GLOBAL` at runtime in all supported MariaDB versions (10.6+). Classifying them as static forced unnecessary rolling restarts, which in replication topology triggers switchover and service disruption.

## Test plan
- [ ] Reconfigure max_connections via OpsRequest — verify SET GLOBAL applied without restart
- [ ] Reconfigure back_log via OpsRequest — verify SET GLOBAL applied without restart
- [ ] Verify no rolling restart triggered for either parameter change

### DIFF
--- a/addons/mariadb/config/mariadb-config-effect-scope.yaml
+++ b/addons/mariadb/config/mariadb-config-effect-scope.yaml
@@ -3,7 +3,6 @@
 ## dynamicParameters: modifications are applied via SET GLOBAL without restart.
 
 staticParameters:
-  - back_log
   - basedir
   - bind_address
   - character_sets_dir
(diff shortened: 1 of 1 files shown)


leon-ape:
back_log is READ_ONLY in MariaDB (Sys_back_log uses READ_ONLY GLOBAL_VAR(back_log)), so classifying it as dynamic makes the reconfigure action issue SET GLOBAL back_log=..., which fails instead of taking effect without restart. This does not match the target MariaDB variable API.

leon-ape (approved):
LGTM

PR #2926, opened 2026-06-25
chore(mariadb): propagate all SET GLOBAL errors in reloader

weicao:
## Summary
- Fix update-parameter.sh to exit 1 on any SET GLOBAL failure, not just ERROR 1045
- Previously, non-auth errors (out-of-range values, read-only variables, etc.) were silently swallowed with exit 0, causing reconfigure OpsRequest to falsely report success

## Test plan
- [ ] Reconfigure with invalid parameter value — verify OpsRequest fails (not Succeed)
- [ ] Reconfigure with valid parameter value — verify OpsRequest succeeds

### DIFF
--- a/addons/mariadb/reloader/update-parameter.sh
+++ b/addons/mariadb/reloader/update-parameter.sh
@@ -37,12 +37,8 @@ else
 fi
 
 if [[ ${status} -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


leon-ape (approved):
LGTM

PR #2928, opened 2026-06-25
chore(mariadb): replace mutating write probe with read-only SHOW GRANTS

weicao:
## Summary
- Replace INSERT/DELETE write probe in switchover with read-only SHOW GRANTS check
- The old probe created orphan GTIDs on the demoted primary via binlog events
- The remote candidate path was already fixed (alpha.127), but the local 127.0.0.1 path was missed
- New probe checks privilege state via SHOW GRANTS — no binlog side effects

## Test plan
- [ ] Switchover with async replication — verify no orphan GTID events
- [ ] Switchover with semisync replication — verify clean GTID state
- [ ] Verify fence detection still works (fenced account correctly identified)

### DIFF
--- a/addons/mariadb/scripts-ut-spec/replication_switchover_spec.sh
+++ b/addons/mariadb/scripts-ut-spec/replication_switchover_spec.sh
@@ -2049,8 +2049,8 @@ case "$*" in
     echo "GRANT SELECT, PROCESS, RELOAD, REPLICATION SLAVE, REPLICATION CLIENT, REPLICATION MASTER ADMIN ON *.* TO 'root'@'127.0.0.1'"
     exit 0
     ;;
-  *"-uroot"*"INSERT"*)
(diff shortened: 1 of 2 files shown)


leon-ape:
This changes _local_root_write_probe_127 from a write probe to SHOW GRANTS, but the caller still treats __PROBE_RC=0 as writable_unexpected. A fenced root@127.0.0.1 with no write grants returns rc=0 from SHOW GRANTS, sets __PROBE_ERRNO=1290, and _verify_host_is_fenced fails the switchover fence verification.

leon-ape:
`_local_root_write_probe_127` now uses `SHOW GRANTS`; when that query succeeds with non-write grants, it sets `__PROBE_RC=0` and `__PROBE_ERRNO=1290`, but `_verify_host_is_fenced` treats any `write_rc=0` as `writable_unexpected` before checking errno. A fenced `root@127.0.0.1` account can be reported as writable.

leon-ape (approved):
LGTM

PR #2937, opened 2026-06-25
fix(qdrant): render config for backup and restore

shanshanying:
## Summary

- Move the Qdrant config body into `configs/config.yaml.tpl` and load it from the ConfigMap template.
- Render Qdrant API key, TLS, and telemetry settings through `config.yaml` instead of hardcoded runtime env/config fragments.
- Make API-based backup and restore inherit the rendered config by declaring the `qdrant-config` target volume in `BackupPolicyTemplate.spec.backupMethods.targetVolumes.volumeMounts`.
- Keep `runOnTargetPodNode: true` for backup and restore actions because the current dataprotection controller only projects target pod volumes into the job when target-node scheduling is enabled.

## Why

Qdrant deployments can be configured through `config.yaml`, but backup/restore and startup paths previously did not consistently inherit those settings. This keeps API key, TLS, and telemetry behavior config-driven and prevents env-only behavior from bypassing the rendered config.

The snapshot backup/restore implementation is API-based and does not need the data PVC. It does need the rendered Qdrant config. A live check showed that using `targetVolumes.volumeMounts` without `runOnTargetPodNode` left `/qdrant/config` unmounted in the backup job and caused Qdrant API calls to fail with 401. So this PR keeps target-node scheduling for functional correctness while limiting target volume mounts to `qdrant-config` only.

## Validation

- `git diff --check`
- `helm lint addons/qdrant`
- Full stack validation is covered by the final e2e PR in this series.


shanshanying:
Closing this draft because the Qdrant work is being re-split locally into configuration/examples, versions, and memberLeave/e2e branches.

PR #2938, opened 2026-06-25
fix(qdrant): harden member leave lifecycle

shanshanying:
## Summary

- Drain Qdrant collection shards before removing a leaving peer: generate `move_shard` when the leaving peer is the only shard owner, `drop_replica` when another active replica already exists, and wait for in-flight transfers to finish.
- Call Qdrant APIs through a non-leaving live peer, use the packaged `/qdrant/tools/curl`, and prefer lower-ordinal peers as drain targets during multi-pod scale-in.
- Return success only when the leaving peer is already absent or Qdrant `DELETE /cluster/peer/{peer_id}` succeeds; no `force=true` path is used.
- Keep duplicate lifecycle invocations retry-safe: lock contention returns success only after the peer is confirmed absent, otherwise it returns non-zero so KubeBlocks retries.
- Increase the memberLeave retry window to cover shard migration (`timeoutSeconds: 60`, `maxRetries: 120`, `retryInterval: 5`).
- Move liveness/readiness/startup probe bodies into scripts mounted from the scripts ConfigMap.

## Dependency

Depends on #2937.

## Why

Qdrant rejects normal peer removal while the peer still owns collection shards. For scale-in to be both functional and safe, `memberLeave` must be re-entrant: start or continue shard drain, observe transfer completion, then remove the peer, and only then return success so the pod can be deleted.

The implementation avoids forcing peer removal. If drain cannot complete in the current action attempt, the hook exits non-zero and relies on the lifecycle retry policy.

## Validation

- `bash -n addons/qdrant/scripts/qdrant-member-leave.sh`
- `shellcheck --severity=error addons/qdrant/scripts/qdrant-member-leave.sh`
- `helm lint addons/qdrant`
- `git diff --check -- addons/qdrant/scripts/qdrant-member-leave.sh addons/qdrant/templates/cmpd.yaml`
- Full live kind validation is included in dependent PR #2940.


shanshanying:
Closing this draft because the Qdrant work is being re-split locally into configuration/examples, versions, and memberLeave/e2e branches.

PR #2946, opened 2026-06-26
fix(qdrant): support configurable auth and tls backups

### LINKED ISSUE #2934: [BUG] Qdrant backup fails with "parse error: Invalid numeric literal at line 1, column 5 failed with exit code 4"
**Describe the bug**
When creating a backup of qdrant having an api key, the backup fails with "parse error: Invalid numeric literal at line 1, column 5
failed with exit code 4".

**To Reproduce**
1. Create a backuprepo, cluster (having an api key set) and backup job as described here: https://kubeblocks.io/docs/preview/kubeblocks-for-qdrant/05-backup-restore/02-create-full-backup
2. Wait for the backup to start
3. You will see the error in the backup jobs/pods.

**Expected behavior**
I expect the backup job to use the api key which is already gathered from the secret and available as env var `QDRANT__SERVICE__API_KEY`

**Screenshots**

Image

shanshanying:
-    Move the Qdrant config template into a standalone configs file and render the ConfigMap from it.
-  Add config variables for service.api_key and telemetry_disabled, while preserving QDRANT__SERVICE__API_KEY as the higher-priority runtime override.
-  Render TLS settings into config.yaml through the KubeBlocks TLS API and make probes, backup, and restore use a shared qdrant_curl helper so auth and TLS are handled consistently.
-  Mount the rendered qdrant config into backup and restore actions so they can read service.api_key without mounting the data volume.
- fix https://github.com/apecloud/kubeblocks-addons/issues/2934


addons/qdrant/scripts/qdrant-restore.sh
```diff
@@ -26,7 +32,7 @@ for snapshot in $(datasafed list /) ; do
   datasafed pull "${snapshot}" "${SNAPSHOT_DIR}/${snapshot}"
 
   while true; do
-    curl $CURL_TLS -X POST "${QDRANT_SCHEME}://${DP_DB_HOST}:6333/collections/${collection_name}/snapshots/upload?priority=snapshot" \
+    qdrant_curl -sS -f -X POST "${QDRANT_SCHEME}://${DP_DB_HOST}:6333/collections/${collection_name}/snapshots/upload?priority=snapshot" \
```
leon-ape: `qdrant_curl -sS -f ...` runs under `set -e`. On an HTTP upload failure, curl exits non-zero and the script exits before `/tmp/qdrant-restore.log` is checked, so the retry loop is skipped.

examples/qdrant/cluster-with-api-key.yaml
```diff
@@ -0,0 +1,32 @@
...
+  namespace: demo
+spec:
+  terminationPolicy: Delete
+  clusterDef: qdrant
+  topology: cluster
+  componentSpecs:
+    - name: qdrant
+      serviceVersion: 1.10.0
+      env:
+        - name: QDRANT__SERVICE__API_KEY
```
leon-ape, line 14: This example sets `QDRANT__SERVICE__API_KEY` only on the Qdrant component, but the backup/restore ActionSet env only injects `DATA_DIR`. The action pod will not have this API key, and the mounted config has no `service.api_key` for env-key clusters, so the new env-key backup/restore path sends no `api-key` header.

leon-ape (changes requested):
Findings.

examples/qdrant/cluster-with-api-key.yaml
shanshanying: the action pod env is not limited to ActionSet.Spec.Env.

For backup, pkg/dataprotection/backup/request.go builds the job env from targetPod.Spec.Containers[0].Env first and also copies targetPod.Spec.Containers[0].EnvFrom, then appends ActionSet/BackupMethod env. 

So when QDRANT__SERVICE__API_KEY is set on the Qdrant container through the cluster env, the backup/restore action receives it and qdrant_curl sends the api-key header.

addons/qdrant/scripts/qdrant-restore.sh
shanshanying: Fixed. The upload command is now part of the if condition together with the response status check, so a non-zero qdrant_curl -f exit is handled by the retry branch instead of escaping through `set -e`.

addons/qdrant/templates/script-template.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
 data:
   common.sh: |-
     #!/bin/bash
+    {{- .Files.Get "scripts/qdrant-common.sh" | nindent 4 }}
```
leon-ape, line 10: `common.sh` now provides `qdrant_curl`, but `qdrant-member-leave.sh` still calls Qdrant APIs through `$CURL` directly. In API-key clusters, the memberLeave action sends no `api-key` header and scale-in fails.

leon-ape (changes requested):
Findings.

addons/qdrant/templates/script-template.yaml
shanshanying: now sets QDRANT_CURL_BIN=/qdrant/tools/curl and routes all Qdrant API calls through qdrant_curl, so API-key clusters get the api-key header during shard moves and peer removal. I also made qdrant_curl temporarily disable xtrace while it resolves/sends the API key, because memberLeave currently runs with set -x and should not print the key into logs.

leon-ape (approved):
LGTM

cjc7373 (approved):

shanshanying:
Should pick following commits to fix the chery-pick failure action: 
- 47ec9963 fix(qdrant): support configurable auth and tls backups (#2946)
- 2eb87582 fix(qdrant): add memberLeave timeoutSeconds and diagnostic logging (#2910)
- eac698db chore(qdrant): use KB_LEAVE_MEMBER_POD_FQDN for memberLeave cluster check (#2882)

PR #2947, opened 2026-06-26
chore(falkordb): sharding lifecycle actions

### LINKED ISSUE #2861: FalkorDB: replace removed built-in KB_* lifecycle env dependencies
Parent: #2858

## Symptom

On KubeBlocks 1.1, FalkorDB cluster lifecycle actions can miss the topology information they expect during `postProvision` and `preTerminate`.

The affected script is invoked from `addons/falkordb/templates/cmpd-falkordb-cluster.yaml`:

- `postProvision` runs `falkordb-cluster-manage.sh --post-provision`
- `preTerminate` runs `falkordb-cluster-manage.sh --pre-terminate`

Those actions only explicitly inject current-pod envs such as `CURRENT_POD_NAME`, `CURRENT_POD_IP`, and `CURRENT_POD_HOST_IP`.

## Problem

`addons/falkordb/falkordb-cluster-scripts/falkordb-cluster-manage.sh` still reads legacy KubeBlocks built-in lifecycle envs that were removed in 1.1, including:

- `KB_CLUSTER_POD_NAME_LIST`
- `KB_CLUSTER_POD_IP_LIST`
- `KB_CLUSTER_POD_HOST_IP_LIST`
- `KB_CLUSTER_COMPONENT_POD_NAME_LIST`
- `KB_CLUSTER_COMPONENT_POD_HOST_IP_LIST`
- `KB_CLUSTER_COMPONENT_LIST`
- `KB_CLUSTER_COMPONENT_DELETING_LIST`
- `KB_CLUSTER_COMPONENT_UNDELETED_LIST`
- `KB_CLUSTER_COMPONENT_IS_SCALING_IN`

This is not just a same-name addon variable case: these names are not populated by the FalkorDB lifecycle action env block, and the script comments also identify them as built-in lifecycle envs.

## Expected fix direction

Replace the dependency on removed built-in env injection with an explicit supported mechanism, such as ComponentDefinition vars, template-derived values, or addon-owned topology discovery logic.

shanshanying:
- fix: https://github.com/apecloud/kubeblocks-addons/issues/2861

- update shardingdefinition lifecycle with shardRemove
- remove perTerminate from Cluster ComponentDefinitions
- update addons-clusters/falkordb, creating clusters with shardingdef reference
- update falkordb examples
- update shellspec test

### DIFF
--- a/addons/falkordb/templates/cmpv-falkordb-cluster.yaml
+++ b/addons/falkordb/templates/cmpv-falkordb-cluster.yaml
@@ -28,6 +28,7 @@ spec:
       accountProvision: {{ $falkordbRepository }}:{{ .imageTag }}
       switchover: {{ $falkordbRepository }}:{{ .imageTag }}
       preTerminate: {{ $falkordbRepository }}:{{ .imageTag }}
+      shardRemove: {{ $falkordbRepository }}:{{ .imageTag }}
       memberLeave: {{ $falkordbRepository }}:{{ .imageTag }}
       memberJoin: {{ $falkordbRepository }}:{{ .imageTag }}
       metrics: {{ include "metrics.repository" $ }}:{{ $.Values.metrics.image.tag }}
--- a/addons/falkordb/templates/cmpv-falkordb-cluster.yaml
+++ b/addons/falkordb/templates/cmpv-falkordb-cluster.yaml
@@ -27,8 +27,6 @@ spec:
       postProvision: {{ $falkordbRepository }}:{{ .imageTag }}
       accountProvision: {{ $falkordbRepository }}:{{ .imageTag }}
       switchover: {{ $falkordbRepository }}:{{ .imageTag }}
-      preTerminate: {{ $falkordbRepository }}:{{ .imageTag }}
-      shardRemove: {{ $falkordbRepository }}:{{ .imageTag }}
       memberLeave: {{ $falkordbRepository }}:{{ .imageTag }}
       memberJoin: {{ $falkordbRepository }}:{{ .imageTag }}
       metrics: {{ include "metrics.repository" $ }}:{{ $.Values.metrics.image.tag }}
--- a/addons/falkordb/templates/shardingdefinition.yaml
+++ b/addons/falkordb/templates/shardingdefinition.yaml
@@ -21,6 +21,7 @@ spec:
     shardRemove:
       timeoutSeconds: 600
       exec:
+        image: {{ include "falkordb4.image" . }}
         container: falkordb-cluster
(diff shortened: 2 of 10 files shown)


addons/falkordb/templates/shardingdefinition.yaml
```diff
@@ -14,3 +14,18 @@ spec:
     maxShards: 64
   provisionStrategy: Parallel
   updateStrategy: Parallel
+  systemAccounts:
+    - name: default
+      shared: true
+  lifecycleActions:
+    shardRemove:
```
leon-ape, line 21: `shardRemove` is added as a sharding custom action, but the falkordb ComponentVersion still only defines the action image under `preTerminate`. This custom action will not pick up that image key, so the pre-terminate script can run in the default kbagent worker image instead of the FalkorDB image.

leon-ape (changes requested):
Findings.

addons/falkordb/templates/shardingdefinition.yaml
shanshanying: Addressed in the rebased PR branch: `addons/falkordb/templates/cmpv-falkordb-cluster.yaml` now defines `shardRemove` with the same FalkorDB image as `preTerminate`, so the sharding custom action runs in the FalkorDB image instead of the default worker image.

addons/falkordb/templates/cmpv-falkordb-cluster.yaml
```diff
@@ -28,9 +28,10 @@ spec:
       accountProvision: {{ $falkordbRepository }}:{{ .imageTag }}
       switchover: {{ $falkordbRepository }}:{{ .imageTag }}
       preTerminate: {{ $falkordbRepository }}:{{ .imageTag }}
+      shardRemove: {{ $falkordbRepository }}:{{ .imageTag }}
```
leon-ape: `images.shardRemove` is not resolved onto sharding custom actions. KubeBlocks only maps ComponentVersion action images through ComponentDefinition lifecycle actions, while this `shardRemove` is added to `Component.spec.customActions` with no `exec.image`, so it still runs without the FalkorDB action image.

leon-ape (changes requested):
Findings.

addons/falkordb/templates/cmpv-falkordb-cluster.yaml
shanshanying: images.shardRemove removed.

leon-ape (approved):
LGTM

shanshanying:
-  #2960

PR #2955, opened 2026-06-30
fix(redis): set sharding definition in cluster manifests

shanshanying:
## Summary
- set shardingDef: redis-cluster in the Redis sharding example
- render shardingDef: redis-cluster from the Redis cluster Helm chart

## Why
Redis cluster creation can stay in Creating when the sharding spec does not explicitly bind the Redis ShardingDefinition. The Redis addon already defines the cluster topology with shardingDef: redis-cluster, so the example and cluster chart should render the same binding.

### DIFF
--- a/addons-cluster/redis/templates/_helpers.tpl
+++ b/addons-cluster/redis/templates/_helpers.tpl
@@ -17,11 +17,11 @@ Define redis cluster shardingSpec with ComponentDefinition.
 */}}
 {{- define "redis-cluster.shardingSpec" }}
 - name: shard
+  shardingDef: redis-cluster
   shards: {{ .Values.redisCluster.shardCount }}
   template:
     name: redis
     {{- include "redis-cluster.tls" . | indent 4 }}
-    componentDef: redis-cluster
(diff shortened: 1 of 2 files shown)


wangyelei (dismissed):

addons-cluster/redis/templates/_helpers.tpl
```diff
@@ -17,6 +17,7 @@ Define redis cluster shardingSpec with ComponentDefinition.
 */}}
 {{- define "redis-cluster.shardingSpec" }}
 - name: shard
+  shardingDef: redis-cluster
```
leon-ape, line 20: This binds the shard to `redis-cluster`, whose ShardingDefinition template requires `^redis-cluster-\d+`, but the rendered shard template still uses `componentDef: redis-cluster`; the bound template does not match the ShardingDefinition.

shanshanying: ` componentDef: redis-cluster` removed.

leon-ape (approved):
LGTM

PR #2956, opened 2026-06-30
chore: Add keep policy to Milvus config maps

shanshanying:
Summary: Add Milvus config and scripts ConfigMaps to the existing chart annotations helper so versioned ConfigMaps rendered from CMPD references keep helm.sh/resource-policy=keep.


cjc7373 (approved):

leon-ape (approved):
LGTM

PR #2959, opened 2026-06-30
chore(tidb): backport 1.1 branch fix to main

cjc7373:
check commits for more info


leon-ape:
Findings:

- `addons/tidb/templates/_helpers.tpl` unconditionally mounts `component-tls` into PD/TiDB/TiKV, but `component-tls` is only added by `addons-cluster/tidb` when `extra.enableTLSBetweenComponents=true`. The default rendered Cluster sets `KB_ENABLE_TLS_BETWEEN_COMPONENTS=false` and leaves component `volumes` empty, so the generated Pods reference a volume that does not exist.
- With `KB_ENABLE_TLS_BETWEEN_COMPONENTS=true`, `pd_start.sh` advertises PD client URLs as `https://...` and the config templates enable component TLS, but `tikv_start.sh` still starts TiKV with `--pd=http://${PD_ADDRESS}`. TiKV will connect to the TLS-enabled PD client endpoint over plain HTTP.
- `set_component_tls_variables` reads `$KB_ENABLE_TLS_BETWEEN_COMPONENTS` under `set -u` callers such as `pd_start.sh`, `pd_memberleave.sh`, `pd_roleprobe.sh`, `pd_switchover.sh`, and `tikv_member_leave.sh`. When that env is absent, these lifecycle actions exit with `unbound variable` before running.

cjc7373:
reply to above problems:
- this will be addressed in a seperated PR
- tikv will automatically switch to https
- there's no `set -u` set

wangyelei (approved):

PR #2975, opened 2026-07-02
fix(postgresql): initdb with checksum

caiq1nyu:
manual cherry-pick from main branch


leon-ape (commented):
This changes PostgreSQL initdb behavior on the release-1.1 addon path, but the PR has no current-head runtime evidence that a new PostgreSQL cluster provisions with data checksums enabled and remains service-connectable.

leon-ape (approved):

PR #2980, opened 2026-07-02
chore(valkey): unify sentinel monitor registration, remove dead switchover code, redact ACL sync logs

### LINKED ISSUE #2979: valkey: inconsistent sentinel monitor registration, dead switchover code, ACL rule leak in memberJoin logs
Found during a function-by-function review of the valkey addon at main HEAD (1d081a1e), reviewed against kubeblocks-addon-docs/docs/addon-api.

## Problem 1: sentinel monitor registration paths apply different failover tunables

`valkey-register-to-sentinel.sh` (postProvision) and `post-restore-sentinel.sh` (restore postReady) register the master with `down-after-milliseconds 20000`, `failover-timeout 60000`, `parallel-syncs 1`, `auth-user` + `auth-pass`.

The third registration path — `_register_monitor` in `valkey-sentinel-start.sh` (background discovery loop, used when all sentinels restart simultaneously and the monitor stanza is lost) — only issued `SENTINEL MONITOR` + `auth-pass`. A master registered through this path runs with engine defaults (`down-after-milliseconds 30000`, `failover-timeout 180000`) and no `auth-user`, so failover timing silently depends on which registration path happened to run.

**Test scenario / reproduction (deterministic):**
1. Create a replication-topology valkey cluster (3 data + 3 sentinel).
2. Delete the sentinel state on all 3 sentinel PVCs (or recreate sentinel pods with fresh PVCs) so `redis-sentinel.conf` has no `sentinel monitor` stanza, and restart the sentinel pods together.
3. The background discovery loop re-registers the master. `SENTINEL masters` then reports `down-after-milliseconds 30000` / `failover-timeout 180000` instead of the 20000/60000 that postProvision-registered clusters get.

## Problem 2: dead manual-switchover code kept alive only by its unit tests

`switchover.sh` fails closed on the no-Sentinel path (by design), but still carried the entire manual-promotion implementation from before that decision: `promote_replica`, `wait_until_master`, `repoint_one`, `repoint_replicas`, `pick_any_secondary`, plus the superseded `wait_sentinel_sees_priority` (replaced by `wait_sentinel_sees_priority_bias`). None of these has a production call site; the only references are shellspec cases. ~150 lines of dead code and ~325 lines of specs testing code that can never run.

Reproduction: static — `grep` shows no call sites outside `scripts-ut-spec/`.

## Problem 3: memberJoin ACL sync logs full ACL rule payloads

`sync-acl.sh` echoed each replayed rule verbatim (`ACL SETUSER <user> <flags>`); the flags contain password material (sha256 hashes and plain-password tokens for users created that way). These end up in kbagent action logs.

Reproduction (deterministic): create a custom ACL user on the primary, scale out the data component, read the memberJoin action log — the user's full rule line including password hash is printed.

## Problem 4 (minor): raw sleeps and stale comments

- `valkey-start.sh` / `valkey-member-leave.sh` used raw `sleep` in two retry loops instead of `sleep_when_ut_mode_false`, making unit tests sleep for real.
- `cmpd.yaml` claimed `preCondition: RuntimeReady` "ensures sentinel is already up" (actually guaranteed by ClusterDefinition `orders.provision`); `pre-stop.sh` described a `REPLICAOF NO ONE` step that does not exist.

## Validation

- Full valkey shellspec suite green after fix: 299 examples, 0 failures (bash 5), including new behavioral specs for `_register_monitor` (tunables + auth matrix) and a redaction regression test for `sync-acl.sh`.
- `valkey-sentinel-start.sh` gained the standard shellspec sourcing guard so its functions are unit-testable.
- `helm template` renders cleanly.

weicao:
Fixes #2979

Confirmed findings from a function-by-function review of the valkey addon against kubeblocks-addon-docs/docs/addon-api. Runtime-suspected findings from the same review (HA-restore postReady sentinel credential gap, backup AOF consistency) are intentionally NOT in this PR — they will be reproduced on a live cluster first and fixed separately.

## Changes

1. **Unify sentinel monitor registration** (`valkey-sentinel-start.sh`)
   The background discovery loop now applies the same failover tunables as `valkey-register-to-sentinel.sh` and `post-restore-sentinel.sh` (`down-after-milliseconds 20000`, `failover-timeout 60000`, `parallel-syncs 1`) plus `auth-user`, so failover timing no longer depends on which of the three registration paths registered the master. Non-OK `SENTINEL SET` answers are logged instead of silently leaving engine defaults.

2. **Remove dead manual-switchover code** (`switchover.sh`)
   The no-Sentinel path fails closed by design; the pre-decision manual-promotion helpers (`promote_replica`, `wait_until_master`, `repoint_one`, `repoint_replicas`, `pick_any_secondary`) and the superseded `wait_sentinel_sees_priority` had no production call sites and were kept alive only by their unit tests. Removed both the functions (~112 lines) and their specs (~325 lines).

3. **Redact ACL rule payloads in memberJoin logs** (`sync-acl.sh`)
   ACL rule flags contain password material (sha256 / plain tokens); the replay log now prints only the username.

4. **Test hygiene**
   `valkey-sentinel-start.sh` gets the standard shellspec sourcing guard (it was previously untestable — sourcing it executed main); new behavioral spec `valkey_sentinel_start_spec.sh` covers the `_register_monitor` tunables/auth matrix and quorum calculation; `valkey_sync_acl_spec.sh` gains a redaction regression test; two raw `sleep`s in retry loops switched to `sleep_when_ut_mode_false`.

5. **Comment accuracy** (`cmpd.yaml`, `pre-stop.sh`)
   Fixed comments describing behavior the code does not have (postProvision ordering is guaranteed by ClusterDefinition `orders.provision`, not `preCondition`; pre-stop performs no `REPLICAOF NO ONE`).

## Validation

- Full valkey shellspec suite: **299 examples, 0 failures** (bash 5.3; suite also runs ~100s faster due to the sleep fixes). New specs were written first and confirmed red before implementation.
- `helm template` renders cleanly.
- No behavior change on any production path except the background-discovery registration now setting the documented tunables; that path's scenario and reproduction steps are in issue #2979.


leon-ape (approved):
LGTM

PR #2982, opened 2026-07-02
fix(valkey): restore postReady no longer requires unavailable sentinel credentials

### LINKED ISSUE #2981: valkey: replication-topology restore postReady always fails with NOAUTH (no channel injects the sentinel password)
## Problem

`post-restore-sentinel.sh` (restore postReady, ActionSet `valkey-physical-br`) builds its sentinel CLI with `-a "${SENTINEL_PASSWORD}"` **only if that variable is set** — but no channel injects it into the postReady job:

- KubeBlocks DataProtection injects only the data component's target account (`DP_DB_USER`/`DP_DB_PASSWORD`, from BPT `target.account: default`).
- The ActionSet env carries only `DATA_DIR`/`TLS_MOUNT_PATH`; the BPT env schema supports only `versionMapping` here (acknowledged by comments in `backup.sh` and the ActionSet itself: "cross-component Sentinel FQDN/password injection is not available").

Meanwhile the Sentinel component **always** protects its default user with a password: `valkey-sentinel-start.sh` writes `user default on #<sha256> ...` from `SENTINEL_PASSWORD`, which is a `credentialVarRef ... Required` in `cmpd-valkey-sentinel.yaml`.

Consequence: every sentinel `PING` from the postReady job answers `NOAUTH`, the script counts all sentinels unreachable, exits 1, and the postReady action (`onError: Fail`) fails the whole restore of a replication-topology cluster.

1. Install the valkey addon from main; create a replication-topology cluster (3 data + 3 sentinel), write some data.
2. Configure a BackupRepo, run a `datafile` Backup to `Completed`.
3. Restore a new cluster with `examples/valkey/restore.yaml` (replication-9 topology).
4. Observe the postReady job (the one running post-restore-sentinel.sh): expected failure mode is "Sentinel ... not reachable, skipping" for every sentinel followed by "no Sentinel pod was reachable", and the Restore/Cluster never completes.

Expected reproducibility: deterministic on any replication-topology restore (the sentinel password always exists). Please attach the postReady job log to this issue when verifying.

## Fix

Split postReady into two credential-dependent modes:

1. **`SENTINEL_PASSWORD` provided** (user passes it through the restore env): keep the existing fail-closed behavior — register the discovered primary with every expected Sentinel, with the standard failover tunables.
2. **`SENTINEL_PASSWORD` absent** (the default): the job cannot authenticate to Sentinel at all, so it now verifies what it *can* prove with the credentials it has — data-plane convergence (a primary exists among the restored pods and the expected replicas are attached) — and delegates Sentinel monitor registration to the Sentinel startup **self-discovery loop** (`valkey-sentinel-start.sh`), which owns the Sentinel credentials and registers the discovered master itself. In the restore flow the sentinel PVCs are fresh, so the discovery loop always runs.

Also: `find_primary_fqdn` now records the reachable data pod set so the convergence check knows how many replicas to expect, and the script is restructured into functions with the shellspec sourcing guard, adding behavioral unit tests for `verify_replication_converged` and `find_primary_fqdn`.

## Verification asks (for the validating agent)

- Re-run the reproduction above with this fix: restore must complete; postReady log should show "verifying data-plane convergence ... delegated to the Sentinel self-discovery loop" and "replication converged".
- Confirm sentinels converge: `SENTINEL masters` on each sentinel shows the master registered (by the discovery loop) and a post-restore failover drill works.
- Optionally verify mode 1 by passing `SENTINEL_PASSWORD` through the restore env parameter (`dataprotection.kubeblocks.io/restore-env`) and confirming the fail-closed registration path runs.
- Related hardening (registration tunables consistency) is in #2979 / PR #2980; verifying both together is recommended.

weicao:
Fixes #2981 — see the issue for the full failure chain, reproduction steps and verification asks.

**Behavior change**: when SENTINEL_PASSWORD is not provided to the restore (the default, since no API channel injects cross-component credentials into postReady jobs), postReady now verifies data-plane convergence (primary elected, restored replicas attached — bounded poll) instead of failing on NOAUTH sentinel PINGs; sentinel monitor registration is delegated to the sentinel startup self-discovery loop which owns the credentials. When SENTINEL_PASSWORD IS provided via restore env, the original fail-closed sentinel registration path is kept unchanged.

Validation: valkey shellspec suite green locally (post_restore spec: 25 examples, 0 failures — new behavioral tests for verify_replication_converged / find_primary_fqdn); needs live replication backup->restore verification per the issue checklist. Best verified together with #2980 (registration tunables consistency).

### DIFF
--- a/addons/valkey/dataprotection/post-restore-sentinel.sh
+++ b/addons/valkey/dataprotection/post-restore-sentinel.sh
@@ -1,11 +1,27 @@
 #!/bin/bash
-# post-restore-sentinel.sh — postReady phase: re-register the primary with
-# all Sentinel pods after a full cluster restore.
+# post-restore-sentinel.sh — postReady phase: verify the restored cluster
+# converged, and re-arm Sentinel monitoring when credentials allow.
 #
 # After restore the Sentinel conf is empty (Sentinel has its own PVC which
-# was not backed up).  This script discovers the current primary among the
-# data pods and issues SENTINEL monitor + configuration commands on every
-# Sentinel pod, restoring HA monitoring.
+# was not backed up). Sentinel monitor re-registration is owned by TWO
+# mechanisms, selected by credential availability:
+#
+#   1. SENTINEL_PASSWORD available (user passed it through the restore env,
+#      e.g. Cluster spec.restore.parameters
+#      dataprotection.kubeblocks.io/restore-env): this script registers the
+#      discovered primary with every Sentinel pod, fail-closed — restore
+#      only succeeds when all expected Sentinels are configured.
+#
+#   2. SENTINEL_PASSWORD NOT available (the default: KubeBlocks
+#      DataProtection injects only the data-component account into postReady
+#      jobs; there is no API channel for cross-component credentials): this
+#      script CANNOT authenticate to Sentinel — every Sentinel command would
+#      fail with NOAUTH. Instead it verifies data-plane convergence (a
+#      primary exists and the restored replicas are attached) using the
+#      data credentials it does have, and delegates Sentinel monitor
+#      registration to the Sentinel startup self-discovery loop
+#      (valkey-sentinel-start.sh registers the discovered master with the
+#      full failover tunables and it owns the Sentinel credentials).
 #
 # KubeBlocks DataProtection injects:
 #   DP_TARGET_POD_NAME   — name of the restored target pod (e.g. ns-valkey-0)
@@ -17,10 +33,6 @@
 # The script derives Sentinel pod FQDNs from the naming convention:
 #   data pod:     <cluster>-<comp>-<n>
 #   sentinel comp: <cluster>-valkey-sentinel  (standard topology name)
-#
-# Current BackupPolicyTemplate env schema cannot inject cross-component
-# Sentinel credentials. The chart currently supports exactly 3 Sentinel
-# replicas, so the DNS fallback has an independent expected count.
 
 set -e
 set -o pipefail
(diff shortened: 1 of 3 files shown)


leon-ape (commented):
The no-SENTINEL_PASSWORD path can pass before restored replicas attach. find_primary_fqdn builds reachable_data_fqdns from pods reachable during primary discovery, and verify_replication_converged uses reachable_data_fqdns - 1 as the expected replica count; if only the primary is reachable when discovery succeeds, expected_replicas is 0 and postReady succeeds without proving replica convergence.

weicao:
Addressed in `5ae5ff094764e2a2142e7cff8189c66e8f210f3f`.

The credential-less postReady path no longer derives the replica success contract from the partial reachable set captured during primary discovery. For HA restores, it now requires `DATA_REPLICA_COUNT` or `POST_RESTORE_DATA_EXPECTED_COUNT` and waits for `connected_slaves >= expected_data_pod_count - 1`. The replication restore example passes `DATA_REPLICA_COUNT=3` through restore-env.

Checks:
- `bash -n addons/valkey/dataprotection/post-restore-sentinel.sh`
- `git diff --check`
- `post_restore_sentinel_contract_spec.sh`: 27 examples, 0 failures
- `examples/valkey/restore.yaml` YAML parse

weicao:
@leon-ape confirming the current head `5ae5ff09` closes the scenario you described: `resolve_expected_data_pod_count` no longer trusts the reachable-pod scan when it could be partial — with Sentinel endpoints present and no explicit count, postReady now FAILS with guidance to set `DATA_REPLICA_COUNT`/`POST_RESTORE_DATA_EXPECTED_COUNT` via restore-env (behavioral spec: "requires an explicit expected data count for credential-less HA restores" + "does not let a partial primary-only discovery become a zero-replica success"). The reachable-count fallback only remains for the no-Sentinel (standalone) shape where 0 expected replicas is correct. Residual edge — Sentinel DNS fully empty at the same moment as partial data discovery — is mitigated by the documented restore-env example (DATA_REPLICA_COUNT in examples/valkey/restore.yaml); for HA restores that env is the documented procedure.

leon-ape (approved):
LGTM

PR #2984, opened 2026-07-02
fix(valkey): datafile backup archives only the BGSAVE snapshot and ACL file

### LINKED ISSUE #2983: valkey: datafile backup archives live AOF directory — restore ignores the BGSAVE snapshot and can load a torn AOF
## Problem

The `datafile` backup method (`dataprotection/backup.sh`) carefully triggers BGSAVE and waits for it to complete — then archives the **entire live `/data` directory** (`tar -cvf - ./`), including `appendonlydir/` which the server keeps writing during the copy.

Because the chart enables `appendonly yes` (config/valkey-config.tpl), a restored server **prefers the AOF over the RDB** at startup. Consequences:

1. The consistent BGSAVE snapshot the script waited for is effectively ignored on restore; restore fidelity rides on the racy AOF copy.
2. The AOF manifest and its segment files are copied at different instants — the archive can contain a manifest referencing segments in a different rewrite state, or a torn segment tail.
3. The archive also drags in noise: `running.log` and pod-local marker files, which get planted into the restored cluster's data directory.

Redis/Valkey backup practice for RDB-based physical backup is to copy the RDB snapshot alone; backing up a live AOF requires disabling AOF rewrites first.

1. Create a replication-topology valkey cluster; run a continuous write load (e.g. a SET loop with increasing sequence numbers).
2. Trigger a `datafile` Backup while writes continue; let it complete.
3. Inspect the archive: it contains `appendonlydir/` copied mid-write (compare manifest vs segment mtimes/contents across repeated runs), plus `running.log` and marker files.
4. Restore a new cluster from the Backup: the engine loads the archived AOF (not `dump.rdb`); with an inconsistent manifest/segment set the restored dataset can fail to load or load an undefined point-in-time.

The torn-manifest case is probabilistic (window = tar duration vs AOF rewrite/segment rotation); repeat step 2-4 under load to hit it. The "restore point is not the BGSAVE moment" aspect is deterministic.

## Fix

Archive only the consistent snapshot artifacts: `dump.rdb` (guaranteed fresh by the BGSAVE wait) plus `users.acl` (node-local ACL state, needed because ACL is not replicated). On restore, the data directory has no AOF manifest, so the engine falls back to loading `dump.rdb` and seeds a fresh AOF from it — the BGSAVE moment becomes the well-defined restore point.

Trade-off (intentional, please confirm during verification): the restore point is the BGSAVE completion time, not the last write before backup. That is the semantic a snapshot-based `Full` backup method promises; capturing later writes is PITR territory (a future `Continuous` method), not something a racy AOF copy should pretend to provide.

## Verification asks (for the validating agent)

1. Repeat the reproduction with the fix: archive must contain exactly `dump.rdb` (+ `users.acl`), restore must load the RDB (check `INFO persistence` `loading` path / dataset present) and reach Running with data readable and writable.
2. Confirm restore point semantics: writes committed before BGSAVE completion are present after restore; writes after may not be — document expectation.
3. Confirm a restored node regenerates `appendonlydir/` (AOF re-enabled from loaded dataset) — check the engine startup log line about missing AOF manifest falling back to RDB for the exact serviceVersions in the chart (8.x and 9.x).
4. Custom ACL users present in the backup source survive restore (users.acl included).

weicao:
Fixes #2983 — see the issue for the failure chain, reproduction under write load, the restore-point trade-off to confirm, and the verification checklist (archive contents, RDB fallback loading on 8.x/9.x, ACL survival).

Validation: backup/restore contract specs green locally (8 examples, 0 failures); needs live backup->restore verification under write load per the issue.

---

## Merge-gate status (updated 2026-07-06)

**Code side — reviewed** (Athena independent review: multipart-AOF seeding technique verified against Valkey 7+ manifest format; hardcoded file names safe because `dir`/`dbfilename`/`appendfilename`/`appenddirname` are `immutableParameters`; legacy-archive refusal accepted by owner decision — no existing users).

**Live backup→restore evidence under write load — ALL PASS** (Emily, r3 on this PR's head `529e461c`, 32 PASS / 0 FAIL / 0 SKIP, evidence `artifacts/valkey-live/2984-r3/`, log sha256 `575ae976...`, namespace cleaned):

| Check | Result |
|---|---|
| backup on replica target with continuous writer running | PASS (`vlk2984-valkey-1`) |
| archive contains ONLY `./dump.rdb` + `./users.acl` (no AOF artifacts) | PASS (tar listing inspected) |
| restore prepareData seeds multipart AOF manifest from RDB | PASS (log evidence) |
| snapshot-point fidelity: `baseline(100) ≤ restore(109) ≤ post-BGSAVE(123) < final(156)` | PASS — restore lands inside the fork window, post-snapshot writes excluded (the torn-AOF behavior this PR eliminates would land near 156) |
| restored DBSIZE non-zero and equals snapshot-window value | PASS (109) |
| custom ACL account (`backupuser`) survives restore | PASS |

Ready for human approve + merge (second in the train, after #2982).

### DIFF
--- a/addons/valkey/dataprotection/backup.sh
+++ b/addons/valkey/dataprotection/backup.sh
@@ -114,10 +114,23 @@ if [ "${_bgsave_elapsed}" -ge "${_bgsave_timeout}" ]; then
   exit 1
 fi
 
-echo "INFO: Archiving data directory..."
+echo "INFO: Archiving consistent snapshot artifacts..."
 cd "${DATA_DIR}" || { echo "ERROR: cannot cd to DATA_DIR '${DATA_DIR}'" >&2; exit 1; }
-# Archive the entire data directory (includes dump.rdb, appendonlydir/, users.acl)
-tar -cvf - ./ | datasafed push -z zstd-fastest - "${DP_BACKUP_NAME}.tar.zst" || exit 1
+# Archive ONLY the BGSAVE-produced RDB plus the ACL file — NOT the whole
+# data directory. With appendonly enabled the server keeps writing
+# appendonlydir/ while tar runs, so a wholesale copy captures a torn AOF
+# manifest/segment set; on startup the engine PREFERS the AOF over the RDB,
+# which would make restore fidelity ride on that racy copy instead of the
+# consistent BGSAVE snapshot we just waited for. With no AOF manifest in
+# the restored data directory, the engine loads dump.rdb and seeds a fresh
+# AOF from it, making the BGSAVE moment the well-defined restore point.
+if [ ! -f "./dump.rdb" ]; then
+  echo "ERROR: dump.rdb not found in ${DATA_DIR} after BGSAVE" >&2
+  exit 1
+fi
+backup_files=("./dump.rdb")
+[ -f "./users.acl" ] && backup_files+=("./users.acl")
+tar -cvf - "${backup_files[@]}" | datasafed push -z zstd-fastest - "${DP_BACKUP_NAME}.tar.zst" || exit 1
 
 save_sentinel_acl || \
   echo "WARNING: Sentinel ACL save failed — ACL rules will not be restored after a cluster restore." >&2
--- a/addons/valkey/dataprotection/backup.sh
+++ b/addons/valkey/dataprotection/backup.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# backup.sh — physical (RDB+AOF) backup for Valkey.
+# backup.sh — physical RDB+ACL snapshot backup for Valkey.
 #
 # KubeBlocks DataProtection injects:
 #   DP_DB_HOST           — target pod hostname/FQDN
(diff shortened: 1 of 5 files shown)


leon-ape (commented):
This changes the Valkey physical backup archive contract from the live data directory to dump.rdb/users.acl only, but the PR has no current-head live backup-restore evidence under write load showing the restored cluster loads the intended RDB snapshot and preserves ACL state.

weicao:
@leon-ape agreed — this ask is still open and is a merge precondition in our train: Emily will run the live backup→restore under write load (archive contents, restored RDB fidelity, ACL survival) on the current head before this PR merges. Evidence will be appended to the PR body like PR #2988's merge-gate section.

weicao:
@leon-ape the live evidence you asked for is complete on the current head `529e461c` — backup under a continuous write load on a replica target, archive content inspection (only `dump.rdb`+`users.acl`), restored dataset proven to be the BGSAVE fork-window snapshot via a bounded assertion (`baseline 100 ≤ restore 109 ≤ post-BGSAVE 123 < final 156` — the old whole-dir behavior would restore near 156), and ACL account survival. See the "Merge-gate status" section in the PR body (32/0/0, evidence sha `575ae976...`).

leon-ape (approved):
LGTM

Issue #2991, opened 2026-07-02
valkey: self-heal daemon overlaps Sentinel built-ins (+fix-slave-config / +convert-to-slave) — evidence alignment needed before keep/remove decision


weicao:
## This is an investigation item, not a fix PR — deliberately

`valkey-self-heal.sh` is a 446-line always-on daemon inside every data pod (spawned by valkey-start.sh) with three repair loops. Two of them overlap capabilities Sentinel documents as built-in:

| Daemon loop | Sentinel built-in it overlaps |
| --- | --- |
| cascade-topology repair (slave of a slave → REPLICAOF to the chain head) | `+fix-slave-config` — Sentinel reconfigures replicas attached to the wrong master |
| dual-master demote (self-demote when quorum says another master) | `+convert-to-slave` — Sentinel demotes a returned old master |

The code carries real evidence anchors (PR #2615 guards, KB-10196, "Bug 5" full-sync stall), so these failures did happen. The open first-principles question: **were they root-caused to Sentinel genuinely failing to converge, or to the addon's own bootstrap (lexicographic election / REPLICAOF in valkey-start.sh) racing Sentinel and creating masters outside Sentinel's view?** If the latter, the durable fix is tightening bootstrap, not running a second control loop that issues REPLICAOF behind Sentinel's back — two controllers repairing the same state is itself a failure source (the daemon's own guards — 120s promotion grace, quorum-clear, skip-stale-role — exist to avoid fighting Sentinel).

## Asks

For each of the three loops, produce or retrieve the original failure evidence and answer "why was the Sentinel built-in insufficient":

1. **Cascade repair**: sentinel logs from the original incident — did `+fix-slave-config` fire? Too slowly (10s INFO period × N) or not at all (replica unknown to sentinel)? If "too slow", is the correct fix tuning rather than a parallel repair loop?
2. **Dual-master demote (KB-10196)**: was the rogue master registered with Sentinel as a replica at the time? `+convert-to-slave` only covers masters Sentinel knows; a bootstrap-created stranger is invisible to it — which would point back at bootstrap, not at a missing demote loop.
3. **Full-sync stall (Bug 5)**: this one has no Sentinel-native equivalent (a stalled slave still answers PING); likely the only loop that must stay. Confirm the diskless-sync race is still reproducible on current valkey 8.x/9.x, and whether upstream has a fix (then the loop becomes version-gated).

[...]

PR #2992, opened 2026-07-02
chore(valkey): correct self-heal daemon justification with measured evidence

weicao:
Refs #2991 (evidence transcript posted there: three-scenario lab on valkey 8.1.3 — cascade repair, known rogue demotion, sentinel-blind rogue).

Comment-only change. Outcome of the investigation westonnnn asked about: the daemon **stays** — its two contested loops are the only mechanism for Sentinel-blind instances (Test C: a master that never attached to the monitored master receives no +convert-to-slave, structurally) — but its written justification was factually wrong for Sentinel-known instances and is now corrected with the measured numbers. The full-sync stall loop is documented as having no Sentinel-native equivalent.

Validation: bash -n + self-heal shellspec green; zero behavior change (comments only).


leon-ape (approved):
LGTM

PR #2993, opened 2026-07-02
fix(kafka): reject unsupported combined KRaft scale-out

weicao:
## Problem

Task #20 N8 v4 hit the same `ops-scale-out` path after helper-level topic readiness fixes. The frozen evidence package shows that `Cluster`/`OpsRequest` can report ready/succeeded while combined KRaft quorum/topic metadata is still not usable for topic roundtrip:

- evidence: `/evidence/kafka-release-full-main-n8-v4-20260702-204855/diagnosis-index/`
- broker-0 bootstrap can describe `kb-kafka-scale-out`, while broker-1/2 report the topic does not exist
- metadata quorum transitions later, but per-broker topic metadata still splits
- produce returns rc=0, but there is no durable offset and consume reads 0 messages

Current README contract exposed combined-mode scale-out as supported. For Kafka 3.3.2 combined KRaft, this addon does not implement online quorum voter add/convergence semantics, so allowing the OpsRequest to succeed creates a false supported path.

## Change

- Add an explicit `lifecycleActions.memberJoin` guard to `cmpd-combine.yaml` so combined-mode scale-out fails early with a clear unsupported reason.
- Update README scale-out examples to route online scale-out to separated topology broker-only components.
- Update the skip-immutable-check comment to cover both `memberJoin` and `memberLeave` hscale blockers.

## Validation

- `helm dependency build addons/kafka`
- `helm lint addons/kafka`
- `helm template kafka addons/kafka >/tmp/kafka-task21-render.yaml && rg -n "memberJoin|combined KRaft scale-out|memberLeave|combined KRaft scale-in" /tmp/kafka-task21-render.yaml`

## Boundary

This is a spec-boundary hardening PR, not a flaky-test workaround and not a release-ready claim. It prevents the known unsupported combined KRaft hscale path from being counted as a valid product/test path. If we later implement safe combined KRaft quorum membership convergence, this guard should be replaced by a real `memberJoin` implementation plus focused hscale validation.

### DIFF
--- a/addons/kafka/README.md
+++ b/addons/kafka/README.md
@@ -267,23 +267,25 @@ When the cluster creation is done, refer to a secret named `$(CLUSTER_NAME)$-kaf
 
 #### Scale-out
 
-Horizontal scaling out `kafka-combine` component in cluster `kafka-combined-cluster` by adding ONE more replica:
+> **Warning**: Combined KRaft mode (`kafka-combine`) does **not** support online horizontal scale-out for serviceVersion 3.3.2. Adding a combined node requires the Kafka controller quorum membership and per-broker metadata view to converge; this addon does not implement that workflow yet. The `memberJoin` action will reject combined-mode scale-out. Use separated topology and scale broker-only components when you need online broker scale-out.
+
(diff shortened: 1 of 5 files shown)


leon-ape (approved):
LGTM

PR #2997, opened 2026-07-03
fix(mariadb): fail closed on unsafe Galera bootstrap

### LINKED ISSUE #2996: fix(mariadb): fail closed on unsafe Galera crash recovery bootstrap
## Problem

The Galera startup script currently lets pod-0 rewrite `safe_to_bootstrap: 0` to `safe_to_bootstrap: 1` after `wsrep-recover` when no peers are alive.

That is unsafe after a full cluster crash because pod-0 can only recover its own local seqno. It cannot prove it is the latest node across peer PVCs. In a reproduced 3-node case, pod-0 recovered seqno 7 while peer PVCs had seqno 8 and 9, but pod-0 still bootstrapped the new cluster.

The observed run did not lose rows, so this issue is specifically about the bootstrap-election contract: an older pod must not be selected automatically when the latest node cannot be proven.

## Test scenario

1. Start a 3-node Galera cluster and write baseline rows.
2. Force-delete pod-0.
3. Write more rows through a surviving node.
4. Force-delete pod-1 and pod-2.
5. Let the cluster recover with all peers initially down.
6. Compare recovered seqnos from the peer data directories and inspect the startup logs.

Observed evidence from the reproduced case:

- pod-0 recovered seqno 7.
- pod-1 recovered seqno 8.
- pod-2 recovered seqno 9.
- pod-0 logged crash recovery bootstrap and rewrote `safe_to_bootstrap=1`.

## Expected behavior

If pod-0 has `grastate.dat` with `safe_to_bootstrap: 0` and no peer is alive, the script should:

- run `wsrep-recover` only to log local recovery evidence,
- keep `safe_to_bootstrap: 0` unchanged,
- refuse automatic bootstrap because the latest node cannot be proven locally,
- exit non-zero so the failure is visible instead of forming a cluster from an unproven node.

## Validation

Add focused shellspec coverage for both cases:

- unsafe pod-0 full-crash recovery refuses bootstrap and preserves `safe_to_bootstrap: 0`,
- already-safe pod-0 recovery with `safe_to_bootstrap: 1` still bootstraps.

weicao:
## Summary

- stop pod-0 from rewriting `safe_to_bootstrap: 0` to `safe_to_bootstrap: 1` during full Galera crash recovery
- log the local `wsrep-recover` seqno as evidence only
- fail closed when no peers are alive and the latest node cannot be proven locally
- add focused shellspec coverage for unsafe and already-safe pod-0 bootstrap paths

Closes #2996.

## Validation

- `shellspec --load-path ./shellspec --shell bash addons/mariadb/scripts-ut-spec/galera_start_spec.sh` -> 7 examples, 0 failures
- `bash -n addons/mariadb/scripts/galera-start.sh addons/mariadb/scripts-ut-spec/galera_start_spec.sh`
- `git diff --check -- addons/mariadb/scripts/galera-start.sh addons/mariadb/scripts-ut-spec/galera_start_spec.sh`

## Evidence boundary

This PR fixes the source-level unsafe bootstrap path and has focused unit coverage. It does not claim that the full Galera crash-recovery scenario is runtime-passed until the patched chart/image is validated in a live cluster.

### DIFF
--- a/addons/mariadb/scripts-ut-spec/galera_start_spec.sh
+++ b/addons/mariadb/scripts-ut-spec/galera_start_spec.sh
@@ -15,7 +15,7 @@ Describe "galera-start.sh"
   cleanup() {
     rm -rf "${TEST_DIR}"
     unset DATA_DIR POD_NAME MARIADB_ROOT_USER MARIADB_ROOT_PASSWORD PEER_FQDNS
-    unset GALERA_PRIMARY_PEER_WAIT_SECONDS
+    unset GALERA_PRIMARY_PEER_WAIT_SECONDS GALERA_BOOTSTRAP_DEFER_REASON
   }
   AfterEach "cleanup"
 
@@ -27,6 +27,14 @@ Describe "galera-start.sh"
 
   script_contains() {
     grep -F "$1" "$(script_file)"
+  }
+
+  write_grastate() {
+    cat > "${DATA_DIR}/grastate.dat" <<EOF
+uuid:    631a68d0-7697-11f1-923e-42be04dfa95f
+seqno:   ${1}
+safe_to_bootstrap: ${2}
+EOF
   }
 
   socketless_self_heal_kills_mariadbd() {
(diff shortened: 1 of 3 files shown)


leon-ape (commented):
With the new pod-0 fail-closed path, the existing non-pod-0 guard still ignores `safe_to_bootstrap: 1`. In a clean full shutdown where Galera marks a non-pod-0 as the safe node, pod-0 exits with the defer reason and that safe non-pod-0 also returns false to join, so no pod bootstraps the cluster.

weicao:
Addressed in `cdb834aec9d68b981a210b4a902356540ce14c87`.

Change made:
- a node with local `safe_to_bootstrap: 1` can bootstrap after the no-Primary-peer check, regardless of pod index
- `safe_to_bootstrap: 0` still stays fail-closed on pod-0 and no longer rewrites `grastate.dat`
- added shellspec coverage for the clean full-shutdown case where Galera marks a non-pod-0 as safe

Local validation on the rebased commit:
- `shellspec --load-path ./shellspec --shell bash addons/mariadb/scripts-ut-spec/galera_start_spec.sh` -> 8 examples, 0 failures
- `bash -n addons/mariadb/scripts/galera-start.sh addons/mariadb/scripts-ut-spec/galera_start_spec.sh`
- `git diff --check -- origin/main..HEAD`

weicao:
Addressed fresh Galera bootstrap regression in `3a4ae56a85ed10ddea4fe9dae13891b11c6cc224`.

Change made:
- treat `grastate.dat` with `seqno=-1` as fresh bootstrap state, not a clean-shutdown election result
- keep fresh bootstrap single-owner on pod-0 to prevent pod-0/1/2 from each forming their own Primary component
- preserve the previous clean full-shutdown behavior: a non-pod-0 node with `safe_to_bootstrap: 1` and a real seqno can still bootstrap after the no-Primary-peer check
- keep pod-0 with `safe_to_bootstrap: 0` fail-closed and non-mutating

Local validation:
- `shellspec --load-path ./shellspec --shell bash addons/mariadb/scripts-ut-spec/galera_start_spec.sh` -> 10 examples, 0 failures
- `bash -n addons/mariadb/scripts/galera-start.sh addons/mariadb/scripts-ut-spec/galera_start_spec.sh`
- `git diff --check -- origin/main..HEAD`

leon-ape (approved):
LGTM

PR #3003, opened 2026-07-03
fix(mongodb): KubeBlocks 1.1 compatibility patch for MongoDB sharding

gnolong:
Changes:
- Replace template functions in mongodb-shard-config.tpl and mongos-config.tpl.
- Remove hard-coded serviceVersion: 8.0.17 from sharding ComponentDefinitions so that Cluster serviceVersion 6.0.27 can match via ComponentVersion.

### DIFF
--- a/addons/mongodb/config/mongodb-shard-config.tpl
+++ b/addons/mongodb/config/mongodb-shard-config.tpl
@@ -2,16 +2,8 @@
 # for documentation of all options, see:
 #   http://docs.mongodb.org/manual/reference/configuration-options/
 
-{{- $log_root := getVolumePathByName ( index $.podSpec.containers 0 ) "log" }}
-{{- $mongodb_root := getVolumePathByName ( index $.podSpec.containers 0 ) "data" }}
-{{- $mongodb_port_info := getPortByName ( index $.podSpec.containers 0 ) "mongodb" }}
-{{- $phy_memory := getContainerMemory ( index $.podSpec.containers 0 ) }}
-
-# require port
-{{- $mongodb_port := 27017 }}
-{{- if $mongodb_port_info }}
-{{- $mongodb_port = $mongodb_port_info.containerPort }}
-{{- end }}
+{{- $mongodb_root := "/data/mongodb" }}
+{{- $mongodb_port := $.KB_SERVICE_PORT }}
 
 # where and how to store data.
 storage:
--- a/addons/mongodb/config/mongos-config.tpl
+++ b/addons/mongodb/config/mongos-config.tpl
@@ -2,14 +2,8 @@
 # for documentation of all options, see:
 #   http://docs.mongodb.org/manual/reference/configuration-options/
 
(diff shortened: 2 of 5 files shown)


leon-ape (commented):
`mongodb-shard-config.tpl` and `mongos-config.tpl` now hard-code `/data/mongodb`, while the chart still exposes `values.yaml:dataMountPath` and the sharding ComponentDefinitions mount the data volume from that value. With a non-default `dataMountPath`, the rendered volumeMount moves but dbPath/log/tmp/pid paths in the config templates stay under `/data/mongodb`, so sharding pods no longer use the configured data volume path.

weicao:
Design-contract review (classes 1–8 walked) against the release-1.1 contract.

**Verified — the change is precedent-aligned and internally consistent:**

- The replicaset `mongodb5.0-config.tpl` on release-1.1 already uses exactly this pattern (`$mongodb_root := "/data/mongodb"`, `$mongodb_port := $.KB_SERVICE_PORT`), and the replicaset `cmpd.yaml` already has no `serviceVersion`. This PR brings the sharding surfaces to the same, proven contract.
- All three touched cmpds define the vars the templates consume at this head: `KB_SERVICE_PORT` (shard, mongos, config-server), `MONGODB_CLUSTER_ROLE` (shard=`shardsvr`, config-server=`configsvr`), `CLUSTER_COMPONENT_NAME`. No `<no value>` render risk — class 2 checked.
- `cmpv-shard` covers `^mongo-shard-` + `^mongo-config-server-`, `cmpv-mongos` covers `^mongo-mongos-`; these match the `*.compDefName` helpers, and `6.0.27` / `8.0.17` are both active (non-deprecated) releases — so serviceVersion resolution via ComponentVersion works, and default resolution (latest = 8.0.17) is unchanged after removing the hard-coded cmpd `serviceVersion`.
- `values.yaml` `dataMountPath: /data/mongodb` matches the hard-coded path.
- Removed `$phy_memory` / `getContainerMemory` / `getPortByName` vars were defined-but-unused after the change — no dangling references at head.
- Classes 1, 3–8: clean (pure template/metadata change, no state or cleanup paths).

**Agree with the earlier review comment on hard-coding `/data/mongodb`:** the replicaset tpl carries the same `# TODO: .Values.dataMountPath` — so this PR is consistent with the existing debt rather than adding new debt. Fine to keep this compat fix minimal and fix the TODO across all mongodb tpls in a follow-up.

**Gaps to close before merge:**

1. No linked issue. Repo rule: every code PR needs an issue stating what it fixes, the test scenario, and how to reproduce (here: KB 1.1 + sharding topology + serviceVersion 6.0.27 fails to match / config render fails with podSpec-based template functions).
2. No runtime evidence on the PR: please attach a KB 1.1 render/create check — `helm template` plus creating a 6.0.27 sharding cluster showing the rendered `server.properties`-equivalent mongo config has the right port/path and the cluster reaches Running.

gnolong:
invalid modification

PR #3011, opened 2026-07-04
fix(postgresql): undefined leader_pod broke pending-restart watchdog

### LINKED ISSUE #3004: postgresql: undefined variable leader_pod disables pending-restart watchdog when leader is pending (postgres-setup.sh:67)
### Problem

`addons/postgresql/scripts/postgres-setup.sh` line 67 references `${leader_pod}`, which is never assigned anywhere in the script. The variable actually assigned (line 53) is `leader_pending_restart_pod`.

```bash
leader_pending_restart_pod=$(echo ${result} | jq -r ".members[] | select(.role == \"leader\" and .pending_restart == true) | .name")
...
if [[ "${pending_restart}" == "true" && (-z $leader_pending_restart_pod || "${leader_pod}" == "${CURRENT_POD_NAME}") ]]; then
```

Confirmed by shellcheck 0.11.0: `SC2154 (warning): leader_pod is referenced but not assigned.`

### Effect

The watchdog's intent (per the comment) is: "if leader pod is not pending_restart or current pod is leader pod, restart it". Because `${leader_pod}` expands to empty, whenever the leader itself is pending_restart, the second clause is always false, so **no pod (including the leader) is ever restarted by the watchdog** in that state. Restart-required parameter changes then never converge on members whose pending_restart coincides with the leader's.

Related minor bug in the same file, line 103: `restart_for_pending_restart_flag 2>&1 >> /home/postgres/.kb_set_up.log &` — redirect order is wrong (SC2069), stderr never reaches the log file.

Deterministic from the code path (static + shellcheck evidence; not yet reproduced at runtime). Runtime construction: 2-replica PG cluster; mark pending_restart on all members without going through update-parameter.sh's own `POST /restart` (e.g. `PATCH :8008/config` directly with a postmaster-context parameter); observe members stay pending_restart indefinitely.

Note: the repo's shellspec suite has zero coverage of `restart_for_pending_restart_flag` (9 existing examples cover 5 other functions), which is how this survived.

### Suggested fix

Use `leader_pending_restart_pod` in the comparison; fix the redirect to `>> log 2>&1`; add a shellspec spec for `restart_for_pending_restart_flag` covering leader-pending and leader-not-pending branches. I can submit the PR.

### Context

Found during the function-by-function review of the postgresql addon at main `12e01049` (requested by westonnnn, Slock #addon-reviewer4 task #1).

weicao:
## Problem

`postgres-setup.sh:67` referenced `${leader_pod}`, which is never assigned (the assigned variable is `leader_pending_restart_pod`, line 53). Whenever the Patroni leader itself is pending_restart, the watchdog condition is always false, so no pod — including the leader — is ever restarted by `restart_for_pending_restart_flag`. Restart-required parameter changes then fail to converge in that state.

Independent confirmation: shellcheck 0.11.0 SC2154 on the old line.

Details, effect analysis and runtime construction scenario: #3004

## Change

- Extract the restart decision into `need_restart_for_pending()` and use `leader_pending_restart_pod` (semantics per the original comment: restart when no leader is pending, or when the pending leader is the current pod — leader first, secondaries follow).
- Fix the watchdog background-job redirect `2>&1 >> log` → `>> log 2>&1` (SC2069; stderr previously never reached `.kb_set_up.log`).
- Add 4 shellspec cases for the decision function — this function previously had zero UT coverage, which is how the bug survived. Also updated one existing assertion in `regenerate_spilo_configuration_and_start_postgres` that had encoded the buggy redirect order (it asserted the redirect error appeared on stdout, which only happens with the wrong order).

## Validation

- `shellspec --shell bash5`: 13 examples, 0 failures (9 pre-existing + 4 new).
- `shellcheck -S warning postgres-setup.sh`: SC2154/SC2069 gone; no new findings.
- Boundary: static + UT validation only; no runtime cluster reproduction yet. The fix is a pure decision-logic correction with unchanged external behavior in the common path (leader not pending).

Origin: function-by-function review of the postgresql addon requested by westonnnn (Slock #addon-reviewer4 task #1).

Fixes #3004

### DIFF
--- a/addons/postgresql/scripts-ut-spec/postgres_setup_spec.sh
+++ b/addons/postgresql/scripts-ut-spec/postgres_setup_spec.sh
@@ -105,7 +105,7 @@ Describe "PostgreSQL Initialization Script Tests"
         :
       }
       When call regenerate_spilo_configuration_and_start_postgres
-      The stdout should include "/home/postgres/.kb_set_up.log: No such file or directory"
+      # with the ">> file 2>&1" redirect order, failed redirections report on stderr only
       The stderr should include "/home/postgres/.kb_set_up.log: No such file or directory"
       The status should be success
       The file "tmp_patroni.yaml" should be exist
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

PR #3020, opened 2026-07-06
fix(mariadb): escape backslashes before quotes in reconfigure SQL literals

### LINKED ISSUE #3019: mariadb: SQL injection in reconfigure — quote-only escaping defeated by trailing backslash under default sql_mode
## Problem (HIGH — SQL injection / privilege escalation)

The reconfigure lifecycle action embedded in `addons/mariadb/templates/_helpers.tpl` builds `SET GLOBAL \`<name>\` = '<value>';` for string-typed parameters. The value is escaped with single-quote doubling only:

```sh
escaped_value="$(printf "%s" "${param_value}" | sed "s/'/''/g")"
query="SET GLOBAL \`${param_name}\` = '${escaped_value}';"
```

Under MariaDB's **default** sql_mode there is no `NO_BACKSLASH_ESCAPES`, so backslash is an escape character inside string literals. A value ending in a backslash defeats the quote-doubling: `\'` escapes the first quote of the doubled `''` pair and the second quote closes the string early, so the remainder executes as SQL — run by `kb_internal_root` (ALL PRIVILEGES). Parameter values originate from `ComponentParameter`, i.e. a Kubernetes-level actor who may hold no DB credentials, so this is privilege escalation.

`is_safe_param_value` rejects control chars and `[section]` markers but **not** backslash or quote, so it does not close the gap; two of the four SQL-escape sites (the base reconfigure helper) have no value guard at all.

## Runtime evidence (docker, mariadb:11.4)

```
$ mariadb -e "SELECT @@sql_mode"
STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION   # no NO_BACKSLASH_ESCAPES
$ mariadb -e "SELECT 'x\''"    # -> returns  x'   (trailing quote consumed as data via the backslash-escaped quote)
```

Attack payload `v\'; CREATE USER pwned; -- ` with quote-only escaping breaks out; with the fixed escaping the injected `CREATE USER` does **not** run (verified: `mysql.user` count of `pwned` = 0).

## Fix

Escape backslashes **before** quotes at all four SQL-value escape sites (lines ~380/449/724/894):

```sh
escaped_value="$(printf "%s" "${param_value}" | sed -e 's/\\/\\\\/g' -e "s/'/''/g")"
```

This mirrors the escaping the my.cnf double-quoted path already uses. Verified against a real MariaDB with the injection payload above; normal values (`ON`, `3`, `volatile-lru`, `a b c`, values with apostrophes) round-trip unchanged.

1. Reconfigure a string parameter whose value ends in a backslash followed by `'; <sql>; -- `.
2. With the old escaping the injected `<sql>` executes as kb_internal_root; with the fix it is stored as literal data.

Shellspec Guard 3b added (executes the exact sed pipeline against the payload + regression guard that no `escaped_value` site uses quote-only doubling).

weicao:
Fixes #3019 — HIGH SQL-injection / privilege-escalation in the reconfigure action.

Root cause + runtime evidence (docker mariadb:11.4/11.8: default sql_mode has no NO_BACKSLASH_ESCAPES; `SELECT 'x\''` returns `x'`; injection payload's CREATE USER does not run after fix) are in the issue.

Change: all four SQL-value escape sites in _helpers.tpl now double backslashes before quotes. Validation: full mariadb shellspec suite green (653 examples, 0 failures, 9 pre-existing pendings), incl. new Guard 3b executing the exact sed pipeline against the breakout payload + a regression guard that no escaped_value site uses quote-only doubling; helm template renders. mariadb team full-suite verification per westonnnn.

### DIFF
--- a/addons/mariadb/scripts-ut-spec/reconfigure_persisted_alpha86_spec.sh
+++ b/addons/mariadb/scripts-ut-spec/reconfigure_persisted_alpha86_spec.sh
@@ -371,6 +371,53 @@ Describe "alpha.86 reconfigureAction.persisted semisync gates"
     End
   End
 
+  Describe "Guard 3b: SQL string-literal escaping is backslash-safe"
+    # is_safe_param_value rejects control chars / section headers but NOT
+    # backslash or quote, so the SQL `SET GLOBAL x = '<value>'` literal is
+    # the last line of defense. Under MariaDB default sql_mode (no
+    # NO_BACKSLASH_ESCAPES) backslash is an escape char, so doubling only
+    # single quotes (`sed "s/'/''/g"`) is defeated by a value ending in a
+    # backslash: `\'` escapes the first quote of the `''` pair and the
+    # second quote closes the string early, letting the remainder run as
+    # SQL as kb_internal_root. The escaping MUST double backslashes first,
+    # then quotes. These tests EXECUTE the exact sed pipeline extracted from
+    # the persisted helper against an injection payload (runtime-confirmed
+    # against mariadb:11.4/11.8: the injected statement does not execute).
+
+    escape_pipeline_from_helper() {
+      # Pull the first SQL-value sed pipeline out of the persisted helper.
+      grep -m1 'escaped_value=.*sed' "${HELPERS_TPL}"
+    }
+
+    It "the persisted helper doubles backslashes before quotes"
+      When call escape_pipeline_from_helper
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

PR #3023, opened 2026-07-06
fix(mariadb): start mariadbd with portable --read-only=ON, not enum NO_LOCK_NO_ADMIN

### LINKED ISSUE #3022: mariadb: --read-only=NO_LOCK_NO_ADMIN on mariadbd CLI silently degrades to OFF (fail-closed intent becomes fail-open)
## Problem (HIGH — fail-closed startup flag silently fail-open)

`replication-entrypoint.sh` launches mariadbd with `--read-only=NO_LOCK_NO_ADMIN` to start the server fenced (read-only) so there is no writable window before the SQL-based `set_fail_closed_read_only` runs.

But `NO_LOCK_NO_ADMIN` is a **runtime-variable enum value**, not a valid command-line boolean. `my_getopt` parses `--read-only` as a boolean and does not recognize the enum, so it warns and sets `read_only=OFF` — the exact opposite of the intent. This makes the fail-closed startup flag **fail-open on every shipped version**, leaving a writable window at startup before the first SQL fence.

## Runtime evidence (docker)

```
$ docker run --rm mariadb:11.4 mariadbd --read-only=NO_LOCK_NO_ADMIN --verbose --help 2>&1 | grep read_only
[Warning] option 'read_only': boolean value 'NO_LOCK_NO_ADMIN' wasn't recognized. Set to OFF.

$ docker run --rm mariadb:11.8 mariadbd --read-only=NO_LOCK_NO_ADMIN --verbose --help 2>&1 | grep read_only
[Warning] option 'read_only': boolean value 'NO_LOCK_NO_ADMIN' wasn't recognized. Set to OFF.
```

Both 11.4 (the addon default serviceVersion) and 11.8 reject it → OFF. It is NOT version-specific; it fails on the command line for every version. Contrast with the portable boolean:

```
$ docker run -d mariadb:11.4 --read-only=ON  ...  ->  SELECT @@global.read_only  ->  1   (fenced, no warning)
```

## Fix

Use the portable boolean on the command line:

```
--read-only=ON \
```

The SQL path (`set_fail_closed_read_only`) is unchanged and keeps its `NO_LOCK_NO_ADMIN → ON → 1` tiering, so the stronger runtime fence is still applied post-start where the engine supports it — and because mariadbd now genuinely starts read-only, a NO_LOCK_NO_ADMIN SQL failure on older versions no longer opens a write window either.

## Reproduce
1. Start a replication pod on mariadb:11.4; observe the startup warning and `SELECT @@global.read_only` = 0 during the pre-fence window.
2. With the fix, `@@global.read_only` = 1 from process start.

weicao:
Fixes #3022 — HIGH fail-closed-startup-flag-goes-fail-open.

Runtime evidence in the issue (11.4 AND 11.8 both: `boolean value 'NO_LOCK_NO_ADMIN' wasn't recognized. Set to OFF.`; `--read-only=ON` on 11.4 → @@global.read_only=1, no warning).

Change: one-line CLI value + explanatory comment; SQL fail-closed tiering unchanged. Validation: focused source-contract shellspec green (3 examples), helm renders. mariadb team full-suite verification per westonnnn.

### DIFF
--- a/addons/mariadb/scripts-ut-spec/replication_entrypoint_readonly_spec.sh
+++ b/addons/mariadb/scripts-ut-spec/replication_entrypoint_readonly_spec.sh
@@ -1,0 +1,47 @@
+# shellcheck shell=sh
+#
+# E1 contract: the mariadbd command line must start fail-closed with a
+# PORTABLE read_only value. NO_LOCK_NO_ADMIN is a runtime-variable enum,
+# not a valid command-line boolean — my_getopt does not recognize it and
+# silently sets read_only=OFF on every shipped version (verified on
+# mariadb 11.4 AND 11.8 via `mariadbd --read-only=NO_LOCK_NO_ADMIN
+# --verbose --help` → "boolean value 'NO_LOCK_NO_ADMIN' wasn't recognized.
+# Set to OFF."), turning fail-closed intent into a fail-open startup window.
+# The SQL path (set_fail_closed_read_only) keeps the NO_LOCK_NO_ADMIN →
+# ON → 1 tiering for the stronger runtime fence where the engine supports it.
+
+Describe "replication-entrypoint read_only fail-closed startup contract"
+  ENTRYPOINT="${SHELLSPEC_CWD:?}/addons/mariadb/scripts/replication-entrypoint.sh"
+
+  extract_mariadbd_launch() {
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

PR #3025, opened 2026-07-06
fix(postgresql): fail reconfigure action when patroni API rejects a parameter change

### LINKED ISSUE #3006: postgresql: reconfigure action ignores patroni API errors — failed parameter changes report success (update-parameter.sh)
### Problem

`addons/postgresql/reloader/update-parameter.sh` (the cmpd `reconfigure` action) applies parameter changes with unchecked curl calls:

```sh
curl -s -X PATCH -H "Content-Type: application/json" --data "$json_params" "http://localhost:8008/config"
...
curl -s -X POST "http://localhost:8008/restart"   # or /reload
```

`curl` exits 0 on HTTP 4xx/5xx unless `-f` is used, and the response body is discarded. If Patroni rejects the change (invalid value, bad request), the action still exits 0, so KubeBlocks marks the reconfigure as succeeded while the engine state diverges from desired. This is exactly the half-wired reconfigure path warned against in kubeblocks-addon-docs `docs/addon-api/08-parameters-and-config.md` (wrong desired state carried to new pods after restart/scale/upgrade).

Deterministic. Reconfigure with an invalid value (e.g. `max_connections=abc`): OpsRequest reports success, `pg_settings`/Patroni config unchanged.

### Suggested fix

Use `curl -fsS` plus explicit HTTP status/body check; exit non-zero with the Patroni error on stderr so the action fails visibly. Same treatment for the restart/reload POSTs.

### Context

Function-by-function review of postgresql addon at main `12e01049` (Slock #addon-reviewer4 task #1). Static evidence; not yet reproduced at runtime.

weicao:
## Problem

`reloader/update-parameter.sh` fires three curl calls (`PATCH /config`, then `POST /reload` or `POST /restart`) with `-s` and no status/response checking. When patroni rejects a parameter change (bad value, malformed request) or is unreachable, the script still exits 0 and the reconfigure action reports success while the parameter was never applied.

`docs/addon-api/08-parameters-and-config.md` calls out both sides of this: reconfigure scripts must not assume parameters are valid even after schema validation ("reconfigure 脚本不能假设参数一定合法……脚本仍要检查参数名、文件名、值格式和当前引擎状态"), and half-wired paths carry wrong desired state to new pods on restart/scale/upgrade.

Fixes #3006

## Changes

- All patroni REST calls go through `patroni_api()`: curl connection failures and non-2xx responses now fail the script (the cmpd reconfigure wrapper runs `set -eu`, so the action fails visibly), and the patroni response body is logged for diagnosis.
- `/reload` / `/restart` are skipped when the config `PATCH` already failed — no point reloading a config that was never accepted.
- Restructured into `update_parameter()` with the `__SOURCED__` guard (same pattern as orioledb / vanilla-postgresql reloaders) so the script is unit-testable. Parameter routing logic is unchanged.
- New `update_parameter_spec.sh` (9 examples): routing of PG GUC vs patroni DCS param vs restart-class param against the real yaml lists, plus error propagation (PATCH 4xx skips reload, curl exit 7, reload/restart 5xx). The script previously had zero spec coverage.
- `patroni-parameter.yaml` gained its missing trailing newline.

## Alternative considered

mariadb's reloader downgrades engine-rejected (permanently invalid) parameters to a loud `[REJECT]` + exit 0 to avoid endless retry of a bad value. I kept this PR strictly fail-loud because patroni's 4xx surface is narrow and the reconfigure action should reflect reality; happy to fold in the REJECT-classification pattern if reviewers prefer it.

## Evidence boundary

Static analysis + shellspec unit tests (22/22 green for the postgresql suite locally, shellcheck clean). Not reproduced on a live cluster.


leon-ape (approved):
LGTM

weicao:

PR #3028, opened 2026-07-06
fix(postgresql): pgdump/pgdumpall restore must fail on missing backup instead of reporting success

### LINKED ISSUE #3008: postgresql: pgdump/pgdumpall restore report success on missing backup file and swallow restore errors
### Problem

Restore scripts for the logical backup methods can report success without restoring anything:

1. **Missing backup file → silent success.** `dataprotection/pgdumpall-restore.sh` and `pgdump-restore.sh`: when `remote_file_exists` misses (backup object absent/renamed in the repo), the `if` block is skipped and the script falls through to implicit `exit 0`. The Restore CR completes successfully with an empty database.
2. **pgdump-restore swallows real errors.** No `set -e`; when `conflict_policy` is neither `DROP` nor `FAIL` (the default), the script runs `set +e`, ignoring every pg_restore failure; finally `grep "pg_restore: warning: errors ignored on restore"` forces `exit_code=0` even when objects failed to restore.
3. **pgdumpall-restore** pipes into `psql` without `ON_ERROR_STOP`, so SQL-level failures don't affect the exit code either.

Deterministic for (1): create a Restore referencing a Backup whose repo file was removed or renamed → Restore succeeds, no data. For (2): restore a dump into a target where some objects fail (e.g. missing role/extension) with default conflict_policy → success reported, objects missing.

### Suggested fix

- Explicit `echo "backup file not found: ..." >&2; exit 1` fallthrough in both scripts.
- Make `FAIL` the default conflict policy (or require explicit `IGNORE`); only downgrade a defined list of tolerated errors (e.g. role already exists for dumpall) instead of all.
- `psql --set ON_ERROR_STOP=on` for dumpall restore, with the same tolerated-error handling.

### Context

Function-by-function review of postgresql addon at main `12e01049` (Slock #addon-reviewer4 task #1). Static evidence; not yet reproduced at runtime.

weicao:
## Problem

Both Selective restore paths can report success without restoring anything (`docs/addon-api/09c-restore-and-rebuild.md`: restore phase success must not be claimed when the artifact/business data was never applied):

1. **pgdumpall-restore.sh** — when `${DP_BACKUP_NAME}.sql.zst` is absent from the repository, the only if-block is skipped and the script falls through to an implicit `exit 0`. A missing backup restores nothing, and the restore job reports success. This is a silent downgrade of restore semantics.
2. **pgdump-restore.sh** — no `set -e`, so a failed `datasafed pull` (missing/corrupt artifact) doesn't stop the script; and the `errors ignored on restore` grep forced `exit 0` under **every** conflict policy, including `FAIL`.
3. **actionset-pgdump.yaml** — `conflict_policy` is declared in `parametersSchema` (enum `CONTINUE/FAIL/DROP`, default `CONTINUE`) and the restore script branches on it, but it was never listed in `restore.withParameters`, so the env was never injected: the `DROP`/`FAIL` branches were dead code and every restore silently ran as `CONTINUE`.

Fixes #3008

## Changes

- Both restore scripts verify the backup artifact exists in the repository before pulling, and exit 1 with a clear error otherwise.
- `pgdump-restore.sh` runs under `set -e` for the setup phase; the `pg_restore` exit code is captured explicitly, and the ignored-errors → success downgrade now applies only to non-`FAIL` policies (`CONTINUE`'s continue-past-conflicts semantics are unchanged).
- `pg_restore` stderr now goes through a synchronous `tee` (PIPESTATUS) instead of an async `2> >(tee …)` process substitution — the log that the policy decision greps is guaranteed complete before it is read.
- `conflict_policy` added to `restore.withParameters`.
- shellspec coverage for both scripts (10 examples): missing artifact fails loudly and never reaches `pg_restore`/`psql`, success path, mid-stream pull failure, `pg_restore`/`psql` failures, and CONTINUE vs FAIL policy semantics. Neither script had any coverage before.

## Evidence boundary

Static analysis + shellspec unit tests (postgresql suite 23/23 green locally, shellcheck clean at CI severity). Not reproduced on a live cluster.

### DIFF
--- a/addons/postgresql/dataprotection/pgdump-restore.sh
+++ b/addons/postgresql/dataprotection/pgdump-restore.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 set -o pipefail
 export PATH="$PATH:$DP_DATASAFED_BIN_PATH"
 export DATASAFED_BACKEND_BASE_PATH="$DP_BACKUP_BASE_PATH"
@@ -8,9 +9,22 @@ psql_cmd="psql -h ${DP_DB_HOST} -U ${POSTGRES_USER} -p ${DP_DB_PORT}"
 mkdir -p $BACKUP_DIR
 trap "[ -d $BACKUP_DIR ] && rm -rf $BACKUP_DIR" EXIT
 
+function remote_file_exists() {
+    local out=$(datasafed list $1)
+    if [ "${out}" == "$1" ]; then
+        echo "true"
+        return
+    fi
+    echo "false"
+}
+
+if [ $(remote_file_exists "${DP_BACKUP_NAME}.tar") != "true" ]; then
+  echo "ERROR: backup file ${DP_BACKUP_NAME}.tar not found in the backup repository, cannot restore" >&2
+  exit 1
+fi
+
 datasafed pull "${DP_BACKUP_NAME}.tar" - | tar -xf - -C $BACKUP_DIR
 
-# Set default values
 # Set default values
 if [ -z "$jobs" ]; then
   jobs=4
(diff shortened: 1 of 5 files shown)


leon-ape (approved):
LGTM

weicao:

PR #3030, opened 2026-07-06
fix(mariadb): reconfigure fails when no parameter was applied

### LINKED ISSUE #3029: mariadb: reconfigure reports success when every parameter was engine-rejected (silent non-convergence)
## Problem (HIGH — reconfigure reports success when nothing was applied)

The reconfigure lifecycle action (both the base and persisted variants in `addons/mariadb/templates/_helpers.tpl`) classifies engine rejections (`ERROR 1231/1232/1193/1064` — bad value / wrong type / unknown variable / syntax) as "skipped" and swallows them:

```sh
case "${error_code}" in
  1231|1232|1193|1064)
    echo "[REJECT] parameter ... rejected by engine ..."
    skipped_count=$((skipped_count + 1))
    ;;
  *) exit 1 ;;
esac
```

The terminal gate only fails when **nothing at all** happened:

```sh
if [ "${applied_count}" -eq 0 ] && [ "${skipped_count}" -eq 0 ]; then
  exit 1
fi
```

So when **every** parameter is rejected (`applied=0, skipped>0`) the action exits 0. KB marks the Reconfiguring OpsRequest **Succeeded** and `ComponentParameter` status claims convergence, while zero declared values reached the engine — declared-value convergence is silently broken and the desired state stays divergent from runtime (carried into the next restart).

This is the same reject-as-success anti-pattern that `reloader/update-parameter.sh` carries; note `error_code` 1064 (SQL syntax) is also swallowed, which additionally masks the action's own SQL-construction bugs.

## Fix

Fail whenever `applied_count == 0`, regardless of `skipped_count` — an all-rejected run did not converge:

```sh
if [ "${applied_count}" -eq 0 ]; then
  if [ "${skipped_count}" -gt 0 ]; then
    echo "All ${skipped_count} parameter(s) were rejected by the engine; no declared value was applied — reconfigure did not converge" >&2
  else
    echo "No parameters were applied during reconfigure action" >&2
  fi
  exit 1
fi
```

Applied to both the base and persisted variants. Partial success (`applied>0` with some skips) still succeeds-with-warning — that is a separate policy question and is left unchanged to avoid forcing the controller into an infinite retry on a genuinely-bad single value.

## Test scenario / reproduce
1. Reconfigure with a single parameter the engine rejects (e.g. an unknown variable → 1193, or a bad value → 1231).
2. Old behavior: action exits 0, OpsRequest Succeeded, runtime unchanged.
3. Fixed: action exits 1 with "reconfigure did not converge".

Shellspec added: source-contract that the gate now keys on `applied_count == 0`, plus executable tests of the gate for (0,>0)→fail, (0,0)→fail, (>0,>0)→success.

weicao:
Fixes #3029 — HIGH silent non-convergence in the reconfigure action.

Root cause + reproduction in the issue. Change: both base and persisted gates in _helpers.tpl now fail whenever applied_count==0 (previously only when applied=0 AND skipped=0), so an all-rejected reconfigure no longer reports Succeeded. Partial success unchanged.

Validation: full mariadb shellspec suite green (655 examples, 0 failures, 9 pre-existing pendings), incl. new executable gate tests; helm renders. Note: touches a different region of _helpers.tpl than the H3 SQL-escaping PR (#3020), no conflict. mariadb team full-suite verification per westonnnn.

### DIFF
--- a/addons/mariadb/scripts-ut-spec/reconfigure_persisted_alpha86_spec.sh
+++ b/addons/mariadb/scripts-ut-spec/reconfigure_persisted_alpha86_spec.sh
@@ -539,23 +539,77 @@ Describe "alpha.86 reconfigureAction.persisted semisync gates"
   End
 
   Describe "Rejected user-input SQL errors do not keep reconfigure on the failing path"
-    It "base reconfigureAction skips classified engine rejects without exiting 1"
+    It "base reconfigureAction skips classified engine rejects without exiting 1 mid-loop"
       When call extract_base_helper_body
       The status should be success
       The output should include 'skipped_count=$((skipped_count + 1))'
       The output should include 'parameter(s) were rejected by engine and skipped'
-      The output should include 'if [ "${applied_count}" -eq 0 ] && [ "${skipped_count}" -eq 0 ]; then'
       The output should not include 'failing reconfigure to avoid accepting invalid rendered config'
     End
 
-    It "persisted reconfigureAction skips classified engine rejects without writing overrides or exiting 1"
+    It "persisted reconfigureAction skips classified engine rejects without writing overrides or exiting 1 mid-loop"
       When call extract_persisted_helper_body
       The status should be success
       The output should include 'skipped_count=$((skipped_count + 1))'
       The output should include 'continue'
       The output should include 'parameter(s) were rejected by engine and skipped'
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

PR #3035, opened 2026-07-06
fix(postgresql): verify patroni switchover result instead of assuming success

### LINKED ISSUE #3007: postgresql: switchover.sh does not verify patroni switchover result; silent no-op paths violate lifecycle contract
### Problem

`addons/postgresql/scripts/switchover.sh` never verifies the switchover result:

- Both `switchover_with_candidate` and `switchover_without_candidate` just `curl -s` the Patroni `/switchover` endpoint, echo the output, and leave `# TODO: check switchover result` (lines 32-47). Any Patroni error (no eligible candidate, lagging replica, API error) still ends with exit 0 → KubeBlocks proceeds as if the switchover succeeded.
- If the script's own primary detection disagrees with the target (`POSTGRES_PRIMARY_POD_NAME != CURRENT_POD_NAME`), it silently exits 0 instead of failing (line 60-63).
- Primary detection uses `.role == "leader"` only; in a standby cluster the Patroni role is `standby_leader`, so switchover there is a silent no-op.

The addon-api contract (`docs/addon-api/05b-lifecycle-action-fields.md`) explicitly requires the candidate-less path to self-confirm the switchover result, and `05c` requires role convergence evidence beyond exit 0.

Deterministic for the error-swallowing path: issue a switchover with a candidate that Patroni rejects (nonexistent or heavily lagging member) — Patroni returns an error, the action exits 0, roles unchanged, OpsRequest succeeds.

### Suggested fix

Parse the Patroni response; on accepted switchover, poll `/cluster` (bounded retry) until the leader equals the expected candidate (or differs from the old leader in the candidate-less case); exit non-zero on rejection/timeout. Treat primary-detection mismatch as an error. Handle `standby_leader`.

### Context

Function-by-function review of postgresql addon at main `12e01049` (Slock #addon-reviewer4 task #1). Static + doc-contract evidence; not yet reproduced at runtime.

weicao:
## Problem

`switchover.sh` fires the patroni `/switchover` request with `curl -s` and never checks the result — two in-code TODOs (`# TODO: check switchover result`) acknowledged this. Consequences:

- Connection errors and non-2xx responses were swallowed; the action exited 0 and KubeBlocks marked the switchover succeeded even when patroni rejected it (e.g. 503 "no good candidates").
- `docs/addon-api/05b-lifecycle-action-fields.md` (switchover row) requires both candidate paths to be handled and **the no-candidate path to verify the switchover outcome itself** — neither happened.
- The leader query matched only `role == "leader"`, so standby clusters (`standby_leader`) could never resolve a leader.
- When patroni was unreachable, the empty leader name fell into the "not primary" branch → silent `exit 0`.

Fixes #3007

## Changes

Follows the single-shot bounded-action pattern (kbagent clamps every action call at 60s):

- `request_switchover()` checks the HTTP status: 2xx proceeds; 5xx defers as transient (patroni 503 while replicas catch up); other codes fail hard; connection failures are classified, not swallowed.
- `verify_switchover()` positively confirms the outcome within a bounded window (4 × 2s): leadership must leave the old leader, and when a candidate was requested it must be the new leader. No unbounded loops.
- Every failure path emits classified stderr (`phase:` + `next-retry-safe: yes|no`) so operators can tell transient-converging from needs-attention.
- Leader resolution matches `leader` **and** `standby_leader`.
- Leader/label skew reports success only when the desired end state is positively observed (candidate already leader, or leadership already moved with no candidate); an unresolved leader is a classified failure instead of silent success.
- cmpd pins `timeoutSeconds: 50` on the switchover action with the budget math inline (worst case ~45s: leader query 5s + switchover API 20s + verification ~20s).

## Bounded-wait budget

| step | window |
|---|---|
| leader query | curl `-m 5` |
| switchover API | curl `-m 20` (patroni blocks until the attempt finishes) |
| verification | 4 × (curl `-m 3` + 2s sleep) ≈ 20s |
| **total worst case** | **~45s ≤ timeoutSeconds 50 < 60s clamp** |

## Tests

`switchover_spec.sh` rewritten — the old file had 1 live case plus ~120 lines of commented-out remnants of a pre-patroni script structure. Now 13 examples: success with/without candidate, standby_leader clusters, idempotent already-switched paths, 503 (retry-safe yes) / 412 (retry-safe no) / unreachable API, unconfirmed result within the bounded window, and leadership moving to a non-candidate pod. postgresql suite 25/25 green locally; shellcheck fully clean on the script.

## Evidence boundary

Static analysis + shellspec unit tests with a stateful curl mock. Not reproduced on a live cluster; the patroni response-shape assumptions (`/cluster` roles, `/switchover` status codes) follow the patroni REST API docs.

### DIFF
--- a/addons/postgresql/scripts-ut-spec/switchover_spec.sh
+++ b/addons/postgresql/scripts-ut-spec/switchover_spec.sh
@@ -24,121 +24,186 @@ Describe "PostgreSQL Switchover Script Tests"
   }
   AfterAll 'cleanup'
 
+  setup() {
+    tmpdir=$(mktemp -d -t pg-switchover-XXXXXX)
+    STATE_FILE="${tmpdir}/switchover-sent"
+    CLUSTER_JSON_BEFORE='{"members":[{"name":"pod-0","role":"leader"},{"name":"pod-1","role":"replica"}]}'
+    CLUSTER_JSON_AFTER='{"members":[{"name":"pod-0","role":"replica"},{"name":"pod-1","role":"leader"}]}'
+    SWITCHOVER_BODY="Successfully switched over"
+    SWITCHOVER_CODE=200
+    CURL_CLUSTER_EXIT=0
+    CURL_SWITCHOVER_EXIT=0
+    SWITCHOVER_VERIFY_ATTEMPTS=2
+    SWITCHOVER_VERIFY_INTERVAL=0
+    export STATE_FILE CLUSTER_JSON_BEFORE CLUSTER_JSON_AFTER SWITCHOVER_BODY \
+      SWITCHOVER_CODE CURL_CLUSTER_EXIT CURL_SWITCHOVER_EXIT \
+      SWITCHOVER_VERIFY_ATTEMPTS SWITCHOVER_VERIFY_INTERVAL
+    unset KB_SWITCHOVER_CANDIDATE_NAME 2>/dev/null || true
+  }
+
+  teardown() {
+    rm -rf "${tmpdir}"
+  }
+
+  BeforeEach 'setup'
+  AfterEach 'teardown'
+
+  # Emulates the two curl call shapes in switchover.sh:
+  #   GET  .../cluster    -> cluster topology JSON (switches once /switchover was called)
+  #   POST .../switchover -> "<body>\n<http_code>" (the -w "\n%{http_code}" format)
+  Mock curl
+    url=""
+    for arg; do url="$arg"; done
+    case "$url" in
+      */cluster)
+        if [ "${CURL_CLUSTER_EXIT:-0}" -ne 0 ]; then
+          exit "${CURL_CLUSTER_EXIT}"
+        fi
+        if [ -f "${STATE_FILE}" ]; then
+          printf '%s' "${CLUSTER_JSON_AFTER}"
+        else
+          printf '%s' "${CLUSTER_JSON_BEFORE}"
+        fi
+        ;;
+      */switchover)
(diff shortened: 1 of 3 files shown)


leon-ape (dismissed):
LGTM

wangyelei:
do not verify the result  when switchover without candicate

weicao:
@wangyelei Done — the no-candidate path no longer poll-verifies. It now relies on patroni's synchronous `/switchover` response: the checked 2xx is the completion signal (non-2xx and unreachable still fail the action), and role convergence is proven by the roleProbe afterwards. The candidate path keeps the bounded verification that the requested candidate actually became leader, since that's a request-specific expectation the HTTP response alone doesn't attest.

Spec updated accordingly (13 examples green): the no-candidate success case now asserts there is **no** poll verification even when the cluster view hasn't changed.

One note for the record: `docs/addon-api/05b-lifecycle-action-fields.md` (switchover row) says the no-candidate path should "自己确认切换结果" — I'm reading the checked synchronous 2xx as that confirmation. I'll record this interpretation in the addon-docs ambiguities file so the next addon doesn't re-litigate it.

wangyelei (dismissed):

weicao:
Note addressed (5230a3ef): verification-path leader queries now run at `curl -m 3` via a timeout arg (initial query keeps 5s), bringing the worst case to 45s — matching the documented budget and staying under `timeoutSeconds: 50` so the classified diagnosis can always emit. Switchover spec 13/13 (bash 5.3).

weicao:

PR #3036, opened 2026-07-06
fix(postgresql): remove ActionSet set -e that killed Continuous backup retry logic

### LINKED ISSUE #3010: postgresql: ActionSet-level set -e makes Continuous backup scripts' retry/error handling unreachable (crash-loop during failover)
### Problem

The DataProtection ActionSets prepend `set -e` (and `set -o pipefail`) when assembling Continuous-backup scripts via `Files.Get`:

- `templates/actionset-wal-g-pitr.yaml` → `dataprotection/wal-g-archive.sh`
- `templates/actionset-postgresql-pitr.yaml` → `dataprotection/postgresql-pitr-backup.sh`

Under `set -e`, the scripts' own error handling is dead code:

- `wal-g-archive.sh` `uploadMissingLogs`: the `exit_code=$?` failure branch (keep tracking file, retry after 180s, ~lines 178-196) is unreachable — a failing `wal-g wal-push` kills the shell before the check.
- Both scripts' `check_pg_process` 3-attempt probe loops: `is_secondary=$(psql ...)` is a plain assignment, so the first psql failure (e.g. during a switchover window) terminates the process instead of retrying; the designed "upload remaining WALs before exiting" path never runs either. The Continuous backup pod crash-loops through failover windows.

Caution for the fix: the same wrapper `set -e` is currently the only thing preventing a data-risk path in `postgresql-pitr-backup.sh` `upload_wal_log` (`datasafed push` then `.ready`→`.done` rename with no check between — without set -e a failed push would still mark the WAL as archived and PostgreSQL would recycle it). Any fix must add an explicit check there before/while touching the wrapper.

Methodology reference: kubeblocks-addon-docs `docs/addon-actionset-wrapper-set-e-vs-in-script-error-handling-guide.md` (error contract must be owned by exactly one layer; audit what set -e is blocking before changing it).

Run a wal-g-archive Continuous backup, then fail over the primary: the backup pod exits 1 on the first psql failure (no retry log lines) rather than entering its retry loop. Static evidence is deterministic from bash `set -e` semantics.

### Suggested fix

Choose the error-contract layer per script: for these long-running loop scripts, remove the wrapper `set -e` and convert every probe/upload into checked conditional execution (`if ! cmd`), including an explicit `|| exit` on the push-before-rename sequence; add assembled-form tests for the failure branches.

### Context

Function-by-function review of postgresql addon at main `12e01049` (Slock #addon-reviewer4 task #1). Static evidence; crash-loop behavior not yet reproduced at runtime.

weicao:
## Problem

Both Continuous backup ActionSets inject `set -e` ahead of long-running archiver scripts whose error handling is written for `set +e` semantics — so the error handling was dead code:

- **actionset-wal-g-pitr.yaml → wal-g-archive.sh**: `uploadMissingLogs`' tracking-file retry system (`eval wal-g wal-push; exit_code=$?` + the failure branch that preserves the tracking file for a delayed retry) was unreachable: the first failed wal-push killed the process. Same for `uploadDoneHistoryWALs` and for `check_pg_process`' bounded 3-probe loop, whose rescue path (upload remaining WALs before exit) never ran. During a failover window the Continuous backup pod crash-loops instead of retrying gracefully as designed.
- **actionset-postgresql-pitr.yaml → postgresql-pitr-backup.sh**: same pattern.

Fixes #3010. Methodology: [addon-actionset-wrapper-set-e-vs-in-script-error-handling-guide.md](https://github.com/apecloud/kubeblocks-addon-docs/blob/main/docs/addon-actionset-wrapper-set-e-vs-in-script-error-handling-guide.md) — the error contract must live in exactly one layer; for long-running archiver loops that layer is the script.

## set -e removal audit (per the guide's hard rule 4)

Before touching the wrappers I audited what `set -e` was actually blocking. One statement was load-bearing:

- `postgresql-pitr-backup.sh` `upload_wal_log`: `datasafed push` followed by an **unchecked** `mv .ready → .done`. `set -e` was the only thing preventing a failed push from being marked done — which would let PostgreSQL recycle a never-uploaded WAL segment and hole the PITR chain. That statement now has an explicit success gate (`if datasafed push …; then mv …; else keep .ready for retry`), landed **before** the wrapper change in the same PR.

Everything else the wrapper's `set -e` covered is either explicitly checked now (wal-g env bootstrap fails fast) or is designed failure-tolerance that the next loop round retries.

## Changes

- Both Continuous `backupData` wrappers: `set -e` removed, `set -o pipefail` kept, comment states the contract. The **restore-side** wrapper of actionset-postgresql-pitr keeps `set -e` — one-shot job, fail-fast is correct there (guide's decision rule 5).
- `postgresql-pitr-backup.sh`: rename-after-successful-push only; failed uploads logged and retried next round; `cd` / `pg_switch_wal` failures logged.
- `wal-g-archive.sh`: bootstrap (`mkdir`/`cp` of wal-g env) fails fast explicitly.
- New specs (11 examples) extract the archiver functions via awk shim (the scripts run infinite loops at top level, so `Include` is impossible) and pin the contract: failed upload keeps `.ready` + tracking file and is **never** marked `.done`; success renames and clears tracking; a fresh tracking file suppresses immediate retry; 3 failed probes trigger the rescue upload then exit 1.

## Evidence boundary

Static analysis + shellspec unit tests (postgresql suite 24/24 green locally, shellcheck clean at CI severity, helm template renders). The crash-loop-during-failover consequence is inferred from code paths, not reproduced on a live cluster; the unit tests pin the retry semantics that set -e previously made unreachable.

### DIFF
--- a/addons/postgresql/dataprotection/postgresql-pitr-backup.sh
+++ b/addons/postgresql/dataprotection/postgresql-pitr-backup.sh
@@ -1,3 +1,4 @@
+# shellcheck disable=SC2148
 export PGPASSWORD=${DP_DB_PASSWORD}
 # use datasafed and default config
 export WALG_DATASAFED_CONFIG=""
@@ -45,7 +46,7 @@ function switch_wal_log() {
     LAST_TRANS=$(pg_waldump $(${PSQL} -Atc "select pg_walfile_name(pg_current_wal_lsn())") --rmgr=Transaction 2>/dev/null |tail -n 1)
     if [ "${LAST_TRANS}" != "" ] && [ "$(find ${LOG_DIR}/archive_status/ -name '*.ready')" = "" ]; then
       DP_log "start to switch wal file"
-      ${PSQL} -c "select pg_switch_wal()"
+      ${PSQL} -c "select pg_switch_wal()" || DP_error_log "pg_switch_wal failed, will retry on the next round"
       for i in $(seq 1 60); do
         if [ "$(find ${LOG_DIR}/archive_status/ -name '*.ready')" != "" ]; then
           DP_log "switch wal file successfully"
@@ -60,7 +61,7 @@ function switch_wal_log() {
 # upload wal log
 function upload_wal_log() {
     local TODAY_INCR_LOG=$(date +%Y%m%d);
-    cd ${LOG_DIR}
+    cd ${LOG_DIR} || { DP_error_log "failed to cd to ${LOG_DIR}"; return 1; }
     for i in $(ls -tr ./archive_status/ | grep .ready); do
       wal_name=${i%.*}
       LOG_STOP_TIME=$(pg_waldump ${wal_name} --rmgr=Transaction 2>/dev/null | grep 'desc: COMMIT' |tail -n 1|awk -F ' COMMIT ' '{print $2}'|awk -F ';' '{print $1}')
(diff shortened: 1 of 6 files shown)


leon-ape (approved):
LGTM

weicao:

PR #3041, opened 2026-07-06
fix(mariadb): do not kill mariadbd during an in-progress Galera SST

### LINKED ISSUE #3040: mariadb: Galera socketless self-heal kills healthy SST longer than 90s (permanent non-convergence)
## Problem (HIGH — self-heal killer aborts healthy Galera SST)

`galera-start.sh` runs a self-heal watcher that kills mariadbd (to force a container restart) if the process runs **without a SQL socket** for `GALERA_SOCKETLESS_MARIADBD_THRESHOLD` × 3s (default 30 → 90s):

```sh
if pgrep -x mariadbd >/dev/null 2>&1 || pidof mariadbd >/dev/null 2>&1; then
  NO_SOCKET_COUNT=$((NO_SOCKET_COUNT + 1))
  if [ ${NO_SOCKET_COUNT} -ge ${NO_SOCKET_THRESHOLD} ]; then
    _restart_mariadbd_for_self_heal "mariadbd running without ${SOCK} for ..."
```

But a joiner performing an **SST (State Snapshot Transfer)** runs mariadbd **without creating the SQL socket until SST completes** — exactly the state this timer punishes. Any dataset whose rsync/mariabackup SST exceeds ~90s gets killed at 90s, restarted, and killed again → permanent non-convergence, and each aborted round re-blocks the donor. The addon's own memberJoin comment already notes "Galera SST can run much longer than 60s".

## Fix

Skip the socketless kill while an SST is in progress. Galera's own SST scripts create the marker `${DATA_DIR}/sst_in_progress` (defined in `wsrep_sst_common`: `SST_PROGRESS_FILE="$WSREP_SST_OPT_DATA/sst_in_progress"`) and run a `wsrep_sst_*` helper process for the transfer duration; either signal means SST is underway, not a stall:

```sh
elif [ -f "${DATA_DIR}/sst_in_progress" ] || pgrep -f 'wsrep_sst_' >/dev/null 2>&1; then
  NO_SOCKET_COUNT=0
else
  # existing socketless-kill path
```

The `NON_PRIMARY` killer (socket present, cluster non-Primary) is unchanged — a node with a working SQL socket that stays non-Primary is genuinely stuck. Only the socketless path (which SST legitimately produces) is guarded.

## Test scenario / reproduce
1. Join a Galera node whose SST (rsync of a dataset large enough) takes > 90s.
2. Old behavior: mariadbd is SIGTERM/SIGKILLed mid-SST at 90s and never converges.
3. Fixed: while `sst_in_progress` exists / the SST helper runs, the socketless counter is held at 0 and SST completes.

The marker name was confirmed from the mariadb:11.4 image: `grep sst_in_progress /usr/bin/wsrep_sst_common` → `SST_PROGRESS_FILE="$WSREP_SST_OPT_DATA/sst_in_progress"`.

Shellspec added: SST-guard present, and source-order assert that the guard precedes the socketless increment.

weicao:
Fixes #3040 — HIGH: the socketless self-heal watcher aborted any Galera SST longer than ~90s.

Root cause, the confirmed SST marker (wsrep_sst_common: sst_in_progress), and reproduction are in the issue. Change: skip the socketless kill while sst_in_progress exists or a wsrep_sst_ helper runs; NON_PRIMARY killer unchanged.

Validation: galera spec + full mariadb shellspec green (652 examples, 0 failures, 9 pre-existing pendings); helm renders. mariadb team full-suite verification per westonnnn.

### DIFF
--- a/addons/mariadb/scripts-ut-spec/galera_start_spec.sh
+++ b/addons/mariadb/scripts-ut-spec/galera_start_spec.sh
@@ -175,5 +175,28 @@ EOF
       The status should be success
       The output should include "GALERA_SOCKETLESS_MARIADBD_THRESHOLD"
     End
+
+    It "does not count socketless ticks while an SST is in progress"
+      # H9: a joiner performing SST runs mariadbd without a SQL socket until
+      # the transfer completes; the socketless killer must skip it, keyed on
+      # the Galera SST marker (sst_in_progress) or a live wsrep_sst_ helper.
+      When call script_contains "sst_in_progress"
+      The status should be success
+      The output should include "sst_in_progress"
+    End
+
+    It "gates the SST skip BEFORE the socketless-kill increment (source order)"
+      # The sst_in_progress guard branch must precede the NO_SOCKET_COUNT
+      # increment so an in-progress SST resets the counter instead of
+      # accumulating toward a kill.
+      When run sh -c '
+        f="$(printf "%s/addons/mariadb/scripts/galera-start.sh" "'"${SHELLSPEC_CWD:?}"'")"
+        sst=$(grep -n "sst_in_progress" "$f" | head -1 | cut -d: -f1)
+        inc=$(grep -n "NO_SOCKET_COUNT=\$((NO_SOCKET_COUNT + 1))" "$f" | head -1 | cut -d: -f1)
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

PR #3038, opened 2026-07-06
feat(valkey): cluster topology phase C — roles, roleProbe, switchover, member join/leave

weicao:
Fixes #3037. **Stacked on #3032 (phase B)** — base is `athena/valkey-cluster-phase-b`; retarget as the stack merges.

Contracts pinned by behavioral specs:
- roleProbe emits a SINGLE token (`primary` / `secondary`) read from the CLUSTER NODES `myself` line; unknown ⇒ non-zero skip, never an empty role (review class 4). The earlier `<role> <config-epoch>` versioned output was withdrawn in review: a node-local config epoch cannot prove a complete role fact, so shipping it would imply an anti-replay guarantee the engine view does not provide.
- switchover = `CLUSTER FAILOVER` on the replica (engine-native consistency); deterministic candidate pick; explicit candidate validated against the in-shard roster; bounded confirm with classified failure
- memberLeave of the current master: failover first, delete only after mastership provably moved — no orphaned slots; join/leave confirmed by cluster-view readback (slave flag + parent id binding), never by command rc alone

Validation: shellspec-test CI green at current head (12 new examples in this phase); renders pass. Runtime pass (role labels on 3 shards, master-kill label convergence, targeted switchover with writes, replica scale 2→3→2) is Emily's phase-C matrix.

Reviewers: @Alice / @Bob.

PR #3044, opened 2026-07-06
feat(valkey): cluster topology phase D — per-shard backup/restore with shard-count guard

### LINKED ISSUE #3042: valkey: cluster topology phase D — per-shard backup/restore with shard-count guard
**修什么**：cluster 拓扑（#3021 系列）还没有备份恢复。本 issue 交付 per-shard 备份与恢复，v1 恢复边界 = 同分片数。

**范围（PR-D）**：BPT 覆盖 valkey-cluster compDef（datafile 复用 #2984 快照口径：BGSAVE 快照 + users.acl，不打包 AOF/nodes.conf——nodes.conf 是集群身份，恢复必须重新生成；target 每分片 secondary fallback primary）；backup.sh cluster 感知（备份时从引擎 CLUSTER NODES 数 master 行得源分片数，写归档内 cluster-meta——引擎真值不依赖注入变量）；restore.sh cluster 感知（归档含 cluster-meta 且 restore-env 给了 RESTORE_TARGET_SHARDS 时不一致即 prepareData 阶段 fail-fast；未给时打显著 WARNING）。多余目标分片天然 fail-closed（无归档可拉）。恢复矩阵写 BPT 注释：v1 仅同分片数 + 位置映射，跨分片数/拓扑明确拒绝（错位后果 = key 落在不拥有其 slot 的分片上静默不可访问）。

**验收**：3 分片写跨 slot 数据 → 写负载中备份（沿用 #2984 区间断言）→ 恢复到新 3 分片集群全量读回 + ACL 存活；恢复到 4 分片（RESTORE_TARGET_SHARDS=4）被拒。

**平台边界**：DP 任务拿不到分片数变量（kubeblocks#10540 同族），自动强制的诚实上限 = 归档内引擎真值 + 可选 env 比对 + 文档。

**可复现**：确定性——当前 BPT 不匹配 valkey-cluster。

weicao:
Fixes #3042. **Stacked on #3038 (phase C)**.

Key contracts:
- per-shard datafile backup = #2984 snapshot semantics (dump.rdb + users.acl); nodes.conf deliberately excluded (cluster identity, regenerated on restore)
- backup embeds `cluster-meta` as engine truth for a FUTURE slot-aware restore design: `source_shards` counted from CLUSTER NODES master lines, plus `shard_master_id` and `shard_slot_ranges` resolved from the target's MASTER line via the myself-parent id (the backup target is a secondary and owns no slots itself); empty metadata ⇒ fail-fast; the file never remains in DATA_DIR (cleanup chained into the existing exit trap)
- **v1 restore stance: any archive containing cluster-meta is refused outright before any pod starts** (`refuse_cluster_restore`, exit 1). Same shard count is NOT a sufficient safety condition — post-rebalance slot layouts differ from a re-formed even split, so keys could land on shards that do not own their slots and become silently unreachable; refusal beats misplacement. The earlier `RESTORE_TARGET_SHARDS` same-shard-count gate was withdrawn in review for exactly this reason.
- platform boundary documented: DP jobs cannot see target sharding vars (kubeblocks#10540 family) — archive-embedded truth is the honest maximum
- behavioral spec harness runs backup.sh end-to-end against PATH-stubbed valkey-cli/datasafed and proves the archive contains cluster-meta with the master's ranges while DATA_DIR stays clean

Validation: valkey-scoped suite local **383 examples, 0 failures, 0 warnings** at head `d045e16e` (the two unconsumed-output warnings that failed shellspec-test are fixed — tar member list and BGSAVE log are now asserted); renders pass. Live pass (cross-slot write-load backup on 3 shards; restore-refusal verification) is Emily's phase-D matrix.

Reviewers: @Alice / @Bob.

### DIFF
--- a/addons/valkey/dataprotection/backup.sh
+++ b/addons/valkey/dataprotection/backup.sh
@@ -132,6 +132,22 @@ if [ ! -f "./dump.rdb" ]; then
 fi
 backup_files=("./dump.rdb")
 [ -f "./users.acl" ] && backup_files+=("./users.acl")
+
+# Cluster (sharding) mode: embed the source shard count as engine truth
+# (master lines in CLUSTER NODES) so restore can verify the same-shard-count
+# v1 boundary. Sentinel/standalone targets skip this (cluster_enabled:0).
+_cluster_enabled=$("${connect_base[@]}" INFO cluster 2>/dev/null | grep "^cluster_enabled:" | tr -d "\\r" | cut -d: -f2)
+if [ "${_cluster_enabled}" = "1" ]; then
+  _source_shards=$("${connect_base[@]}" CLUSTER NODES 2>/dev/null | tr -d "\r" | awk '$3 ~ /master/' | grep -c . || true)
+  if [ -z "${_source_shards}" ] || [ "${_source_shards}" -lt 1 ]; then
+    echo "ERROR: cluster mode detected but could not count source shards from CLUSTER NODES." >&2
+    exit 1
+  fi
+  printf 'source_shards=%s
+' "${_source_shards}" > ./cluster-meta
+  backup_files+=("./cluster-meta")
+  echo "INFO: cluster mode — embedded cluster-meta (source_shards=${_source_shards})."
+fi
 tar -cvf - "${backup_files[@]}" | datasafed push -z zstd-fastest - "${DP_BACKUP_NAME}.tar.zst" || exit 1
 
 save_sentinel_acl || \
--- a/addons/valkey/dataprotection/backup.sh
+++ b/addons/valkey/dataprotection/backup.sh
@@ -145,15 +145,33 @@ if [ "${_cluster_enabled}" = "1" ]; then
     echo "ERROR: cluster mode detected but could not count source shards from CLUSTER NODES." >&2
     exit 1
   fi
-  # Record THIS shard's current slot ranges too: shard count alone is not
-  # a sufficient restore precondition (post-rebalance layouts differ from
-  # the even split a re-formed cluster gets) — the ranges make the future
-  # slot-aware restore designable and keep archives self-describing.
-  _my_master_line=$("${connect_base[@]}" CLUSTER NODES 2>/dev/null | tr -d "\r" | awk '$3 ~ /myself/')
-  _my_slot_ranges=$(echo "${_my_master_line}" | cut -d' ' -f9- | tr ' ' ',')
+  # Record THIS SHARD's slot ranges + master identity: shard count alone
+  # is not a sufficient restore precondition (post-rebalance layouts
+  # differ from a re-formed even split). The backup target is normally a
+  # SECONDARY (BPT role) which owns no slots — so the ranges must come
+  # from this shard's MASTER line, resolved from the target's view:
+  # myself's parent id when the target is a slave, else myself itself.
+  _nodes=$("${connect_base[@]}" CLUSTER NODES 2>/dev/null | tr -d "\r")
+  _myself_line=$(echo "${_nodes}" | awk '$3 ~ /myself/')
+  if echo "${_myself_line}" | awk '{print $3}' | grep -q slave; then
+    _shard_master_id=$(echo "${_myself_line}" | awk '{print $4}')
+  else
+    _shard_master_id=$(echo "${_myself_line}" | awk '{print $1}')
+  fi
+  _master_line=$(echo "${_nodes}" | awk -v id="${_shard_master_id}" '$1 == id')
+  if [ -z "${_master_line}" ] || [ "${_shard_master_id}" = "-" ]; then
+    echo "ERROR: cluster mode — cannot resolve this shard's master from the target's CLUSTER NODES view." >&2
+    exit 1
+  fi
+  _shard_slot_ranges=$(echo "${_master_line}" | cut -d' ' -f9- | tr ' ' ',')
+  if [ -z "${_shard_slot_ranges}" ]; then
+    echo "ERROR: cluster mode — this shard's master ${_shard_master_id} owns no slot ranges; refusing to write empty metadata." >&2
+    exit 1
+  fi
   {
     printf 'source_shards=%s\n' "${_source_shards}"
-    printf 'shard_slot_ranges=%s\n' "${_my_slot_ranges}"
+    printf 'shard_master_id=%s\n' "${_shard_master_id}"
+    printf 'shard_slot_ranges=%s\n' "${_shard_slot_ranges}"
   } > ./cluster-meta
   backup_files+=("./cluster-meta")
   # cluster-meta is backup metadata, not engine data: it must not remain
--- a/addons/valkey/dataprotection/backup.sh
+++ b/addons/valkey/dataprotection/backup.sh
@@ -20,6 +20,8 @@ set -o pipefail
 
 function handle_exit() {
   local exit_code=$?
+  # cluster-mode metadata must never remain on the live data volume
+  [ "${_cluster_meta_cleanup_needed:-0}" = "1" ] && rm -f "${DATA_DIR}/cluster-meta"
   if [ "${exit_code}" -ne 0 ]; then
     echo "ERROR: backup failed with exit code ${exit_code}" >&2
     touch "${DP_BACKUP_INFO_FILE}.exit"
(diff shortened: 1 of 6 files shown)


weicao:
Focused restore/backup contract review for Phase D (`00ab9ec36c29ed255250a095a2e5369002a1e4b1`).

Evidence boundary: static review + Helm render + targeted shellspec only. I did not run DataProtection live restore. This PR is stacked on PR-B/PR-C, so their cluster lifecycle blockers still apply underneath it.

Findings:

1. Blocker: same shard count is necessary but not sufficient for cluster restore correctness; the PR does not preserve or re-create slot ownership per shard.

   Backup now writes only `source_shards=<N>` into `cluster-meta` (`addons/valkey/dataprotection/backup.sh:136`). Restore only checks `RESTORE_TARGET_SHARDS == source_shards` when the env is present (`addons/valkey/dataprotection/restore.sh:83`). The BPT comment says same shard count is supported by positional mapping (`addons/valkey/templates/backuppolicytemplate.yaml:80`).

   That does not prove restored keys land on the shard that owns their hash slots. Cluster backups are per-shard archives, but `nodes.conf` is intentionally not restored, so Phase B will create fresh cluster identity and fresh slot ownership. If the source cluster was ever rebalanced, scaled out/in, or otherwise had non-initial slot ranges, equal shard count alone does not reconstruct the source slot map. Even with the same shard count, archive A can be restored onto a shard that does not own archive A's slots, making keys silently unreachable.

   Please either:
   - embed and restore/validate the exact slot ranges owned by the backed-up shard, then make cluster formation assign the same ranges to the target shard before serving, or
   - fail-fast for cluster datafile restore until the PR can prove a stable source-shard -> target-shard -> slot-range mapping contract.

   A total shard-count guard alone is not enough to claim same-shard-count cluster restore support.

2. Blocker: the source shard identity / target shard identity is not validated.

   `cluster-meta` records only the cluster-wide shard count. It does not record which source shard this archive came from, its component short name, its master id, or its slot ranges. `restore.sh` has no target-shard identity input to compare against. This means two archives from the same backup set can be swapped between target shards and still pass the current guard.

   Please add a positive mapping proof. At minimum the backup metadata needs source shard identity plus slot ownership, and restore needs a target identity/slot expectation to compare. If DataProtection cannot inject the target shard identity today, the safe behavior is to refuse cluster restore or keep it explicitly unsupported, not warn-and-continue.

3. Needs stronger behavior tests for the backup side.

   The new backup test only greps for `cluster_enabled`, `source_shards=`, and `cluster-meta` (`addons/valkey/scripts-ut-spec/backup_contract_spec.sh:34`). It does not execute the cluster-meta path or prove the archive contains correct metadata. Please add a fake `valkey-cli`/`datasafed` behavior test that exercises `INFO cluster -> cluster_enabled:1`, `CLUSTER NODES`, the generated `cluster-meta`, and the tar file list.

Checks I ran:

- `shellspec --load-path ./shellspec --shell bash addons/valkey/scripts-ut-spec/restore_contract_spec.sh addons/valkey/scripts-ut-spec/backup_contract_spec.sh` -> 17 examples, 0 failures.
- `bash -n addons/valkey/dataprotection/backup.sh` and `restore.sh` -> pass.
- `helm dependency build addons/valkey` and `helm template valkey-addon addons/valkey` -> render includes `valkey-cluster-backup-policy-template` for `serviceKind: valkey-cluster`.
- PR body and Phase-D commit range are clean for generated-by/model/vendor attribution.

Design-contract walkthrough: checked classes 1-8. Class 2 is the main blocker: the restore success path does not enforce the required shard identity/slot ownership fields. Class 1 applies because the optional `RESTORE_TARGET_SHARDS` warning path can hide an unsupported cluster restore. Class 4 applies to `source_shards` being the only sentinel metadata for a richer mapping problem. Classes 3 and 5 inherit risk from fresh cluster re-formation after restore. Classes 6-8 did not show a separate blocker in this focused pass.

weicao:
Round-2 focused recheck for Phase D (`aa2a5d47106c35f7fcdd700d8e05fbce30ddc957`).

Evidence boundary: static review + Helm render + targeted ShellSpec only. I did not run DataProtection live backup/restore.

What is fixed from my previous review:

- Restore now fail-fast refuses any cluster archive in v1 instead of treating same shard count as supported. That closes the dangerous restore-success path.
- `cluster-meta` is removed from `DATA_DIR` after restore refusal.
- Backup cleanup uses the existing `EXIT` trap instead of adding a second trap that would clobber `handle_exit`.

Remaining blocker: the new `shard_slot_ranges` metadata is not a reliable shard slot-range record for the actual backup target.

The cluster BackupPolicyTemplate targets `role: secondary` with `fallbackRole: primary`. In `backup.sh`, the slot ranges are derived from this line:

```sh
_my_master_line=$("${connect_base[@]}" CLUSTER NODES 2>/dev/null | tr -d "\r" | awk '$3 ~ /myself/')
_my_slot_ranges=$(echo "${_my_master_line}" | cut -d' ' -f9- | tr ' ' ',')
```

When the backup target is the normal `secondary`, the `myself` line is a replica line and does not own slot ranges. The archive can therefore record `shard_slot_ranges=` even though the shard's current master owns real ranges. That is not the source shard slot ownership proof requested in round 1, and the current grep-style backup test would not catch it.

Please derive the slot ranges from the current master line for this shard, not from the backup pod's `myself` line when the pod may be a replica. The backup-side behavior test should execute the cluster path with a secondary target and assert the generated `cluster-meta` includes the expected source shard identity/slot ranges and that the file is cleaned from `DATA_DIR` after tar/failure.

There are two smaller consistency issues tied to the same point:

- The metadata still does not record a source shard identity, so it is not yet enough for a future source-shard -> target-shard mapping proof.
- `backup.sh` still has a stale comment saying the metadata supports a same-shard-count v1 boundary, while the implementation now correctly refuses cluster restore outright.

Checks I ran:

- `bash -n addons/valkey/dataprotection/backup.sh` and `restore.sh` -> pass.
- `shellspec --load-path shellspec --shell bash addons/valkey/scripts-ut-spec/restore_contract_spec.sh` -> 10 examples, 0 failures.
- `helm template valkey addons/valkey` -> pass.
- Round-2 commit range and PR body are clean for generated-by/model/vendor attribution.

weicao:
Round-3 focused review: still blocked.

1. CI is red at head `36066b57`: `shellspec-test` exits non-zero because the new `backup_cluster_meta_spec.sh` emits unconsumed output. The job summary shows 0 failures but 2 warnings, both from this new spec:
   - line 80 case: tar member list appears on stderr: `./dump.rdb`, `./users.acl`, `./cluster-meta`
   - line 91 case: backup INFO output appears on stdout and tar member list appears on stderr

Please either assert these outputs explicitly or suppress/capture them in the harness. A PR with shellspec warnings is not mergeable here because the workflow treats warnings as suite status failure.

2. The PR body still describes the old restore contract: `restore fail-fasts on mismatch when RESTORE_TARGET_SHARDS is set` and `same shard count only`. The code and BPT now refuse any cluster archive in v1, which is the right product boundary. Please update the PR body to say: cluster backup is supported per shard; cluster datafile restore is not supported in v1 and is refused at prepareData for any cluster archive.

Previous code blockers look closed in the static read: cluster-meta is cleaned from DATA_DIR; secondary backup target now records the shard master's slot ranges and `shard_master_id`; restore refuses cluster archives outright. XP classes 1-8 walked; no additional code-level blocker found beyond the CI harness warning and stale public contract text above.

weicao:
Focused re-review on head `40be33e9a70214eb17b9e2801998116737a2f535`: PASS for the blockers from my prior review.

Checked:
- CI is green for chart, shell-check, shellspec.
- PR body now states the v1 full-refusal restore stance and no longer describes the withdrawn `RESTORE_TARGET_SHARDS` / same-shard-count path.
- `restore.sh` refuses any archive containing `cluster-meta` before pod start; it removes `cluster-meta` and exits non-zero with an explicit unsupported-cluster-restore message.
- `backup.sh` records `source_shards`, `shard_master_id`, and `shard_slot_ranges` from the shard master line, so the normal secondary backup target no longer produces empty slot metadata.
- `cluster-meta` cleanup is preserved via immediate removal after tar and existing EXIT-trap cleanup on failure.
- Behavioral backup harness now runs backup against PATH-stubbed `valkey-cli`/`datasafed`, proves `cluster-meta` is inside the archive, proves master-line slot ranges are recorded, and proves `DATA_DIR` is clean afterward.
- The shellspec warning blocker is closed: tar member stderr and BGSAVE/embed stdout are now asserted.
- Local checks: `git diff --check origin/main..HEAD`, `bash -n` on dataprotection scripts, and targeted shellspec subset including `backup_cluster_meta_spec.sh` / `restore_contract_spec.sh` passed: 48 examples, 0 failures.
- PR body and commit bodies are clean for public attribution issues.

XP design-contract classes 1-8 walked: class 1 silent safety overclaim is closed by refusing all cluster restores in v1; class 2 metadata non-empty is enforced before archive write; class 5 cleanup is covered by immediate cleanup plus the existing EXIT trap; class 4/6/7/8 have no new blocker in this reviewed surface.

Evidence boundary: static/spec pass. Live phase-D evidence is still needed for real 3-shard backup completion and restore-refusal behavior in the cluster stack.

weicao:
Consolidated into #3064 (single draft PR per review-process decision; converts to ready-for-review after the full live gate passes). All review conclusions from this thread are incorporated at the consolidated head — same tree as the final phase-E stack head 7b27e5fc. Review history stays here for traceability.

PR #3049, opened 2026-07-06
fix(postgresql): make reload_patroni_config retry reachable, check PATCH results, fail closed on missing PG_MODE

### LINKED ISSUE #3047: postgresql: reload_patroni_config.sh retry loop is dead under set -e, PATCH errors unchecked, and empty PG_MODE silently clears standby config
`scripts/reload_patroni_config.sh` (used by the pg-update-standby-config OpsDefinition path) has three related defects:

1. **Retry loop is dead code.** The script runs under `set -e`; `result=$(curl -s …/config)` is a plain assignment, so the first curl failure kills the whole script — the 5-attempt/10s retry loop and its `$? -eq 0` check are unreachable. (Same class as the actionset set -e issue #3010.)
2. **PATCH results unchecked.** Both `curl -X PATCH …/config` calls use `-s` without `-f` or status inspection: an HTTP 4xx/5xx from patroni exits 0 and the script prints '==== Reload patroni config done ===='. (Same class as #3006.)
3. **Empty PG_MODE fails open.** `is_standby` is derived from `PG_MODE` containing 'standby'; when PG_MODE is unset (it is injected only by the cluster chart — a Cluster CR created directly never gets it), the script takes the non-standby branch and **PATCHes `standby_cluster: null`, silently destroying the standby configuration** of a standby cluster.

**Test scenario / reproduction**: static analysis (bash -x under stubbed curl shows the loop body executing at most once and exit-on-first-failure); for (3), run the ops on a Cluster CR created without the cluster chart → PG_MODE empty → standby config cleared. Reproducible: always.

**Fix direction**: make the retry condition-context (`if result=$(curl -s -f …)`), check PATCH with `-f` + explicit error, and fail closed with a clear error when PG_MODE is unset.

Origin: deep-review findings 一.9 and 一.8(3) (2026-07-03, main@12e01049); priority approved by weston.

weicao:
## Problem

Three related defects in `scripts/reload_patroni_config.sh` (the pg-update-standby-config path) — details and reproduction in #3047:

1. The 5-attempt retry loop was **dead code**: under `set -e`, `result=$(curl …)` kills the script on the first failure, so the `$?` check and retry branch never ran (same class as #3010's finding).
2. Both `PATCH /config` calls ran `curl -s` with no `-f`/status check: a patroni rejection still printed '==== Reload patroni config done ====' and exited 0 (same class as #3006).
3. **PG_MODE failed open**: when unset (only the cluster chart injects it — a directly-created Cluster CR never gets it), the script took the non-standby branch and PATCHed `standby_cluster: null`, silently destroying a standby cluster's replication config.

Fixes #3047 (deep-review findings 一.9 + 一.8(3), priority approved by weston)

## Changes

- Config GET moved into an if-condition so the bounded retry actually retries (set -e-safe); retry count/interval env-tunable.
- PATCH calls go through `patch_patroni_config()`: `curl -f` + classified stderr on failure; 'done' banner can no longer print over a rejected change.
- Unset PG_MODE is now a hard error naming the risk, instead of a silent standby-config wipe. (Set-but-non-standby values, e.g. `primary`, keep today's clear-standby behavior.)
- `__SOURCED__` guard added; new spec — 7 examples: fail-closed PG_MODE, primary clear path, standby patch path, invalid endpoint format, GET retry recovery, GET exhausting retries, rejected PATCH.

## Evidence boundary

Static analysis + shellspec (postgresql suite 20/20 green locally, shellcheck clean at CI severity). Not reproduced on a live cluster; the PG_MODE fail-open consequence follows from the code path and the cluster-chart-only injection point (report 一.8).

### DIFF
--- a/addons/postgresql/scripts-ut-spec/reload_patroni_config_spec.sh
+++ b/addons/postgresql/scripts-ut-spec/reload_patroni_config_spec.sh
@@ -1,0 +1,133 @@
+# shellcheck shell=sh
+
+Describe "scripts/reload_patroni_config.sh"
+
+  script_path() {
+    printf "%s" "../scripts/reload_patroni_config.sh"
+  }
+
+  setup() {
+    tmpdir=$(mktemp -d -t pg-reload-config-XXXXXX)
+    bindir="${tmpdir}/bin"
+    mkdir -p "${bindir}"
+    PATH="${bindir}:${PATH}"
+    CALL_LOG="${tmpdir}/calls.log"
+    : > "${CALL_LOG}"
+    RELOAD_CONFIG_MAX_RETRIES=2
+    RELOAD_CONFIG_RETRY_INTERVAL=0
+    CURRENT_POD_IP="127.0.0.1"
+    export PATH CALL_LOG RELOAD_CONFIG_MAX_RETRIES RELOAD_CONFIG_RETRY_INTERVAL CURRENT_POD_IP
+    unset PG_MODE primaryEndpoint CURL_GET_EXIT CURL_PATCH_EXIT 2>/dev/null || true
+    write_curl_stub
+  }
+
+  cleanup() {
+    rm -rf "${tmpdir}"
+  }
+
+  BeforeEach 'setup'
+  AfterEach 'cleanup'
+
+  # Emulates the two curl call shapes: GET /config returns the current config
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

Issue #3052, opened 2026-07-06
postgresql: two eval injection surfaces — accountProvision statement expansion and wal-g env command prefix


weicao:
Two places pass credential-bearing strings through `eval`:

1. **cmpd.yaml accountProvision**: `eval statement=\"${KB_ACCOUNT_STATEMENT}\"` expands the raw statement (with `${KB_ACCOUNT_NAME}`/`${KB_ACCOUNT_PASSWORD}` placeholders) through a full shell parse. A password containing shell metacharacters (`$`, backtick, `\"`) is executed, not substituted. System accounts are generated with `numSymbols: 0` so the platform path is safe today, but user-supplied secrets for the same accounts flow through the same eval.
2. **wal-g-archive.sh** (`uploadDoneHistoryWALs`, `uploadMissingLogs`): builds `env_cmd_prefix=\"env NAME=\"value\"...\"` from the wal-g env dir files — including `DATASAFED_ENCRYPTION_PASS_PHRASE` — and `eval`s it. Values traverse a shell parse (metacharacter breakage/injection) and land on a visible command line, while the addon already uses `envdir` for the exact same env directory in five other call sites (archive_command in backuppolicytemplate.yaml, wal-g-wal-archive.sh, wal-g-wal-restore.sh, both restore scripts).

**Test scenario / reproduction**: static; for (1) a password like `x\";touch /tmp/pwned;\"` demonstrates command execution through the eval (reproducible always at shell level). For (2) `ps`/proc argv shows the passphrase during wal-push under the eval form.

**Fix**: (1) eval-free placeholder substitution via bash `${var//pattern/replacement}` (replacement is literal — same contract mariadb-galera implements with printf|sed); (2) replace the eval prefix with `envdir \"${VOLUME_DATA_DIR}/wal-g/env\" wal-g …`, matching the addon's own established pattern.

Origin: deep-review finding 二.15 (2026-07-03, main@12e01049); priority approved by weston.

PR #3054, opened 2026-07-06
fix(postgresql): remove eval from accountProvision and wal-g upload paths

weicao:
## Problem

Two credential-bearing strings go through `eval` (deep-review finding 二.15, issue #3052):

1. **accountProvision** (`cmpd.yaml`): `eval statement=\"${KB_ACCOUNT_STATEMENT}\"` expands the account-statement placeholders through a full shell parse — metacharacters in a password **execute** instead of substituting. The platform's generated passwords (`numSymbols: 0`) can't trigger it, but user-supplied secrets for the same accounts flow through the same eval.
2. **wal-g-archive.sh** (`uploadDoneHistoryWALs` / `uploadMissingLogs`): builds an `env NAME=\"value\"…` prefix from the wal-g env dir — including `DATASAFED_ENCRYPTION_PASS_PHRASE` — and `eval`s it: values get re-parsed and the passphrase lands on a visible command line.

## Changes

- accountProvision expands placeholders with bash pattern substitution (`${var//pat/replacement}`) — the replacement is **literal**, no shell parse. Same contract mariadb-galera implements with escaped `printf | sed`. Verified locally: a password of `x\";touch /tmp/pwned;\"$(id)` remains a literal string in the resulting SQL (no execution).
- wal-g uploads switch to `envdir \"${VOLUME_DATA_DIR}/wal-g/env\" wal-g wal-push …` — the exact pattern this addon already uses in **five** other call sites (the BPT `archive_command` itself, wal-g-wal-archive.sh, wal-g-wal-restore.sh, both archive-restore scripts), which also proves `envdir` availability in the image. No values on argv, no re-parse.

## Boundary notes

- SQL-literal quoting is a separate layer: generated passwords contain no quotes (`numSymbols: 0`), so this PR fixes the shell layer only.
- **Interaction with #3036** (approved, unmerged): it touches other parts of wal-g-archive.sh and adds `walg_archive_spec.sh`; the diffs don't overlap so both merge cleanly, but the spec's wal-g stub needs a one-line `envdir` stub once both are in — I'll follow up on whichever lands second.

## Evidence

Static + rendered chart (block renders verbatim) + local substitution proof; shellcheck clean at CI severity; `bash -n` passes. No runtime cluster test — behavior change is confined to how the identical command lines are constructed.

### DIFF
--- a/addons/postgresql/dataprotection/wal-g-archive.sh
+++ b/addons/postgresql/dataprotection/wal-g-archive.sh
@@ -113,14 +113,6 @@ function save_backup_status() {
 function uploadDoneHistoryWALs() {
   DP_log "Checking for historical .history files with .done status to upload..."
 
-  # Build environment variables command prefix without affecting global env
-  env_cmd_prefix="env"
-  while read -r env_file; do
-    env_name=$(basename "$env_file")
-    env_value=$(cat "$env_file")
-    env_cmd_prefix="$env_cmd_prefix $env_name=\"$env_value\""
-  done < <(find "${VOLUME_DATA_DIR}/wal-g/env" -type f)
-
   # Find all .done status files for .history files in archive_status directory
   for done_file in $(find "${LOG_DIR}/archive_status/" -name "*.history.done" -type f | sort); do
     history_name=$(basename "$done_file" .done)
@@ -128,8 +120,10 @@ function uploadDoneHistoryWALs() {
 
     # Check if the actual .history file still exists
     if [ -f "$history_path" ]; then
-      # Try upload with WAL-G
-      eval ${env_cmd_prefix} ${VOLUME_DATA_DIR}/wal-g/wal-g wal-push "${history_path}"
+      # Try upload with WAL-G. envdir reads the env files directly: no shell
+      # re-parse of values, no passphrase on the command line (same pattern as
+      # the archive_command in backuppolicytemplate.yaml).
+      envdir "${VOLUME_DATA_DIR}/wal-g/env" "${VOLUME_DATA_DIR}/wal-g/wal-g" wal-push "${history_path}"
       exit_code=$?
       if [ "$exit_code" -eq 0 ]; then
         DP_log "Successfully uploaded file: ${history_name}"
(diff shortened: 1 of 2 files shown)


leon-ape (dismissed):
LGTM

weicao:
## Integrated-tree deep review — REQUEST CHANGES on the accountProvision half (wal-g half is READY)

**wal-g half**: the `envdir` invocation is correct, exit-code capture survives the (post-#3036) no-`set -e` wrappers, and it matches the proven `archive_command` pattern. One already-assigned note: `walg_archive_spec.sh` needs an `envdir` stub for hosts without daemontools.

**accountProvision half — bash ≥5.2 silently corrupts passwords containing `&` or `\`.**

`cmpd.yaml:265-266` replaces the `eval` with pattern substitution and comments that "bash pattern substitution treats the replacement as a literal". That is true only on bash ≤5.1. Bash 5.2 enabled `patsub_replacement` by default: in an unquoted replacement, `&` expands to the matched pattern and `\` escapes. Verified repro (bash 5.3 vs 3.2, same script):

```
password: a&b\c
bash 5.3:  CREATE USER u1 PASSWORD 'a${KB_ACCOUNT_PASSWORD}b\c';   <-- silently wrong
bash 3.2:  CREATE USER u1 PASSWORD 'a&b\c';                        <-- literal, correct
```

The statement is valid SQL, so the account is created with the wrong password and the KB Secret no longer authenticates — surfacing later as opaque auth failures, not at provision time. Default generated passwords are safe (`numSymbols: 0`), but the contract explicitly allows user-supplied passwords with symbols (07-accounts-and-tls.md:65-68). Exposure depends on the spilo image's bash; recent Ubuntu-based images ship 5.2+.

**Fix (one line, version-proof), before the substitutions:**

```bash
shopt -u patsub_replacement 2>/dev/null || true
```

**Strongly recommended companion (closes breakage + stderr leak + injection in one line):** pre-escape single quotes in both values before substitution — `KB_ACCOUNT_PASSWORD=${KB_ACCOUNT_PASSWORD//\'/\'\'}` — since a password containing `'` currently breaks the SQL, gets echoed (with the password fragment) in psql's `LINE 1:` error context, and a crafted secretRef password can execute arbitrary SQL in the superuser session (07-accounts-and-tls.md:55 requires the escape boundary + no-password-in-output).

weicao:
Fix pushed (ac90c6b1) per the review: `shopt -u patsub_replacement` before the substitutions (no-op on old bash; verified 5.3 no longer rewrites `&`/`\\`), plus SQL single-quote escaping via a portable quote-variable form (the backslash-escaped replacement itself misbehaves on bash 3.2 — verified both produce `''` correctly now). Closes corruption, error-context leak, and the injection path in one block. Integrated-tree suite 93/93.

weicao:

PR #3057, opened 2026-07-06
fix(mariadb): stop persisting wsrep_sst_auth root credential to the data volume

### LINKED ISSUE #3056: mariadb: Galera writes plaintext root credential (wsrep_sst_auth) onto the needSnapshot data volume — unused under rsync SST
## Problem (HIGH — plaintext root credential leaked into volume snapshots)

`galera-start.sh` wrote the root credential to a file on the data volume:

```sh
local sst_conf="${DATA_DIR}/.galera-sst-auth.cnf"
printf '[mysqld]\nwsrep_sst_auth=%s:%s\n' "${MARIADB_ROOT_USER}" "${MARIADB_ROOT_PASSWORD}" > "$sst_conf"
...
--defaults-extra-file=${sst_conf}
```

Two problems:

1. **Unused.** `wsrep_sst_auth` is only consumed by the mariabackup / xtrabackup / mysqldump SST methods. This addon uses `wsrep_sst_method = rsync` (`config/mariadb-galera.tpl`), which does **not** authenticate via `wsrep_sst_auth`. So the credential served no purpose.
2. **Credential leak into backups.** `${DATA_DIR}` is the `data` volume, declared `needSnapshot: true` in `cmpd-galera.yaml`. The plaintext root user:password therefore rode into every volume snapshot / backup of a Galera cluster — readable by anyone who can read a snapshot.

## Fix

Do not write `wsrep_sst_auth` at all, and drop the `--defaults-extra-file` that loaded it. Also remove any stale `.galera-sst-auth.cnf` left on existing PVs by a previous chart version (so upgrades scrub the already-leaked file):

```sh
rm -f "${DATA_DIR}/.galera-sst-auth.cnf" 2>/dev/null || true
# wsrep_args no longer includes --defaults-extra-file=${sst_conf}
```

If a future change switches the SST method to mariabackup, `wsrep_sst_auth` should be provided through a non-snapshot path (e.g. a mounted Secret / conf.d outside the data volume), not written into `${DATA_DIR}`.

## Verification
- `helm template` renders; galera-start.sh no longer writes `wsrep_sst_auth` or loads it via `--defaults-extra-file`.
- Full mariadb shellspec suite green (653 examples, 0 failures); new contract spec asserts no auth write, no defaults-extra-file for SST, and the stale-file cleanup.

Note: an existing snapshot taken before this fix still contains the old credential; rotating the root password after upgrade is advisable.

weicao:
Fixes #3056 — HIGH: plaintext root credential leaked into Galera volume snapshots.

Root cause in the issue: wsrep_sst_auth (unused under rsync SST) was written to ${DATA_DIR} (needSnapshot:true), so it rode into every snapshot. Change: stop writing it, drop --defaults-extra-file, and scrub any stale file on upgrade.

Validation: helm renders; full mariadb shellspec green (653 examples, 0 failures, 9 pre-existing pendings) incl. new contract spec. Note: snapshots taken before this fix still contain the old credential; root password rotation after upgrade advised. mariadb team full-suite verification per westonnnn.

## Current-head stale-credential removal successor

Current addon exact: `71174128cf814db6178719587966ecfe7b6c01df` (tree `e247719931092625a9fe60e9a702e5e83ae5b115`). This successor makes the upgrade scrub required rather than best-effort: if the stale SST credential path exists, deletion must succeed and an absence readback must confirm it is gone. A delete error or false-success leaves startup at `rc=1` with clear stderr instead of starting with the credential still on the snapshot-backed volume.

- old-source delete-error / false-success cases: `3` failures
- current focused Galera-start spec: `18` examples, `0` failures
- `git diff --check`, Bash syntax, and ShellCheck error-level: pass
- snapshots created before this fix may still contain the credential; rotation guidance remains unchanged
- no current-exact product runtime was run; runtime remains `N=0`

### DIFF
--- a/addons/mariadb/scripts-ut-spec/galera_start_spec.sh
+++ b/addons/mariadb/scripts-ut-spec/galera_start_spec.sh
@@ -199,4 +199,27 @@ EOF
       The output should equal "OK"
     End
   End
+
+  Describe "wsrep_sst_auth is not persisted (H11)"
+    # rsync SST does not use wsrep_sst_auth; writing it to DATA_DIR (a
+    # needSnapshot volume) leaked the plaintext root password into snapshots.
+
+    It "does not write wsrep_sst_auth to any file"
+      When run sh -c "grep -nE 'wsrep_sst_auth=.*>|> +\"?\\\$?\\{?sst_conf' \"$(script_file)\" || true"
+      The status should be success
+      The output should equal ""
+    End
+
+    It "does not load a defaults-extra-file for SST auth"
+      When run sh -c "grep -F 'defaults-extra-file=' \"$(script_file)\" | grep -F 'sst' || true"
+      The status should be success
+      The output should equal ""
+    End
+
+    It "removes any stale .galera-sst-auth.cnf left by a previous chart version"
+      When call script_contains "rm -f \"\${DATA_DIR}/.galera-sst-auth.cnf\""
+      The status should be success
+      The output should include ".galera-sst-auth.cnf"
+    End
+  End
 End
--- a/addons/mariadb/scripts-ut-spec/galera_start_spec.sh
+++ b/addons/mariadb/scripts-ut-spec/galera_start_spec.sh
@@ -221,5 +221,37 @@ EOF
       The status should be success
       The output should include ".galera-sst-auth.cnf"
     End
+
+    It "deletes an existing stale credential and verifies absence"
+      printf '%s\n' "wsrep_sst_auth=root:secret" > "${DATA_DIR}/.galera-sst-auth.cnf"
+
+      When call remove_stale_galera_sst_auth
+      The status should be success
+      The path "${DATA_DIR}/.galera-sst-auth.cnf" should not be exist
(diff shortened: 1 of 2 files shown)


leon-ape (dismissed):
LGTM

weicao:
rebase 到 main 554fd848 + re-author 为 Magnus <magnus@apecloud.com>。语义复核 vs main #3041（同文件 galera-start.sh 的 SST 保护）：sst_in_progress 守卫与本 PR 的 sst-auth 不落盘共存无冲突，全量 ShellSpec 666/0/9p。@leon-ape 的 approve 因 force-push 需重新确认。

weicao:
Superseded by #3276. This PR's current head 29a025d8c5aa7e9d65b8253403d307ec8558c5e7 is included in the consolidated candidate 7978812cf8ca066cfc13ac6baf0ddd78505e0965; further review and testing should use #3276.

weicao:

leon-ape (approved):
LGTM

Issue #3058, opened 2026-07-06
postgresql: naming-convention hardcodes break when the component is not named 'postgresql'


weicao:
Two "works only under default naming" hardcodes (deep-review finding 一.8; the PG_MODE sub-item was already fixed fail-closed in #3047/#3049):

1. **wal-g-backup.sh:12,94 / wal-g-incremental-backup.sh:112** assemble the primary Service hostname as `${CLUSTER_COMPONENT_NAME}-${COMPONENT_NAME}` — i.e. `<cluster>-<comp>-<comp short name>`. The actual Service is `<cluster>-<comp>-<serviceName>` and the cmpd pins `serviceName: postgresql`, so the concatenation is only accidentally correct when the component short name equals "postgresql". These probes must hit the **primary** (`pg_switch_wal()` cannot run on a standby), and the BPT target has `fallbackRole: secondary`, so `DP_DB_HOST` is not a safe substitute — the role-selecting Service is the right mechanism, with the fixed serviceName suffix.
2. **pg14/15/16-config.tpl** resolve serviceVersion by matching `$spec.name == "postgresql"` literally, silently falling back to the template default (14.0/15.0/16.0). Consequence: a differently-named component silently renders the older `shared_preload_libraries` list (loses pg_duckdb at >=14.18/15.13/16.9). Topology-mode clusters always match; direct-componentDef clusters with custom names never do.

**Test scenario / reproduction**: static; a componentSpec named `pg` with `componentDef: postgresql-16-…` → (1) assembled host `<cluster>-pg-pg` matches no Service; (2) rendered postgresql.conf carries the pre-16.9 preload list regardless of serviceVersion. Reproducible: always.

**Fix**: (1) `${CLUSTER_COMPONENT_NAME}-postgresql` (fixed serviceName); (2) extend the match to accept `componentDef` prefix "postgresql", keeping the per-major default only for the legitimate omitted-serviceVersion case.

Origin: deep-review finding 一.8 (2026-07-03, main@12e01049); priority approved by weston.

PR #3059, opened 2026-07-06
fix(postgresql): stop hardcoding the default component name in wal-g host and config version lookup

weicao:
## Problem

Deep-review finding 一.8 (issue #3058) — two "accidentally correct only under default naming" assumptions:

1. wal-g full/incremental backup scripts assemble the primary Service host as `${CLUSTER_COMPONENT_NAME}-${COMPONENT_NAME}`; the real Service is `<cluster>-<comp>-<serviceName>` with `serviceName: postgresql` pinned in the cmpd — so it only resolves when the component short name is "postgresql". `pg_switch_wal()` and the archive_command wait-loop must hit the **primary**; `DP_DB_HOST` is not a safe substitute because the BPT target has `fallbackRole: secondary`.
2. pg14/15/16 config templates find serviceVersion by matching `$spec.name == "postgresql"` literally and silently fall back to 14.0/15.0/16.0 — silently rendering the pre-pg_duckdb `shared_preload_libraries` list for any custom-named component.

## Changes

- Host suffix → the fixed serviceName: `${CLUSTER_COMPONENT_NAME}-postgresql` (comment explains the primary-only constraint).
- Version match → `or (eq "postgresql" $spec.name) (hasPrefix "postgresql" (default "" $spec.componentDef))`: topology-mode clusters keep matching by name (the topology names the component postgresql), direct-componentDef clusters with custom names now match by definition prefix. The per-major default stays only for the legitimate omitted-serviceVersion case.

## Boundary

Static change; chart renders. The config templates render in KB's config engine — `hasPrefix`/`default` are sprig functions, asserted available from the same templates' existing `semverCompare`/`div`/`min` usage; flagging for reviewer confirmation.

### DIFF
--- a/addons/postgresql/config/pg14-config.tpl
+++ b/addons/postgresql/config/pg14-config.tpl
@@ -29,7 +29,7 @@
 
 {{- $pgVersion := "14.0" }}
 {{- range $i, $spec := $.cluster.spec.componentSpecs }}
-{{- if eq "postgresql" $spec.name }}
+{{- if or (eq "postgresql" $spec.name) (hasPrefix "postgresql" (default "" $spec.componentDef)) }}
 {{- if $spec.serviceVersion }}
 {{- $pgVersion = $spec.serviceVersion }}
 {{- end }}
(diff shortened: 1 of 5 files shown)


leon-ape (approved):
LGTM

weicao:

Issue #3062, opened 2026-07-06
postgresql: static IMAGE_TAG fallback 14.8.0 makes version-mapping misses run backups with the wrong-major image


weicao:
Four ActionSets carry a static env default `IMAGE_TAG: 14.8.0-pgvector-v0.6.1`:

- actionset-postgresql-pitr.yaml:31
- actionset-pgbasebackup.yaml:13
- actionset-wal-g-pitr.yaml:23
- actionset-pgdumpall.yaml:11

Each corresponding BPT method carries an `IMAGE_TAG` `versionMapping` generated from the chart's declared versions, which overrides the static default when the target's serviceVersion matches. The static default therefore only ever takes effect on a **mapping miss** (serviceVersion not in the chart's version list — e.g. a custom ComponentVersion minor): a PG 16/17/18 cluster then silently runs its backup/restore tooling with the PG 14 image. Wrong-major `pg_dump`/`pg_basebackup`/`wal-g` against a newer server is exactly the failure class that should be loud, not silent.

(`wal-g` / `wal-g-incremental` methods are unaffected — their actionsets use the fixed walgImage helper, no IMAGE_TAG.)

**Test scenario / reproduction**: static; set a serviceVersion outside the chart's `.Values.versions` minors → the method env resolves nothing and the actionset default supplies 14.8.0. Reproducible: always.

**Fix**: remove the four static defaults. On a mapping miss the image renders with a literal `$(IMAGE_TAG)` tag and the job fails visibly at image resolution, naming the unexpanded variable — attributable fail-fast instead of a silent wrong-major run.

Origin: deep-review finding 三.19 (2026-07-03, main@12e01049); priority approved by weston.

PR #3063, opened 2026-07-06
fix(postgresql): remove static IMAGE_TAG fallback that hid version-mapping misses

weicao:
## Problem

Deep-review finding 三.19 (issue #3062): four ActionSets (postgresql-pitr, pgbasebackup, wal-g-pitr, pgdumpall) default `IMAGE_TAG` to `14.8.0-pgvector-v0.6.1`. The BPT's `versionMapping` overrides it whenever the target serviceVersion is in the chart's declared versions — so the static default only ever takes effect on a **mapping miss**, at which point a PG 16/17/18 cluster silently runs its backup/restore tooling with the PG 14 image. Wrong-major `pg_dump`/`pg_basebackup`/`wal-g` should fail loudly, not run quietly.

## Change

Remove the four static defaults (and the now-empty `env:` block in pgdumpall). On a mapping miss the image renders with a literal `$(IMAGE_TAG)` tag → the job fails at image resolution with the unexpanded variable in the error — attributable fail-fast. `wal-g`/`wal-g-incremental` methods are unaffected (fixed `walgImage` helper).

## Evidence boundary

Static + rendered chart: the only remaining `14.8.0-pgvector` occurrences are the 14.8.0 serviceVersion's own mapping entries and the pg14 cmpd/cmpv images (legitimate). Not runtime-tested; the change removes a fallback rather than altering any matched-path behavior.

### DIFF
--- a/addons/postgresql/templates/actionset-pgbasebackup.yaml
+++ b/addons/postgresql/templates/actionset-pgbasebackup.yaml
@@ -10,8 +10,6 @@ spec:
   env:
     - name: DATA_DIR
       value: {{ .Values.dataMountPath }}/pgroot/data
-    - name: IMAGE_TAG
-      value: 14.8.0-pgvector-v0.6.1
   backup:
     preBackup: []
     postBackup: []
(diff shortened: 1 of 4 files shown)


leon-ape (approved):
LGTM

weicao:

PR #3066, opened 2026-07-06
fix(mariadb): fail closed after uncertain DCS switchover to avoid split-brain

weicao:
## Summary (H2)

In `addons/mariadb/scripts/replication-switchover.sh`, the Stage-3 (DCS switchover) failure path calls `rollback_current_primary_switchover_guard`, which sets the current primary back to `read_only=0` and restores its remote-root write grants. This can produce a **split-brain with two writable primaries**.

## Root cause

`syncerctl_switchover` is invoked with `--force` and wrapped by `timeout(1)`:

```sh
output=$(timeout "${wall}" "${SYNCERCTL_BIN}" ... switchover --force --primary ... --candidate ...)
rc=$?
...
case "${rc}" in 124|125|137) ... return 1 ;; esac   # timeout -> fail-closed sentinel
```

A non-zero / timeout return does **not** prove the DCS switchover record was never persisted. Syncer can accept and persist the request, begin **asynchronously** promoting the candidate, and then have its client (`syncerctl`) killed by the wall-clock budget.

On that return, `run_switchover` does:

```sh
if ! syncerctl_switchover ...; then
  if switchover_final_state_already_reached ...; then return 0; fi
  rollback_current_primary_switchover_guard || true   # <-- sets old primary read_only=0
  return 1
fi
```

`switchover_final_state_already_reached` only detects the **fully-completed** end state (candidate already primary + current following it). During the in-progress window (syncer still promoting), it returns false, so rollback runs and un-fences the old primary. Under the alpha.79 minimalist model the old primary is still fully writable at this point (`prepare_current_primary_for_switchover` is a no-op; the Stage-4 fence has not run yet), so rollback actively guarantees it stays writable. When syncer then finishes promoting the candidate (also writable), **both nodes accept writes**.

Rollback-to-writable is only safe **before** `syncerctl` is ever invoked — i.e. the Stage-2 pre-DCS failure path (`wait_candidate_sql_reachable_before_dcs`), where no DCS record can exist. That call site is unchanged.

## Fix

First principles: once the DCS write cannot be proven to have not landed, fail closed. On the Stage-3 failure path, fence the current primary (`read_only=1`) instead of rolling it back to writable, via a new `fence_current_primary_after_uncertain_dcs` helper. If the DCS record truly did not persist, KB roleProbe + reconcile converge the still-read_only old primary (and the failed OpsRequest may retry the whole switchover) — a bounded, recoverable availability blip, always preferable to unbounded split-brain writes.

## Predecessor tests

`addons/mariadb/scripts-ut-spec/replication_switchover_spec.sh` updated: the two Stage-3-failure cases now assert the current primary is fenced (`set_read_only=ON`), never unfenced (`set_read_only=OFF`), and `rollback` is not called; plus a new case asserting a manual-verification error is surfaced when the fail-closed fence itself cannot be applied. Full mariadb suite: 664 examples, 0 failures.

## Current-head truthful fence-result successor

Current addon exact: `a4805b1d401f188a3fc1c0c47541b5f903f1f74d` (tree `2268c3b07f3647093eaef5fbfdd998c85c3b8e2e`). The uncertain-DCS fail-closed design above is unchanged; this successor fixes its result reporting. `fence_current_primary_after_uncertain_dcs` logs `fail_closed=true` only after the fence succeeds. If the fence fails, it logs `fail_closed=false` with manual-verification guidance and the action returns `1`; it never emits a false-success claim.

- old-source focused failure cases: `2` failures
- current focused successor: `4` examples, `0` failures
- current full switchover spec: `192` pass / `0` fail / `6` historical pending
- `git diff --check`, `sh -n`, and ShellCheck error-level: pass
- no current-exact product runtime was run; runtime remains `N=0`

### DIFF
--- a/addons/mariadb/scripts-ut-spec/replication_switchover_spec.sh
+++ b/addons/mariadb/scripts-ut-spec/replication_switchover_spec.sh
@@ -495,7 +495,12 @@ EOF
     End
 
     Context "when syncerctl cannot create DCS switchover"
-      It "returns failure"
+      # H2 split-brain fix: after syncerctl (invoked with --force under a
+      # timeout(1) wrapper) fails, the DCS record may already be persisted and
+      # syncer may be promoting the candidate. The current primary must be
+      # fenced fail-closed (read_only=1), NOT rolled back to writable — the old
+      # rollback path could leave two writable primaries.
+      It "fails closed by fencing the current primary (does not roll back to writable)"
         make_failing_syncerctl
         prepare_current_primary_for_switchover() {
           return 0
@@ -504,14 +509,24 @@ EOF
           record_call "rollback"
           return 0
         }
+        set_local_read_only() {
+          record_call "set_read_only=$1"
+          return 0
+        }
+        local_read_only_is() {
+          [ "$1" = "1" ]
+        }
         When call run_switchover "mdb-mariadb-1" "mdb-mariadb-1.mdb-mariadb-headless.demo.svc.cluster.local"
         The status should be failure
         The output should include "Switchover: creating syncer DCS switchover"
-        The stderr should include "Switchover failed: syncerctl could not create DCS switchover"
-        The contents of file "${TEST_DIR}/calls" should include "rollback"
+        The output should include "fencing current primary read_only=1"
+        The stderr should include "current primary fenced fail-closed"
+        The contents of file "${TEST_DIR}/calls" should include "set_read_only=ON"
+        The contents of file "${TEST_DIR}/calls" should not include "set_read_only=OFF"
+        The contents of file "${TEST_DIR}/calls" should not include "rollback"
       End
 
-      It "treats a zero-status syncerctl failure message as failure and rolls back"
+      It "treats a zero-status syncerctl failure message as failure and fails closed by fencing"
         make_zero_status_failing_syncerctl
         prepare_current_primary_for_switchover() {
           return 0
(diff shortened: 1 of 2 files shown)


weicao:
rebase 到 main 554fd848 + re-author 为 Magnus <magnus@apecloud.com> + 按 org 规范移除 AI 署名 trailer。内容零变化。

weicao:
Superseded by #3276. This PR's current head baf11e6f0f3f420a2d6c26021fcf695573e1c040 is included in the consolidated candidate 7978812cf8ca066cfc13ac6baf0ddd78505e0965; further review and testing should use #3276.

weicao:

weicao (commented):
The exact head is covered only by mocked uncertain-DCS tests; there is no live timeout-after-DCS-commit run proving the resulting old-primary fence. The split-brain safety path can therefore merge without demonstrating that an ambiguously committed transition cannot leave the old node writable.

weicao:
task #539 exact-head follow-up

- Current head: `a4805b1d401f188a3fc1c0c47541b5f903f1f74d`; tree: `2268c3b07f3647093eaef5fbfdd998c85c3b8e2e`.
- This head is an ancestor of candidate `7fa7cfe7009e4bf353582203b1a0089eb381428b`; fresh exact-head CI is SUCCESS and the combined MariaDB ShellSpec terminal is 931 pass / 0 fail / 9 historical pending.
- The review body names a runtime-evidence gap, not an unresolved source thread: the current source keeps the old primary fail-closed after an uncertain DCS result and reports fence failure truthfully. The live timeout-after-DCS-commit fence proof remains Test-owned and UNKNOWN for the current candidate.
- Fresh independent exact-head design/code re-review has been routed to Magnus. No code or PR-head mutation was made in this follow-up.

weicao:
Review-comment disposition at exact head `a4805b1d401f188a3fc1c0c47541b5f903f1f74d` (tree `2268c3b07f3647093eaef5fbfdd998c85c3b8e2e`):

The source path is fail-closed: after an uncertain `syncerctl_switchover` result it does not call the pre-DCS rollback-to-writable path; it attempts `read_only=1`, verifies the fence, and reports `fail_closed=false` if that fence cannot be proven.

Fresh focused verification: `replication_switchover_spec.sh` = 192 examples / 0 failures / 6 historical pendings (Bash 3.2). All exact-head GitHub checks are SUCCESS.

The review's decisive requirement is still open: there is no current-candidate live timeout-after-DCS-commit run proving the old primary remains fenced. That runtime evidence is Test-owned and remains UNKNOWN; mocked/source coverage is not presented as closure.

Issue #3071, opened 2026-07-06
postgresql: PITR restore_command consumes WAL segments with mv — a mid-recovery restart cannot replay; plus string timestamp comparison


weicao:
Two defects in the PITR restore path (deep-review finding 三.20):

1. **postgresql-pitr-restore.sh:30**: `restore_command='case "%f" in *history) cp …;; *) mv ${PITR_DIR}/%f %p ;; esac'` — WAL segments are **moved** out of the staging dir as PostgreSQL consumes them during recovery. If recovery restarts mid-way (pod eviction, OOM, node restart — routine in Kubernetes), PostgreSQL re-requests earlier segments through restore_command and they are gone: the restore can never complete and the PITR attempt is burned. The `.history` branch already uses `cp`, which concedes the point. PostgreSQL's own documentation examples use `cp` for restore_command precisely for replayability.
2. **postgresql-fetch-wal-log.sh**: `[[ $timestamp > $restore_time ]]` — `>` inside `[[ ]]` is lexicographic. Both operands are epoch seconds (currently 10 digits), so it is accidentally correct today; it is a latent trap for any change in operand shape.

**Test scenario / reproduction**: static for (2); for (1): start a PITR restore with a multi-segment WAL window, kill the pod after several segments have been consumed, let the Job/pod retry — recovery fails requesting an already-moved segment. Reproducible: deterministic once the restart lands mid-recovery.

**Fix**: restore_command uses `cp` for all files (disk note: fetch-wal-log already stages the whole window in PITR_DIR beforehand, so the peak is already paid; cp adds only transient duplication in pg_wal bounded by recovery checkpoint recycling). Timestamp comparison switches to `-gt`.

Origin: deep-review finding 三.20 (2026-07-03, main@12e01049); priority approved by weston.

PR #3072, opened 2026-07-06
fix(postgresql): make PITR restore_command replayable and timestamp comparison numeric

weicao:
## Problem

Deep-review finding 三.20 (issue #3071):

1. **restore_command consumed segments with `mv`**: PostgreSQL re-requests earlier WAL segments through restore_command whenever recovery restarts mid-way (pod eviction, OOM, node restart — routine on Kubernetes). With `mv`, those segments are gone from the staging dir and the restore can never complete — the PITR attempt is burned. The `.history` branch already used `cp`, conceding the point.
2. **Lexicographic timestamp comparison** in fetch-wal-log's stop condition (`[[ $ts > $rt ]]`): both operands are epoch seconds of currently-equal width, so it is accidentally correct today — a latent trap.

## Changes

- `restore_command='cp ${PITR_DIR}/%f %p'` for all files (PostgreSQL's own documented example shape). Disk boundary: fetch-wal-log stages the whole window up front, so the peak is already paid; cp adds only transient pg_wal duplication bounded by recovery checkpoint recycling.
- `-gt` for the epoch comparison.
- New spec (3 examples) that executes **the same concatenation the ActionSet builds** (set -e + common-scripts + fetch-wal-log + pitr-restore): pins the cp restore_command, the staging + idempotent-retry behavior, and the timestamp fix via a 9-digit/10-digit epoch pair that discriminates lexicographic from numeric comparison.

## Evidence boundary

Static + concatenated-form shellspec (suite 23/23 locally, shellcheck clean at CI severity). The mid-recovery-restart replay scenario is not runtime-reproduced; it will be covered in the draft-batch integration round (T3 PITR window) and/or the #postgresql team's kubeblocks-tests negative-path cases (task #29 alignment).

Opened as **draft** per the batch process (tests pass → ready). Will be added to integration/pg-review-batch.

### DIFF
--- a/addons/postgresql/dataprotection/postgresql-fetch-wal-log.sh
+++ b/addons/postgresql/dataprotection/postgresql-fetch-wal-log.sh
@@ -47,7 +47,7 @@ function fetch-wal-log(){
             continue
          fi
          timestamp=`date -d "$latest_commit_time" +%s`
-         if [[ $latest_commit_time != "" && $timestamp > $restore_time ]]; then
+         if [[ $latest_commit_time != "" && $timestamp -gt $restore_time ]]; then
             DP_log "exit when reaching the target time log."
             exit_fetch_wal=1
             break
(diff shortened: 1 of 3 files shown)


weicao:

PR #3084, opened 2026-07-06
fix(mysql): add innodb_redo_log_capacity to 8.0 cue schema (port from apecloud-addons)

### LINKED ISSUE #3083: mysql: innodb_redo_log_capacity missing from 8.0 cue schema (port enterprise fix)
## Problem

`addons/mysql/config/mysql8.0-config.tpl` renders `innodb_redo_log_capacity` into my.cnf (line 130) and `mysql8-config-effect-scope.yaml` lists it as a dynamic parameter (line 541), but `mysql8-config-constraint.cue` has no entry for it (only `innodb_redo_log_encrypt`). Reconfiguring this important modern parameter (MySQL 8.0.30+, supersedes innodb_log_file_size) is therefore either rejected by schema validation or passes without any range guard, depending on controller behavior.

The enterprise repo (apecloud-addons) already added the cue entry; this issue tracks porting it to kubeblocks-addons (requested by @westonnnn in the review thread, 2026-07-06).

Reproducible: yes.

1. Create a MySQL 8.0 cluster on kubeblocks-addons main.
2. Reconfigure `innodb_redo_log_capacity` (e.g. to 1073741824) via OpsRequest.
3. Observed: parameter is not present in the ParametersDefinition cue schema; validation outcome is not range-guarded.

## Expected

`innodb_redo_log_capacity` present in the cue schema with the same constraint as the enterprise repo (`int & >= 8388608`), so reconfigure is validated and accepted.

weicao:
Fixes #3083.

## Problem

The MySQL 8.0 configuration template and dynamic-effect list already expose `innodb_redo_log_capacity`, but the CUE schema omitted the field. That made the supported setting fail schema validation. The field must remain optional so existing configurations that do not set it keep rendering.

## Change

Add the schema entry at the matching position after `innodb_redo_log_encrypt`:

```cue
// The total capacity in bytes of the redo log.
innodb_redo_log_capacity?: int & >= 8388608
```

This preserves the minimum accepted value while avoiding a new required-field contract for every MySQL 8.0 configuration.

## Validation

- full chart render: PASS
- `git diff --check`: PASS
- PR CI on exact head: terminal green

## Evidence Boundary

This closes the schema contract and static rendering gates. Runtime reconfiguration remains N=0 and belongs to the test owner.

### DIFF
--- a/addons/mysql/config/mysql8-config-constraint.cue
+++ b/addons/mysql/config/mysql8-config-constraint.cue
@@ -580,6 +580,9 @@
 	// Controls encryption of redo log data for tables encrypted using the InnoDB tablespace encryption feature.
 	innodb_redo_log_encrypt: string & "OFF" | "ON" | *"OFF"
 
+	// The total capacity in bytes of the redo log.
+	innodb_redo_log_capacity: int & >= 8388608
+
 	// The replication thread delay (in ms) on a slave server if innodb_thread_concurrency is reached.
 	innodb_replication_delay?: int & >=0 & <=4294967295
 
--- a/addons/mysql/config/mysql8-config-constraint.cue
+++ b/addons/mysql/config/mysql8-config-constraint.cue
@@ -580,8 +580,8 @@
 	// Controls encryption of redo log data for tables encrypted using the InnoDB tablespace encryption feature.
 	innodb_redo_log_encrypt: string & "OFF" | "ON" | *"OFF"
 
-	// The total capacity in bytes of the redo log.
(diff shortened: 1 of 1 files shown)


weicao:
Absorbed the follow-up from PR #3160 at head `e59e2216`: `innodb_redo_log_capacity` remains type/range validated but is optional, matching the template behavior that omits it when no memory limit is available. Post-merge checks: `helm lint addons/mysql` PASS, `git diff --check` PASS, and the schema contains exactly one optional declaration. PR #3160 is superseded by this PR.

PR #3088, opened 2026-07-06
fix(mysql): make PITR disk-pressure binlog cleanup functional and prefix-safe

### LINKED ISSUE #3087: mysql: PITR disk-pressure binlog cleanup is a silent no-op (dead KB_ITS env) with a latent purge-gap defect
## Problem

The disk-pressure binlog cleanup in `addons/mysql/dataprotection/mysql-pitr-backup.sh` (enabled by default: `PURGE_BINLOG: "on"` in `actionset-pitr.yaml`, triggered at >=80% disk usage) is a silent no-op on KubeBlocks 1.x:

1. Replica discovery reads `env | grep "KB_ITS_.*_HOSTNAME"` — those env vars were only injected into workload pods by pre-1.0 KubeBlocks and never existed in a DataProtection job pod (the script's own TODO admits it). `REPLICA_HOSTS` is always empty, `get_synced_binlogs` returns 1, cleanup exits before purging anything. The binlog volume can fill the disk while the advertised protection does nothing.

2. Latent second defect in `purge_master_binlogs`: it iterates all files and issues `PURGE BINARY LOGS TO '<file>'` per qualifying file. `PURGE ... TO` is prefix deletion — if an earlier file is NOT uploaded but a later one is, purging to the later file also deletes the missing-from-repo earlier file, tearing a hole in the archived binlog sequence (PITR continuity loss). The "keep latest 5" slice also actually keeps 4 (`${all_binlogs[@]: -4}`).

3. Membership tests use substring `grep -q`, which mismatches at binlog number-width rollover (e.g. `bin.000001` matches `bin.0000010`).

Reproducible: yes (defect 1 deterministic).

1. Create a MySQL cluster with continuous backup (archive-binlog) enabled on KB 1.x.
2. `kubectl exec` into the archive-binlog job pod: `env | grep KB_ITS` → empty.
3. Fill the binlog dir past 80%: log shows "Executing cleanup_mysql_binlogs..." followed by "No synced binlog files found" every cycle; no `PURGE BINARY LOGS` is ever executed on the server (verify with `SHOW BINARY LOGS` count growing).

## Expected

- Replica hosts are discovered from the target primary itself (`SHOW REPLICAS` / 5.7 `SHOW SLAVE HOSTS` — each replica registers its report_host there), which works in the DP job pod.
- Master purge issues a single prefix-safe `PURGE BINARY LOGS TO '<first file that must be kept>'`, stopping at the first not-synced-or-not-uploaded file; newest 5 always kept.
- Exact-match membership tests.
- Behavior boundary unchanged: if no replica can be confirmed (connection failure or zero registered replicas), master purge is still skipped conservatively.

weicao:
Fixes #3087.

## Problem

The PITR disk-pressure cleanup path could misclassify binlog state or continue after command failures:

- replica discovery relied on environment variables that are not available in a DataProtection job;
- the purge loop could delete beyond the safe uploaded/synchronized prefix;
- substring membership checks could match the wrong binlog;
- uploaded-object discovery matched the target Pod name instead of the configured `LOG_PREFIX`;
- failures inside the cleanup function were not guaranteed to stop the script.

## Changes

- discover replicas through the primary using `SHOW REPLICAS`, with the MySQL 5.7 fallback;
- compute one conservative purge boundary and use `PURGE BINARY LOGS TO` only for the safe prefix;
- use exact newline membership checks;
- match uploaded objects using `LOG_PREFIX`;
- enable fail-fast shell behavior while preserving the deliberate no-replica skip boundary.

## Validation

- `bash -n` on the touched script: PASS
- full chart render: PASS
- `git diff --check`: PASS
- exact-head PR CI is green except the currently running ShellSpec job

## Evidence Boundary

These are source, syntax, render, and CI gates. Runtime PITR/disk-pressure behavior remains N=0 and belongs to the test owner.

### DIFF
--- a/addons/mysql/dataprotection/mysql-pitr-backup.sh
+++ b/addons/mysql/dataprotection/mysql-pitr-backup.sh
@@ -193,6 +193,10 @@ cleanup_mysql_binlogs() {
 
     # Get synced binlog files from all replicas
     function get_synced_binlogs() {
+        # Reset per call: this is invoked repeatedly in the long-lived archiver
+        # process, and a leaked value from a previous call would let a purge
+        # proceed on a stale synced position when replica discovery fails.
+        local min_synced_file=""
 
         readarray -t all_binlogs < <(ls -1 "$LOG_DIR"/*-bin.[0-9]* | sort -V)
 
@@ -344,9 +348,12 @@ cleanup_mysql_binlogs() {
     # `if [ $? -ne 0 ]` below reflects get_synced_binlogs' exit code and not
     # the always-zero exit of `local` -- this is what keeps the fail-closed
     # "don't purge when replica sync is unknown" guard actually closed.
+    # Use `if ! ...` so a non-zero get_synced_binlogs does not trip the
+    # actionset-wide `set -e` before this fail-closed guard runs (a plain
+    # `synced_binlogs=$(...)` assignment aborts the whole archiver under set -e;
+    # the `if !` context suppresses that and still catches the failure).
     local synced_binlogs
-    synced_binlogs=$(get_synced_binlogs)
-    if [ $? -ne 0 ] || [ -z "$synced_binlogs" ]; then
+    if ! synced_binlogs=$(get_synced_binlogs) || [ -z "$synced_binlogs" ]; then
         echo "No synced binlog files found"
         return 0
     fi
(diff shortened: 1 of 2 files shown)


weicao:
Absorbed the remaining follow-up from PR #3157 at head 774786e5. The existing main-based PR already contained the set-e and stale-state guards; this update additionally matches uploaded binlogs by the live LOG_PREFIX instead of stale DP_TARGET_POD_NAME. bash -n, helm lint, and git diff --check pass. PR #3157 is superseded by this PR.

PR #3100, opened 2026-07-07
feat: support hadoop and hbase addons on KubeBlocks 1.1

baojun-zhang:
## Description

This PR adds first-class Hadoop HDFS and HBase support to the KubeBlocks 1.1 addon ecosystem by introducing both addon-level definitions and cluster-level deployment charts.

The change is organized around two capability tracks rather than one-off file additions:

1. HDFS capabilities for standalone and HA deployments on KubeBlocks
2. HBase capabilities for both HDFS-backed cluster mode and local-storage standalone mode

At a high level, this PR contributes:

- New `addons/hadoop` and `addons/hbase` packages with `ClusterDefinition`, `ComponentDefinition`, `ComponentVersion`, config templates, startup scripts, config constraints, and release notes
- New `addons-cluster/hadoop` and `addons-cluster/hbase` charts for day-1 cluster provisioning
- Runtime wiring for ZooKeeper-based HA coordination, HDFS/HBase service references, storage layouts, probes, and lifecycle handling
- User-facing documentation and values/schema definitions for installing and operating the new addons on KubeBlocks 1.1

## Related Issue

N/A

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Added a full Hadoop HDFS addon with both standalone and HA topologies, including component models for `hadoop-core`, `journalnode`, `namenode`, and `datanode`
- Added a full HBase addon with both distributed cluster mode and standalone mode, including `hmaster`, `hregionserver`, and `hbase-standalone`
- Added cluster charts for HDFS and HBase so users can provision these stacks directly with cluster-scoped values, resource sizing, storage configuration, and external service references

### HDFS capabilities

- Introduces two HDFS topologies:
  - `standalone`: `hadoop-core` + single `namenode` + single `datanode`
  - `cluster`: `hadoop-core` + `journalnode` quorum + dual `namenode` + scalable `datanode`
- Adds HA-specific runtime building blocks:
  - JournalNode quorum for shared edits
  - Dual NameNode topology with `nn0` and `nn1`
  - ZKFC sidecar for NameNode automatic failover
  - ZooKeeper service reference wiring for HA election and coordination
- Adds NameNode lifecycle automation:
  - one-time format on ordinal `0`
  - `bootstrapStandby` retry loop for the standby NameNode
  - graceful failover attempt during shutdown before daemon termination
  - ZKFC shutdown coordination through a lifecycle signal file
- Adds DataNode operational behavior for production-style HDFS:
  - `hostNetwork` / `hostPID` runtime support
  - per-node data volume mounting
  - decommission flow on `preStop`
  - hostname-aware registration using pod FQDN
- Adds HDFS configuration rendering for both HA and standalone modes:
  - `core-site.xml`
  - `hdfs-site.xml` for NameNode / DataNode / JournalNode
  - `hadoop-env.sh`
  - `log4j.properties`
  - `hadoop-metrics2.properties`
  - include/exclude host lists for decommission control
- Adds day-2 parameterization and validation:
  - replication and max replication
  - NameNode and DataNode handler counts
  - IPC client timeouts and pool sizing
  - permissions and registration checks
  - ZK session timeout, retry count, and parent znode
  - JVM heap sizes for NameNode, DataNode, JournalNode, and ZKFC
- Adds component version mapping for HDFS 3.3.4 and wires dedicated component images for `namenode`, `datanode`, and `journalnode`
- Adds cluster chart values/schema for replicas, storage, service references, ports, resources, HA settings, probes, and update strategy

### HBase capabilities

- Introduces two HBase topologies:
  - `hbase-cluster`: `hmaster` + `hregionserver`
  - `hbase-standalone`: a single-process local-storage HBase deployment
- Adds distributed HBase support on top of HDFS HA:
  - HBase root directory stored in `hdfs://<nameservice>/<rootDir>`
  - HDFS client-side HA wiring via generated `hdfs-site.xml`
  - explicit service references to external ZooKeeper and HDFS NameNode services
  - root layout bootstrap in HDFS before HMaster starts
- Adds standalone HBase support decoupled from HDFS:
  - local `file://`-backed `hbase.rootdir`
  - persistent volume for standalone data
  - ZooKeeper-only external dependency
- Adds runtime scripts for HBase operations:
  - HMaster startup and graceful shutdown
  - RegionServer startup and graceful unload using `RegionMover`
  - HDFS root path initialization for cluster mode
  - lightweight process and HTTP-based readiness/liveness checks
- Adds HBase configuration rendering:
  - `hbase-site.xml` for distributed and standalone modes
  - `core-site.xml` / `hdfs-site.xml` injection for HDFS-backed deployments
  - `hbase-env.sh`
  - HBase metrics and log4j configuration
- Adds operational defaults and tunables:
  - separate JVM sizing for HMaster, RegionServer, and standalone mode
  - JMX ports for HMaster and RegionServer
  - HBase temp/log/PID/data directories
  - HDFS nameservice, HA nodes, RPC/HTTP ports, and replication passthrough
- Adds component version mapping for HBase 2.5.6 and dedicated images for HMaster and RegionServer
- Adds cluster chart values/schema for topology selection, service references, replicas, resources, and standalone storage sizing

### Cross-cutting capabilities

- Adds `README` and release notes for both addons
- Adds chart metadata, helper templates, values, and schema validation for both cluster charts
- Standardizes runtime security context usage and non-root execution defaults across the new components
- Wires KubeBlocks-native `ClusterDefinition`, `ComponentDefinition`, `ComponentVersion`, config templates, and parameter renderers together so the addons are deployable on KubeBlocks 1.1

## Testing

Code review coverage for this draft was based on the full 108-file PR content, including addon definitions, cluster charts, config templates, scripts, values, schemas, and release notes.

Recommended validation for this PR:

- Render `addons-cluster/hadoop` in both `standalone` and `cluster` topologies
- Validate NameNode format / standby bootstrap / ZKFC behavior in an HA environment with ZooKeeper
- Validate DataNode registration and `preStop` decommission behavior under `hostNetwork`
- Render and deploy `addons-cluster/hbase` in both distributed and standalone modes
- Validate HBase cluster mode against an HDFS nameservice-backed deployment
- Validate HBase standalone mode with local `file://` storage

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have tested this on the following platforms:
  - [ ] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist

- [ ] My code follows the project's coding style
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

## Screenshots (if applicable)

N/A

## Additional Notes

- This PR is not just adding charts; it introduces two new addon stacks end to end:
  - Hadoop HDFS addon + cluster chart
  - HBase addon + cluster chart
- HDFS support is broader than a static HA manifest. The PR includes runtime logic for NameNode initialization, standby bootstrap, ZKFC startup, JournalNode-based shared edits, readiness checks, and DataNode decommission handling.
- HBase support is broader than a simple component wrapper. The PR explicitly covers:
  - distributed mode on top of HDFS + ZooKeeper
  - standalone mode on top of local storage + ZooKeeper
  - bootstrap logic for HDFS root layout
  - graceful RegionServer shutdown with region unload
- The HBase distributed path assumes an external HDFS deployment exposed through service references and nameservice-based addressing.
- The HBase standalone path is intentionally decoupled from HDFS and uses `file://` storage semantics.
- Manual integration testing is still important for HA failover semantics, shutdown ordering, and HBase-on-HDFS interoperability under real cluster conditions.


leon-ape (commented):
Findings:
- `addons-cluster/hbase/templates/cluster.yaml` renders `spec.topology: cluster`, while `addons/hbase/templates/clusterdefinition.yaml` only defines `hbase-cluster` and `hbase-standalone`.
- The default HBase cluster renders `serviceRefs.hdfs-namenode...port: client`, but the HDFS namenode ComponentDefinition exposes service ports `fs` and `http`.
- The new Hadoop addon introduces `ParamConfigRenderer` resources in `addons/hadoop/templates/pcr-*.yaml`; PCR is deprecated for the current addon API.
- This PR adds new Hadoop/HBase addons and cluster charts, but the PR has no provision and service-connectivity validation evidence.
- `pr-label-check` is failing because the PR only has `chart-release` and no pick/nopick/pick-* label.

baojun-zhang:
evidence of run success :
image

baojun-zhang:
## Metrics:
### HDFS
image

### HBASE
image

leon-ape (commented):
Finding: The Hadoop addon still introduces ParamConfigRenderer resources for every HDFS configuration path. ParamConfigRenderer is deprecated for the target addon API.\n\nFinding: init-namenode-format.sh suppresses every namenode format failure with || true. A transient or permanent format failure completes the init container successfully, and the unformatted NameNode then enters its normal startup path.\n\nFinding: HBase declares an hdfs-namenode ServiceRef but does not consume the resolved ServiceRef hosts. hdfs-client-site.tpl reconstructs NameNode pod FQDNs from the selected cluster name, fixed component/service names, and ordinals, so valid ServiceRef targets with different backing names cannot be reached.\n\nFinding: This PR adds complete Hadoop and HBase addon surfaces, but provides no actual current-head provisioning and service-connectivity evidence for HDFS standalone/HA or HBase standalone/HDFS-backed modes.

baojun-zhang:
> Finding: The Hadoop addon still introduces ParamConfigRenderer resources for every HDFS configuration path. ParamConfigRenderer is deprecated for the target addon API.\n\nFinding: init-namenode-format.sh suppresses every namenode format failure with || true. A transient or permanent format failure completes the init container successfully, and the unformatted NameNode then enters its normal startup path.\n\nFinding: HBase declares an hdfs-namenode ServiceRef but does not consume the resolved ServiceRef hosts. hdfs-client-site.tpl reconstructs NameNode pod FQDNs from the selected cluster name, fixed component/service names, and ordinals, so valid ServiceRef targets with different backing names cannot be reached.\n\nFinding: This PR adds complete Hadoop and HBase addon surfaces, but provides no actual current-head provisioning and service-connectivity evidence for HDFS standalone/HA or HBase standalone/HDFS-backed modes.



- Upgrade the Hadoop parameter reconfiguration mechanism
 Rebuild the Hadoop parameter reconfiguration flow with the current target addon API’s recommended approach, covering namenode , datanode , journalnode , and shared HDFS settings, with aligned parameter definitions, constraints, defaults, rendering behavior, and documentation.
- Make NameNode format behavior explicit
 Update init-namenode-format.sh to keep the existing idempotent skip path for pre-initialized metadata, use the actual format result as the initContainer success criterion, and add clear initialization logs and validation for both standalone and HA modes.
- Make HBase consume hdfs-namenode ServiceRef results directly
 Inject resolved hdfs-namenode ServiceRef data into hmaster and hregionserver , generate HDFS client configuration from the resolved targets, and refactor hdfs-client-site.tpl so HBase connects to external HDFS through real ServiceRef-derived addresses.
- Add minimal connectivity validation while keeping render checks
 Keep the existing render verification script, add minimal connectivity validation for HDFS standalone, HDFS HA, HBase standalone, and HBase with external HDFS + ZooKeeper, and include current-head deployment and connectivity evidence in the PR.

leon-ape (commented):
Finding: This PR targets release-1.1, but the current head removes ParamConfigRenderer and puts componentDef, templateName, and fileFormatConfig directly in ParametersDefinition. The KubeBlocks 1.1 ParametersDefinition CRD does not define these fields, and its parameter controller still requires ParamConfigRenderer, so the Hadoop parameter definitions cannot bind to their config templates.

Finding: The default hdfs-namenode ServiceRef selects one headless Service. KubeBlocks 1.1 resolves its host and endpoint as one Service DNS value, while hdfs-client-site.tpl requires one host and endpoint for each of nn0 and nn1. The default HBase cluster therefore fails HDFS client config rendering.

Finding: datanode-decommission.sh updates dfs.exclude in the terminating DataNode and then invokes refreshNodes. That config mount is local to the DataNode and does not update the NameNode exclude file, so shutdown and scale-in proceed without decommissioning the DataNode.

Finding: check-hmaster-status.sh and check-hregionserver-status.sh return success as soon as the JVM process exists, before checking the HTTP endpoint. HBase pods can be marked Ready while still blocked on HDFS or ZooKeeper initialization.

Finding: The runtime screenshots predate current head 5dcdb302. The added smoke scripts have no current-head execution output, so HDFS standalone/HA and HBase standalone/HDFS-backed provisioning and service connectivity remain unverified.

baojun-zhang:
evidence of run success :

image

smoke test:
image

image

addons-cluster/hbase/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+    namespace: kubeblocks
+    clusterServiceSelector:
+      cluster: hdfs
+      service:
+        component: namenode
+        service: headless
+        port: fs
+
+hdfs:
+  nameservice: ""
```
leon-ape: The default chart sets `hdfs.nameservice` to an empty string, while the cluster template rejects an empty nameservice. `helm template test addons-cluster/hbase` therefore fails with `hdfs.nameservice must be provided for HBase cluster topology`, so the advertised default topology cannot be installed.

addons/hadoop/scripts/datanode-decommission.sh
```diff
@@ -0,0 +1,176 @@
...
+# 返回值：成功返回 0，超时或状态异常返回非 0。
+wait_for_decommission() {
+    local deadline report_file status
+    report_file="$(mktemp)"
+    trap 'rm -f "${report_file}"' RETURN
+    deadline=$((SECONDS + HDFS_DECOMMISSION_TIMEOUT_SECONDS))
+
+    while (( SECONDS < deadline )); do
+        if "${HADOOP_HOME}/bin/hdfs" dfsadmin -report >"${report_file}" 2>/dev/null; then
+            status="$(awk -v host="${CURRENT_HOST}" '
```
leon-ape: `CURRENT_HOST` is the leaving Pod FQDN, but this parser takes field 2 from the `Name:` line and strips its port. With `dfs.datanode.use.datanode.hostname=false`, that field is the DataNode IP, so `matched` never becomes true and memberLeave waits until timeout even after decommission succeeds.

hack/verify-hbase-hadoop-render.sh
```diff
@@ -0,0 +1,214 @@
...
+  local path="$1"
+  if [[ -e "$path" ]]; then
+    echo "path should not exist: $path" >&2
+    return 1
+  fi
+}
+
+cd "${ROOT_DIR}"
+
+helm dependency build addons-cluster/hbase >/dev/null
```
leon-ape: This verifier builds dependencies only for `addons-cluster/hbase`, then templates three other charts that also declare `kblib`. In a clean checkout it fails immediately because those charts have no local `charts/` dependency, so the added render validation is not reproducible.

leon-ape:
The newly attached runtime screenshots are not bound to current head `fa9252a1` and do not provide auditable commands, environment versions, or complete results for all four modes. Exact-head HDFS standalone/HA and HBase standalone/HDFS-backed provisioning and service connectivity remain unverified.

addons/hbase/scripts/check-hmaster-status.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -e
+
+HMASTER_INFO_PORT=${HMASTER_INFO_PORT:-16010}
+
+if jps 2>/dev/null | grep -q HMaster; then
```
leon-ape, line 6: This readiness script succeeds as soon as the HMaster JVM process exists, before the HTTP readiness endpoint is checked. HMaster can be marked Ready while it is still blocked on HDFS or ZooKeeper initialization.

addons/hbase/scripts/check-hregionserver-status.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -e
+
+REGIONSERVER_INFO_PORT=${REGIONSERVER_INFO_PORT:-16030}
+
+if jps 2>/dev/null | grep -q HRegionServer; then
```
leon-ape, line 6: This readiness script succeeds as soon as the HRegionServer JVM process exists, before the HTTP readiness endpoint is checked. RegionServer can be marked Ready while it is still blocked on HDFS or ZooKeeper initialization.

baojun-zhang: 我核对了当前 HEAD，RegionServer 的 readiness 语义同样已经由单独的 readiness 脚本承担，而不是由这个 status helper 承担。

当前 HRegionServer 的 probe wiring 是：
- startup/liveness：`addons/hbase/templates/cmpd-hregionserver.yaml` 第 62-79 行，使用 `check-hregionserver-live.sh`
- readiness：`addons/hbase/templates/cmpd-hregionserver.yaml` 第 80-88 行，使用 `check-hregionserver-ready.sh`

真正的 readiness 判断在：
- `addons/hbase/scripts/check-hregionserver-ready.sh` 第 1-6 行

这个脚本检查的是 RegionServer 的 HTTP status endpoint（`/rs-status`）。因此，如果 RegionServer 仍然阻塞在 HDFS 或 ZooKeeper 初始化阶段，readiness 会保持失败，直到该 endpoint 可用。

相对地，进程存活判断被单独收敛在：
- `addons/hbase/scripts/check-hregionserver-live.sh` 第 1-12 行

这里的设计是保持职责单一：live 脚本只负责进程是否存活，ready 脚本单独负责服务是否真正 ready，不在一个脚本里同时混合进程探活和依赖初始化完成判断。

addons/hbase/scripts/check-hmaster-status.sh
baojun-zhang: 我核对了当前 HEAD，这里的 readiness 语义已经由单独的 readiness 脚本承担，不是由这个 status helper 承担。

当前 HMaster 的 probe wiring 是：
- startup/liveness：`addons/hbase/templates/cmpd-hmaster.yaml` 第 99-116 行，使用 `check-hmaster-live.sh`
- readiness：`addons/hbase/templates/cmpd-hmaster.yaml` 第 117-125 行，使用 `check-hmaster-ready.sh`

真正的 readiness 判断在：
- `addons/hbase/scripts/check-hmaster-ready.sh` 第 1-6 行

这个脚本只检查 HMaster 的 HTTP status endpoint（`/master-status`）。也就是说，如果 HMaster 仍然阻塞在 HDFS 或 ZooKeeper 初始化阶段，readiness 会继续失败，直到该 endpoint 真正可用为止。

相对地，进程存活判断被单独收敛在：
- `addons/hbase/scripts/check-hmaster-live.sh` 第 1-12 行

这个 live 脚本的职责就是做本地进程存活判断（`jps` / `pgrep`），不负责依赖初始化完成与否的语义。这里保持职责单一：live 只看进程是否存活，ready 单独判断服务是否真正就绪，不把两类语义混在一个脚本里。

baojun-zhang: @leon-ape

addons/hbase/scripts/check-hregionserver-status.sh
baojun-zhang: @leon-ape

addons/hbase/templates/cmpd-hmaster.yaml
```diff
@@ -0,0 +1,283 @@
...
+        ports:
+          - name: hmaster
+            port: {{ .Values.ports.hmaster.rpc }}
+          - name: http
+            port: {{ .Values.ports.hmaster.info }}
+  serviceRefDeclarations:
+    - name: hbase-zookeeper
+      serviceRefDeclarationSpecs:
+        - serviceKind: zookeeper
+          serviceVersion: "^*"
```
leon-ape: ^* is not a valid regular expression. KubeBlocks falls back to literal equality when this pattern cannot be compiled, so these HDFS/ZooKeeper ServiceRef declarations cannot match normal service versions.

addons/hadoop/templates/cmpd-hdfs-datanode.yaml
```diff
@@ -0,0 +1,260 @@
...
+spec:
+  provider: kubeblocks
+  description: {{ .Chart.Description }}
+  serviceKind: hdfs-datanode
+  exporter:
+    containerName: jmx-exporter
+    scrapePath: /metrics
+  {{- include "hadoop.decommissionPolicyRules" . | nindent 2 }}
+  lifecycleActions:
+    memberLeave:
```
leon-ape, line 18: memberLeave.timeoutSeconds is unset, so KubeBlocks applies the system timeout, typically 30 seconds, while datanode-decommission.sh waits for up to 300 seconds. Scale-in can terminate the action before the DataNode reaches Decommissioned.

addons-cluster/hadoop/templates/decommission-rbac.yaml
```diff
@@ -0,0 +1,37 @@
...
+  labels:
+{{ include "kblib.clusterLabels" . | nindent 4 }}
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+    resourceNames:
+      - {{ $stateConfigMapName }}
+    verbs:
```
leon-ape, line 23: The decommission scripts POST this ConfigMap when it is missing, but this Role grants only get, patch, and update and is restricted by resourceNames. The POST is denied, so an absent state ConfigMap cannot be recreated and DataNode decommission fails.

addons/hbase/scripts/init-hdfs-root-layout.sh
```diff
@@ -0,0 +1,28 @@
...
+    "/${HBASE_ROOT_DIR}/MasterData"
+    "/${HBASE_ROOT_DIR}/MasterData/WALs"
+)
+
+for p in "${PATHS[@]}"; do
+    if timeout 30 ${HBASE_HOME}/bin/hbase org.apache.hadoop.fs.FsShell -test -e "${p}" 2>/dev/null; then
+        echo "[$(date)] Path ${p} already exists, skipping."
+    else
+        echo "[$(date)] Creating ${p}..."
+        timeout 30 ${HBASE_HOME}/bin/hbase org.apache.hadoop.fs.FsShell -mkdir -p "${p}" || true
```
leon-ape: Every HDFS mkdir failure is discarded by || true, and the init container then exits successfully. HMaster can start after the required root, WAL, and data layout failed to initialize.

addons/hadoop/scripts/start-datanode.sh
```diff
@@ -0,0 +1,32 @@
...
+shutdown() {
+  echo "[$(date)] Stopping DataNode..."
+  "${HADOOP_HOME}/bin/hdfs" --daemon stop datanode || true
+  exit 0
+}
+trap shutdown SIGTERM SIGINT
+
+if [[ "${HDFS_DECOMMISSION_ENABLED:-true}" == "true" ]]; then
+  # ponytail: 启动前清理残留 decommission 状态属于幂等善后，不该因为控制面瞬时抖动阻塞 DataNode 拉起；若后续需要强一致，可改为带退避的预检查。
+  "$(dirname "$0")/datanode-decommission.sh" unregister || true
```
leon-ape: A failed unregister is ignored and there is no later retry. If this DataNode still has an entry in the shared decommission ConfigMap, it starts while remaining in dfs.exclude and never returns to service.

addons/hbase/templates/cmpd-hmaster.yaml
baojun-zhang: replaced it with a valid catch-all regex so the HBase/Hadoop ServiceRef declarations match normal ServiceDescriptor versions as intended.

addons/hadoop/templates/cmpd-hdfs-datanode.yaml
baojun-zhang: added `memberLeave.timeoutSeconds: -1` for the DataNode lifecycle action. This prevents KubeBlocks from terminating the action before `datanode-decommission.sh` completes, and leaves the effective timeout under the script's own 300-second control.

addons-cluster/hadoop/templates/decommission-rbac.yaml
baojun-zhang: The RBAC rules now match the script behavior. `get/patch/update` remain scoped to the target state ConfigMap, and I added a separate `create` permission for `configmaps` so `ensure_state_configmap` can recreate the ConfigMap when it is absent.

addons/hbase/scripts/init-hdfs-root-layout.sh
baojun-zhang: removed the `|| true` from `init-hdfs-root-layout.sh`. Creating the HBase root/WAL/data layout is a required init step, so the init container should fail and retry if HDFS is unavailable or the mkdir operation fails, instead of allowing HMaster to start with an incomplete layout.

addons/hadoop/scripts/start-datanode.sh
baojun-zhang: kept the startup path non-blocking, but changed it from a one-shot `unregister || true` into: try unregister once before startup, and if it fails, retry in the background until it succeeds. This preserves availability during transient control-plane failures while closing the gap where a DataNode could start and remain stuck in the shared decommission state / `dfs.exclude`.

baojun-zhang:
## 问题背景

在较新的 Helm / Kubernetes 校验链路，以及 HDFS HA + HBase 多副本拓扑场景下，当前 HBase/Hadoop addons 暴露出一组相互关联的问题：

1. 部分 `ComponentDefinition` 使用了 CRD schema 不支持的字段；
2. HBase 对 NameNode pod FQDN 的依赖处理不完整，导致拓扑配置无法稳定渲染；
3. RegionServer 默认 JVM 堆大小与常见容器内存限制之间缺少安全余量，容易在运行时 OOM。

这些问题会表现为：
- addon definitions 安装失败；
- HBase cluster 拓扑配置渲染失败或 pod 无法创建；
- RegionServer 启动后因内存不足反复崩溃。

---

## 具体问题

### 1. `ComponentDefinition.spec.vars[].valueFrom` 中使用了非法 `fieldRef`

当前部分 HBase/Hadoop `ComponentDefinition` 在 `spec.vars[].valueFrom` 下定义了 `fieldRef`，例如通过 pod annotation 获取 pod FQDN。

但从 CRD schema 角度看，`spec.vars[].valueFrom` 并不支持 `fieldRef`。这会带来两类问题：

- 在较宽松的处理链路下，这类字段可能被静默剪枝，表现为“安装看似成功，但变量实际上从未生效”；
- 在更严格的 Helm / typed patch / schema 校验链路下，会直接报错，导致 `ComponentDefinition` 无法安装。

这说明该配置不仅存在兼容性问题，本身也是无效配置。

---

### 2. HBase 对 NameNode pod FQDN 的处理不完整，HA 拓扑无法稳定部署

当前 HBase addon 依赖 NameNode pod FQDN 来渲染 HDFS client 配置，但相关逻辑存在缺口：

- `serviceRefVarRef.podFQDNs` 在某些位置被当作必填，但引用的 service 并不一定提供该字段；
- 即使改成可选，如果没有 values 层的明确 fallback，配置仍然无法渲染；
- RegionServer 的 hostname 处理默认假设 pod FQDN 一定存在；
- HMaster readiness 对 HA 场景不够友好，standby master 可能永远无法进入 Ready。

这会导致以下结果之一：

- HBase 配置模板渲染失败；
- HMaster / RegionServer pod 因变量无法解析而无法创建；
- RegionServer hostname 为空或不正确；
- 多副本 HMaster 拓扑中，standby 实例长期不 Ready。

---

### 3. HBase RegionServer 默认堆配置缺少非堆余量，容易 OOM

当前默认 `hregionserverHeapSize` 偏高，在常见容器内存限制下没有给 JVM 非堆内存留下足够空间。

例如：
- 当堆大小接近或等于容器 memory limit 时，
- metaspace、线程栈、direct/native memory、JVM 自身开销都没有安全余量，

最终会表现为启动失败、native memory allocation failed、或反复 CrashLoop / OOMKilled。

这类问题尤其隐蔽，因为从表面上看 values 配置“合法”，但默认值本身并不自洽。

---

## 建议修改方案

### A. 清理所有 `spec.vars[].valueFrom.fieldRef` 非法配置

请从受影响的 HBase/Hadoop `ComponentDefinition` 模板中删除 `spec.vars[].valueFrom.fieldRef`。

建议原则：

- `ComponentDefinition.spec.vars` 里只使用 CRD schema 支持的来源：
  - `clusterVarRef`
  - `serviceRefVarRef`
  - `componentVarRef`
  - `value`
  - `expression`
- 如果确实需要 Kubernetes 原生 `fieldRef`，仅在容器 `env` 中使用，而不要放在 `spec.vars[].valueFrom` 中。

这样可以同时解决：
- 严格 schema 校验下的安装失败；
- 宽松链路下“配置被静默忽略但误以为生效”的问题。

---

### B. 补齐 HBase 对 NameNode pod FQDN 的完整 fallback 机制

建议统一调整为：

1. 将 `serviceRefVarRef.podFQDNs` 从 `Required` 改为 `Optional`；
2. 在 values 中显式提供 `hdfs.namenodePodFQDNs` 作为 fallback 配置入口；
3. 表达式逻辑优先读取 serviceRef 提供的 pod FQDNs，缺失时再回退到 values 中定义的 `hdfs.namenodePodFQDNs`；
4. RegionServer hostname 使用更稳妥的 fallback 形式，例如：
   - `${POD_FQDN:-$(hostname -f)}`
5. 调整 HMaster readiness 探针，避免使用只适用于 active master 的检查方式；建议改为对 active / standby 都更稳妥的探测路径，例如基于 JMX 或等价的健康检查接口。

目标是让 HBase addon 在以下场景下都能稳定工作：

- HDFS HA
- 多副本 HMaster
- `serviceRef` 未提供 `podFQDNs`
- 依赖 values fallback 提供 NameNode pod FQDN 列表

---

### C. 下调 RegionServer 默认堆大小，保证默认值自洽

建议将默认 `hregionserverHeapSize` 从 `4g` 调整为 `3g`，为非堆内存预留明确余量。

同时建议在 values 注释中明确说明：

- JVM heap 必须严格小于容器 memory limit；
- 需要为 metaspace、线程栈、native/direct memory 预留足够空间；
- 默认值应保证“按默认配置部署即可稳定启动”，而不是依赖额外调参。

---

## 期望结果

完成上述修复后，期望达到以下效果：

1. HBase/Hadoop addon definitions 在严格 schema 校验和 Helm 4 场景下可以正常安装；
2. HBase 在 HDFS HA 场景中可以稳定渲染和部署，不依赖隐式字段或不完整 fallback；
3. RegionServer 默认配置具备合理内存余量，不会因默认值本身不自洽而 OOM；
4. 默认交付行为应当是“可安装、可渲染、可启动”，而不是安装成功后在运行期暴露隐藏缺陷。

baojun-zhang:
@leon-ape 辛苦 review 哈

addons-cluster/hbase/templates/cluster.yaml
```diff
@@ -0,0 +1,244 @@
...
+              port: {{ .Values.serviceRefs.hbaseZookeeper.clusterServiceSelector.service.port }}
+{{- if .Values.serviceRefs.hdfsNamenode.enabled }}
+        - name: hdfs-namenode
+          namespace: {{ .Values.serviceRefs.hdfsNamenode.namespace }}
+          clusterServiceSelector:
+            cluster: {{ .Values.serviceRefs.hdfsNamenode.clusterServiceSelector.cluster }}
+            service:
+              component: {{ .Values.serviceRefs.hdfsNamenode.clusterServiceSelector.service.component }}
+              service: {{ .Values.serviceRefs.hdfsNamenode.clusterServiceSelector.service.service }}
+              port: {{ .Values.serviceRefs.hdfsNamenode.clusterServiceSelector.service.port }}
```
leon-ape, line 43: This ServiceRef selects only the HDFS Service and does not request `clusterServiceSelector.podFQDNs`. KubeBlocks 1.1 therefore leaves `serviceRefVarRef.podFQDNs` unset; the default fallback is empty, so `hdfs-client-site.tpl` fails and the default HBase cluster cannot provision.

addons/hadoop/templates/cmpd-hdfs-datanode.yaml
```diff
@@ -0,0 +1,257 @@
...
+    - name: HDFS_DECOMMISSION_POLL_INTERVAL_SECONDS
+      value: {{ .Values.decommission.pollIntervalSeconds | quote }}
+    - name: HDFS_DECOMMISSION_STATE_CONFIGMAP_NAME_DEFAULT
+      value: {{ .Values.decommission.stateConfigMapName | quote }}
+    - name: HDFS_DECOMMISSION_STATE_CONFIGMAP_NAME
+      expression: {{ `{{ if index . "HDFS_DECOMMISSION_STATE_CONFIGMAP_NAME_DEFAULT" }}{{ .HDFS_DECOMMISSION_STATE_CONFIGMAP_NAME_DEFAULT }}{{ else }}{{ .CLUSTER_NAME }}-hdfs-decommission-state{{ end }}` | toYaml }}
+    - name: HDFS_DECOMMISSION_DYNAMIC_EXCLUDE_FILE
+      value: {{ printf "%s/dfs.exclude.dynamic" .Values.runtime.configDir | quote }}
+  configs:
+    - name: hdfs-common-config
```
leon-ape, line 246: The Hadoop config templates do not declare `externalManaged: true`. KubeBlocks 1.1 excludes these configs from ParamConfigRenderer resolution, so the added parameter definitions do not bind to or manage the HDFS configuration files.

addons/hadoop/templates/cmpd-hdfs-datanode.yaml
```diff
@@ -0,0 +1,257 @@
...
+          - name: DEBUG_MODEL
+            value: {{ .Values.runtime.debug | quote }}
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          - name: POD_FQDN
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.annotations['kubeblocks.io/pod-fqdn']
```
leon-ape, line 69: `kubeblocks.io/pod-fqdn` is a controller implementation annotation, not a KubeBlocks 1.1 public addon contract. If it is absent or changes, `POD_FQDN` is empty and the DataNode advertises an empty `dfs.datanode.hostname`.

addons/hadoop/config/hdfs-namenode.tpl
```diff
@@ -0,0 +1,211 @@
...
+        <value>{{- .HDFS_NAMENODE_NAME_DIR }}</value>
+    </property>
+
+    <property>
+        <name>dfs.ha.namenodes.{{- .CLUSTER_NAME }}</name>
+        <value>{{ join "," $name_node_ids }}</value>
+    </property> {{- range $nn_ordinal, $nn_id :=
+    $name_node_ids }} <property>
+        <name>dfs.namenode.rpc-address.{{- $.CLUSTER_NAME }}.{{ $nn_id }}</name>
+        <value>{{- $.CLUSTER_NAME }}-namenode-{{ $nn_ordinal }}.{{- $.CLUSTER_NAME
```
leon-ape: These peer addresses depend on KubeBlocks' internal Pod and headless-Service naming format. Those names are not part of the public addon API, so a compatible controller naming change breaks NameNode RPC/HTTP discovery.

addons/hbase/templates/cmpd-hregionserver.yaml
```diff
@@ -0,0 +1,246 @@
...
+        ports:
+          - name: hregionserver
+            port: {{ .Values.ports.hregionserver.rpc }}
+          - name: http
+            port: {{ .Values.ports.hregionserver.info }}
+  serviceRefDeclarations:
+    - name: hbase-zookeeper
+      serviceRefDeclarationSpecs:
+        - serviceKind: zookeeper
+          serviceVersion: "^*"
```
leon-ape: `^*` is not a valid Go regular expression. KubeBlocks 1.1 falls back to literal equality when compilation fails, so normal ZooKeeper/HDFS ServiceDescriptor versions cannot match. The same invalid pattern remains in HBase standalone, this HRegionServer definition, and the Hadoop NameNode definition.

addons/hadoop/scripts/check-data-status.sh
```diff
@@ -0,0 +1,13 @@
...
+set -o nounset
+set -o pipefail
+
+_PORTS="${HDFS_DATANODE_HTTP_PORT:-9864}"
+_URL_PATH="jmx?qry=Hadoop:service=DataNode,name=DataNodeInfo"
+_CLUSTER_ID=""
+for _PORT in $_PORTS; do
+  _CLUSTER_ID+=$(curl -s http://localhost:${_PORT}/$_URL_PATH | grep ClusterId) || true
+done
+echo $_CLUSTER_ID | grep -q -v null
```
leon-ape: When curl fails or returns no JMX data, `echo "" | grep -q -v null` succeeds. DataNode startup, readiness, and liveness therefore pass while the service is unreachable; `check-journal-status.sh` has the same fail-open behavior.

addons/hadoop/scripts/init-zkfc-format.sh
```diff
@@ -0,0 +1,13 @@
...
+set -o errtrace
+set -o nounset
+set -o pipefail
+
+: "${HADOOP_HOME:=/opt/hadoop}"
+: "${HADOOP_CONF_DIR:=${HADOOP_HOME}/etc/hadoop}"
+export HADOOP_HOME HADOOP_CONF_DIR
+
+echo "Formatting ZKFC (creating HA znode in ZooKeeper)..."
+echo "N" | "${HADOOP_HOME}/bin/hdfs" zkfc -formatZK || true
```
leon-ape: Every NameNode Pod runs this initialization and every `formatZK` failure is discarded. If ZooKeeper is temporarily unavailable or HA state creation fails, both Pods continue without the znode required for automatic failover.

addons/hadoop/templates/cmpd-hdfs-namenode.yaml
```diff
@@ -0,0 +1,433 @@
+apiVersion: apps.kubeblocks.io/v1
+kind: ComponentDefinition
+metadata:
+  name: hdfs-namenode
+  labels:
+    {{- include "hadoop.labels" . | nindent 4 }}
+  annotations:
+    {{- include "hadoop.apiVersion" . | nindent 4 }}
+spec:
```
leon-ape, line 9: This ComponentDefinition has no `replicasLimit`, although its bootstrap logic only supports ordinals 0 and 1. A later scale operation can create an unsupported one- or three-NameNode topology. JournalNode likewise has no limit and can be scaled below the three-member QJM minimum.

addons/hadoop/scripts/init-namenode-format.sh
```diff
@@ -0,0 +1,38 @@
...
+
+if [[ -d "$NN_CURRENT_DIR" ]] && find "$NN_CURRENT_DIR" -maxdepth 1 -type f -name 'fsimage_*' ! -name '*.md5' 2>/dev/null | grep -q .; then
+  echo "Valid fsimage already exists, skipping format"
+  exit 0
+fi
+
+NAMESERVICES=$("${HADOOP_HOME}/bin/hdfs" getconf -confKey dfs.nameservices 2>/dev/null || echo "")
+if [[ -z "$NAMESERVICES" ]]; then
+  echo "No dfs.nameservices configured (standalone mode), using default format"
+  echo "N" | "${HADOOP_HOME}/bin/hdfs" namenode -format
```
leon-ape, line 34: If ordinal 0 loses its local metadata while ordinal 1 and the JournalNodes still contain the valid HA namespace, this path formats a new namespace instead of recovering from the surviving peer. The recreated NameNode gets a different namespace/cluster identity and cannot rejoin the existing HA cluster.

addons/hbase/scripts/check-hregionserver-ready.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -euo pipefail
+
+REGIONSERVER_INFO_PORT="${REGIONSERVER_INFO_PORT:-${HBASE_REGIONSERVER_INFO_PORT:-16030}}"
+
+curl -sf --max-time 2 -o /dev/null "http://127.0.0.1:${REGIONSERVER_INFO_PORT}/rs-status"
```
leon-ape, line 9: HBase 2.5.6 starts the RegionServer web UI before `reportForDuty`, and `/rs-status` returns HTTP 200 before Master registration completes. A RegionServer that cannot yet serve regions is therefore marked Ready.

addons/hbase/scripts/start-hregionserver.sh
```diff
@@ -0,0 +1,32 @@
...
+function shutdown() {
+    echo "[$(date)] Stopping RegionServer gracefully..."
+    host=$(hostname -f 2>/dev/null || hostname)
+    export HBASE_STOP_TIMEOUT=20
+
+    echo "[$(date)] Disabling balancer before region unload..."
+    echo "balance_switch false" | ${HBASE_HOME}/bin/hbase shell 2>/dev/null || true
+
+    echo "[$(date)] Unloading regions from ${host}..."
+    ${HBASE_HOME}/bin/hbase org.apache.hadoop.hbase.util.RegionMover -m 6 -r ${host} -o unload 2>&1 || true
```
leon-ape: RegionMover uses acknowledged unload, but every unload failure is discarded and the RegionServer is stopped anyway. A migration timeout, unavailable Master, or insufficient destination capacity therefore turns planned leave into region unavailability and recovery.

addons/hbase/config/hbase-site-cluster.tpl
```diff
@@ -0,0 +1,294 @@
...
+    </property>
+
+    <property>
+        <name>hbase.security.authorization</name>
+        <value>{{ .HBASE_SECURITY_AUTHORIZATION }}</value>
+    </property>
+
+    <property>
+        <name>hbase.unsafe.stream.capability.enforce</name>
+        <value>false</value>
```
leon-ape: This disables HBase 2.5.6's WAL stream capability check. HBase can start even when the backing filesystem does not provide the required `hflush`/`hsync` semantics, removing a durability precondition for WAL writes.

addons/hbase/templates/cmpd-hmaster.yaml
```diff
@@ -0,0 +1,283 @@
...
+    - name: HBASE_REGIONSERVER_JMX_PORT
+      value: {{ .Values.ports.hregionserver.jmx | quote }}
+    - name: HBASE_HEAP_SIZE
+      value: {{ .Values.jvm.hmasterHeapSize | quote }}
+    - name: ZOOKEEPER_QUORUM
+      valueFrom:
+        serviceRefVarRef:
+          name: hbase-zookeeper
+          host: Required
+    - name: ZOOKEEPER_CLIENT_PORT
```
leon-ape, line 238: The ZooKeeper ServiceRef resolves the selected service port, but HBase ignores it and always renders the fixed `hbase.zkClientPort` value. Any valid ZooKeeper service whose client port is not 2181 is bound through the ServiceRef but remains unreachable.

addons/hbase/config/hbase-site-cluster.tpl
```diff
@@ -0,0 +1,294 @@
...
+    </property>
+
+    <property>
+        <name>hbase.storescanner.parallel.seek.enable</name>
+        <value>true</value>
+    </property>
+
+    <property>
+        <name>hbase.ipc.server.callqueue.type</name>
+        <value>plunger</value>
```
leon-ape: `plunger` is not a supported HBase 2.5.6 call queue type. HBase silently falls back to FIFO, so the shipped configuration does not produce its declared scheduling behavior.

addons/hbase/config/log4j.properties
```diff
@@ -0,0 +1,69 @@
+# 根日志配置
```
leon-ape, line 1: HBase 2.5.6 uses Log4j2, but this is a Log4j 1 configuration mounted as `log4j.properties`. Its rolling, security-audit, and HTTP access logging settings are not applied by the target engine version.

addons/hbase/config/hbase-site-cluster.tpl
```diff
@@ -0,0 +1,294 @@
...
+
+    <property>
+        <name>hbase.unsafe.stream.capability.enforce</name>
+        <value>false</value>
+    </property>
+
+    <property>
+        <name>hbase.master.logcleaner.plugins</name>
+        <value>
+            org.apache.hadoop.hbase.master.cleaner.TimeToLiveLogCleaner,org.apache.hadoop.hbase.master.cleaner.TimeToLiveProcedureWALCleaner,org.apache.hadoop.hbase.replication.master.ReplicationLogCleaner</value>
```
leon-ape: Overriding the HBase 2.5.6 cleaner lists here drops the default Master local-store WAL and HFile cleaners. Archived Master local-store files are no longer covered by the engine's default retention cleanup.

leon-ape (commented):
The existing current-head provision and service-connectivity evidence finding remains valid. The attached runtime screenshots predate the current head and do not cover the current lifecycle, ServiceRef, parameter-management, and HA paths.

addons-cluster/hadoop/values.yaml
```diff
@@ -0,0 +1,146 @@
...
+  registration:
+    ipHostnameCheck: false
+  ha:
+    nameNodeIds: "nn0,nn1"
+    automaticFailoverEnabled: true
+    fencingMethods: "sshfence\nshell(/bin/true)"
+    fencingSshConnectTimeoutMs: 30000
+    zookeeper:
+      parentZnodePrefix: /hadoop-ha
+      includeClusterUID: true
```
leon-ape, line 96: `includeClusterUID` is never consumed, while the ZooKeeper parent remains shared. Two same-named HDFS clusters in different namespaces that reference the same ZooKeeper use the same HA election path and can participate in one failover state.

addons/hadoop/config/hdfs-namenode-config-constraint.cue
```diff
@@ -0,0 +1,252 @@
...
+	"dfs.namenode.fs-limits.min-block-size": int & >=1 | *1048576
+
+	// NameNode的RPC服务端用于监听客户端请求的线程数。默认值128，取值范围1~1024
+	"dfs.namenode.handler.count": int & >=1 & <=1024 | *128
+
+	// 再次检查DataNode可用性的间隔，单位：毫秒。默认值300000，取值范围1000~86400000
+	"dfs.namenode.heartbeat.recheck-interval": int & >=1000 & <=86400000 | *300000
+
+	// Inode属性插件。默认值com.huawei.hadoop.adapter.hdfs.plugin.HWRangerHdfsAuthorizer
+	"dfs.namenode.inode.attributes.provider.class": string | *"com.huawei.hadoop.adapter.hdfs.plugin.HWRangerHdfsAuthorizer"
```
leon-ape: This Apache Hadoop 3.3.4 parameter contract defaults an optional provider to a Huawei-specific class that is not present in the target engine. Once parameter management is active, the generated NameNode configuration can fail class loading. The same contract also ships incorrect Apache defaults for `dfs.ha.tail-edits.period` and `dfs.namenode.checkpoint.period`.

addons-cluster/hbase/templates/cluster.yaml
baojun-zhang: [already fixed eec58ec2] Current HEAD handles this by making `serviceRefVarRef.podFQDNs` optional and adding a values fallback chain. In `cmpd-hmaster.yaml` and `cmpd-hregionserver.yaml`, `HDFS_NAMENODE_POD_FQDNS` is resolved from ServiceRef when available and falls back to `HDFS_NAMENODE_POD_FQDNS_DEFAULT`; `hdfs-client-site.tpl` then fails fast only if both are empty. This keeps the default cluster topology installable on KubeBlocks 1.1 without forcing `clusterServiceSelector.podFQDNs`.

addons/hadoop/templates/cmpd-hdfs-datanode.yaml
baojun-zhang: The current fix plan is to add `externalManaged: true` to the Hadoop config bindings so ParamConfigRenderer actually manages these HDFS files on KubeBlocks 1.1. The parameter definitions added in `7cf3e4af` are not sufficient by themselves; the config entries still need to opt into external management in the ComponentDefinition/PCR wiring.

addons/hadoop/templates/cmpd-hdfs-datanode.yaml
baojun-zhang: The fix here is not to remove the container env injection entirely, but to remove the hard runtime dependency on that annotation. `POD_FQDN` can stay as a best-effort input, and `hadoop-env.sh.tpl` should change `-Ddfs.datanode.hostname=${POD_FQDN}` to `${POD_FQDN:-$(hostname -f 2>/dev/null || hostname)}`. That way the DataNode no longer advertises an empty hostname if `kubeblocks.io/pod-fqdn` is absent or changes.

addons/hadoop/config/hdfs-namenode.tpl
baojun-zhang: Planned change here is to stop hard-coding `cluster-name + ordinal + headless-service` in `hdfs-namenode.tpl` and `hdfs-datanode.tpl`, and instead render the peer RPC/HTTP addresses from explicit pod FQDN inputs. That keeps HA discovery on the addon contract rather than on controller-internal naming rules.

addons/hbase/templates/cmpd-hregionserver.yaml
baojun-zhang: The remaining `^*` copies should all be replaced with a valid catch-all regex, not just the HMaster copy that was already updated earlier. The unresolved set here is the HBase standalone definition, this HRegionServer definition, and the Hadoop NameNode definition; they all need the same cleanup so ServiceRef version matching stays consistent.

addons/hadoop/scripts/check-data-status.sh
baojun-zhang: The current repair plan is to make both `check-data-status.sh` and `check-journal-status.sh` fail closed. `curl` / JMX parse errors must fail the probe, and success should require a non-empty response with a non-null `ClusterId`. The current `... || true` path is too permissive for startup, readiness, and liveness.

addons/hadoop/scripts/init-zkfc-format.sh
baojun-zhang: Planned change is to stop discarding every `zkfc -formatZK` error. The script should only treat the already-initialized case as idempotent success; ZooKeeper connectivity failures or HA state creation failures must fail the init path instead of letting both NameNode Pods continue without the election znode.

addons/hadoop/templates/cmpd-hdfs-namenode.yaml
baojun-zhang: The fix plan is to declare explicit `replicasLimit` for the topologies the bootstrap logic actually supports: cluster NameNode fixed at 2 replicas, standalone NameNode fixed at 1 replica, and JournalNode constrained so it cannot scale below the 3-member QJM minimum. That keeps later scale operations from creating unsupported HA layouts.

addons/hadoop/scripts/init-namenode-format.sh
baojun-zhang: Planned fix is to make formatting contingent on the whole HA namespace state, not only on local metadata presence. Before calling `namenode -format`, the script should check the surviving peer and shared edits; if the namespace already exists elsewhere, this pod should recover from the existing HA cluster instead of formatting a new namespace.

addons/hbase/scripts/check-hregionserver-ready.sh
baojun-zhang: The repair here is to strengthen readiness semantics. `/rs-status` returning HTTP 200 is not sufficient for HBase 2.5.6 because the RegionServer web UI can come up before `reportForDuty`. The readiness script should check that the RegionServer has completed Master registration / entered the online servers set, not only that the local web endpoint is alive.

addons/hbase/scripts/start-hregionserver.sh
baojun-zhang: The shutdown path can stay graceful, but it should stop swallowing every RegionMover unload failure. An acknowledged unload failure needs to block the planned leave, or at least prevent us from treating the handoff as clean; otherwise a planned scale-in can still degrade into region unavailability and recovery.

addons/hbase/config/hbase-site-cluster.tpl
baojun-zhang: The current plan is to stop disabling this globally. Instead of shipping `hbase.unsafe.stream.capability.enforce=false` by default, the addon should either rely on the engine default or gate this behind an explicit compatibility switch for environments that truly require it. That preserves the WAL durability precondition by default.

addons/hbase/templates/cmpd-hmaster.yaml
baojun-zhang: Planned fix is to make the rendered ZooKeeper quorum and client port come from the same source. If the ServiceRef resolves a ZooKeeper service with a non-2181 client port, the port used by HBase must follow that resolved endpoint rather than a separate fixed `hbase.zkClientPort` value.

addons/hbase/config/hbase-site-cluster.tpl
baojun-zhang: Planned change is to drop `plunger` and use a value actually supported by HBase 2.5.6, or omit the property and let the engine default apply. Keeping an unsupported value only produces a silent FIFO fallback while advertising behavior that never takes effect.

addons/hbase/config/log4j.properties
baojun-zhang: Planned fix is to migrate the mounted logging config from Log4j 1 format to a Log4j 2 configuration that HBase 2.5.6 actually consumes. That is the way to make rolling, security-audit, and HTTP access logging settings effective again under the target engine version.

addons/hbase/config/hbase-site-cluster.tpl
baojun-zhang: The repair plan here is not to override the full cleaner lists. The addon should preserve the upstream default cleaners and only append explicit extra cleaners if they are really needed, so Master local-store WAL and HFile cleanup is not dropped by the override.

addons-cluster/hadoop/values.yaml
baojun-zhang: The fix here is to actually consume `includeClusterUID` in the rendered ZooKeeper parent path. `parentZnodePrefix` should remain the base prefix, and when `includeClusterUID=true` the final `ha.zookeeper.parent-znode` should include the cluster UID so different clusters never share one HA election namespace. This should be solved by wiring the UID into the rendered path, not by adding a validation that merely rejects the current configuration.

addons/hadoop/config/hdfs-namenode-config-constraint.cue
baojun-zhang: Planned change is to correct the Apache Hadoop 3.3.4 parameter contract defaults themselves, not just to enable parameter management around them. In particular, remove the Huawei-specific optional provider default and restore the upstream defaults for `dfs.ha.tail-edits.period` and `dfs.namenode.checkpoint.period`.

addons/hadoop/scripts/check-data-status.sh
baojun-zhang: Updated the probe fix with role-specific readiness semantics. DataNode still requires a non-empty ClusterId. JournalNode is different: Hadoop 3.3.4 exposes ClusterIds only after a journal is formatted, so requiring it can deadlock first bootstrap. check-journal-status.sh now fails closed on curl/JMX errors and succeeds only when JournalNodeInfo.HostAndPort is non-empty, which verifies that both the HTTP bean and RPC binding are ready. Added regression cases for a valid binding, an empty binding, and an unreachable JMX endpoint.

addons/hbase/scripts/check-hregionserver-ready.sh
baojun-zhang: Implemented a one-time reportForDuty registration gate without running HBase Shell on every probe. Readiness first checks /rs-status; while the registration marker is absent it runs status simple and matches this RegionServer FQDN and RPC port, then writes ${HBASE_PID_DIR}/report-for-duty.ready. start-hregionserver.sh removes the marker on every container start. Subsequent readiness probes use the HTTP check plus marker only, and the existing 10-second Kubernetes probe timeout bounds the initial registration query. Regression coverage verifies that HBase Shell runs exactly once after registration.

addons/hbase/scripts/start-hregionserver.sh
baojun-zhang: Moved RegionMover out of the SIGTERM trap into a real KubeBlocks lifecycleActions.memberLeave action with a 300-second timeout. The dedicated hregionserver-member-leave.sh disables the balancer, unloads KB_LEAVE_MEMBER_POD_FQDN, restores the balancer through an EXIT trap, and propagates unload failures so KubeBlocks does not release the replica. The SIGTERM path now only performs the RegionServer daemon stop, so ordinary restarts remain graceful. Added regression coverage for successful unload, failure propagation, and balancer restoration.

baojun-zhang:
@leon-ape 感谢提供意见， 我已重新修复了上述问题，幸苦 review 哈～

addons/hbase/scripts/hregionserver-member-leave.sh
```diff
@@ -0,0 +1,32 @@
...
+
+BALANCER_DISABLED=false
+
+# Restores the HBase balancer after the unload attempt.
+# Parameters: none.
+# Returns: 0; restoration failures are logged without replacing the unload result.
+restore_balancer() {
+  if [[ "${BALANCER_DISABLED}" == "true" ]]; then
+    echo "Re-enabling balancer after RegionServer unload..."
+    printf "balance_switch true\n" | "${HBASE_HOME}/bin/hbase" shell -n 2>/dev/null || echo "Failed to re-enable the balancer" >&2
```
leon-ape: **P1:** This cleanup always sends `balance_switch true` and suppresses restoration failures. HBase returns the previous balancer state, but it is discarded here, so scaling in while balancing is already disabled silently enables cluster-wide region movement; conversely, a failed re-enable after a successful unload is still reported as a successful `memberLeave`, leaving the cluster in the wrong global state.

addons/hbase/scripts/start-hmaster.sh
```diff
@@ -0,0 +1,59 @@
...
+    for endpoint in "${zk_endpoints[@]}"; do
+        endpoint="${endpoint//[[:space:]]/}"
+        [[ -z "${endpoint}" ]] && continue
+        host="${endpoint%:*}"
+        port="${endpoint##*:}"
+        if [[ "${host}" == "${port}" ]]; then
+            port="${ZOOKEEPER_CLIENT_PORT:-2181}"
+        fi
+        if exec 3<>"/dev/tcp/${host}/${port}" 2>/dev/null; then
+            if printf 'ruok\n' >&3 2>/dev/null && IFS= read -r -t 2 zk_reply <&3 && [[ "${zk_reply}" == "imok" ]]; then
```
leon-ape: **P1:** ZooKeeper’s standard `ruok` response is the four bytes `imok` without a newline. Bash `read` therefore assigns `imok` but returns non-zero at EOF, so this `&&` predicate rejects even a whitelisted standard server and both HMaster and RegionServer time out before HBase starts. ZooKeeper 3.5.3+ also disables `ruok` by default, so valid external ServiceRefs fail the same gate.

baojun-zhang: The current predicate requires both a successful `read` and `zk_reply == imok`, but ZooKeeper can return `imok` without a trailing newline, which makes Bash `read` return non-zero at EOF even after assigning the value. I will change this gate to use TCP connectivity as the readiness condition and keep `ruok` only as an optional signal instead of a hard requirement.

addons/hbase/scripts/hregionserver-member-leave.sh
baojun-zhang: The script currently discards the original balancer state, always restores `balance_switch true`, and only logs restoration failures. I will preserve the original balancer state, restore it only when it was originally enabled, and propagate restoration failures so `memberLeave` cannot report success while leaving the cluster in the wrong global state.

addons/hbase/scripts/hregionserver-member-leave.sh
```diff
@@ -0,0 +1,84 @@
...
+#   Does not return; exits with the final status code.
+on_exit() {
+  local rc="$1"
+  restore_balancer || rc=1
+  exit "${rc}"
+}
+
+trap 'on_exit $?' EXIT
+
+ORIGINAL_BALANCER_STATE="$(get_balancer_state)" || {
```
leon-ape: **P1:** If the balancer starts enabled, this action disables it, unloads the RegionServer, and then fails to re-enable it, released KubeBlocks requeues `memberLeave`. The retry reads the action-mutated `false` value here as the original state and can succeed without restoring it, so scale-in can complete while silently leaving cluster-wide balancing disabled.

baojun-zhang: Reworked the plan here to remove balancer coordination from `memberLeave` entirely.

The updated approach is:
1. keep `lifecycleActions.memberLeave` as the planned-leave gate;
2. run only `RegionMover -o unload` for `KB_LEAVE_MEMBER_POD_FQDN`;
3. propagate any unload failure so KubeBlocks keeps the replica and retries the leave;
4. stop issuing `balance_switch false/true` in this hook.

This removes the retry hazard you pointed out: `memberLeave` no longer mutates cluster-wide balancer state, so a retried action cannot reinterpret an action-mutated `false` value as the original state and silently complete with balancing left disabled.

I also updated the regression coverage to assert that `hregionserver-member-leave.sh` does not issue balancer shell commands and still returns non-zero when `RegionMover` fails.

addons/hbase/scripts/hregionserver-member-leave.sh
```diff
@@ -0,0 +1,15 @@
...
+
+REGIONSERVER_HOST="${KB_LEAVE_MEMBER_POD_FQDN:-${KB_LEAVE_MEMBER_POD_NAME:-}}"
+if [[ -z "${REGIONSERVER_HOST}" ]]; then
+  echo "KB_LEAVE_MEMBER_POD_FQDN or KB_LEAVE_MEMBER_POD_NAME is required" >&2
+  exit 1
+fi
+
+# ponytail: memberLeave only does RegionMover unload; add balancer coordination later only if unload alone proves insufficient in production.
+echo "Unloading regions from ${REGIONSERVER_HOST}..."
+"${HBASE_HOME}/bin/hbase" org.apache.hadoop.hbase.util.RegionMover -m 6 -r "${REGIONSERVER_HOST}" -o unload
```
leon-ape: **P1 — memberLeave can finish while the balancer returns a region to the server.** The hook now calls RegionMover while the RegionServer remains eligible for normal balancing. HBase 2.5.6 RegionMover removes every region already recorded in `movedRegions` before checking whether the source is empty, so if the active balancer assigns a previously moved region back to this server, RegionMover can filter it out and return success with the server still nonempty. Core can then remove a region-bearing Pod, turning planned drainage into avoidable unplanned reassignment and serving disruption.

baojun-zhang: The fix  is  a `draining/decommission + RegionMover + empty-check` flow for planned leave:
1. mark the leaving RegionServer as draining/decommissioned first so it stops receiving new assignments,
2. run `RegionMover -o unload`,
3. verify the source server is actually empty before allowing `memberLeave` to succeed,
4. if unload or the empty-check fails, recommission the server so retries do not leave it stuck in a drained state.

This keeps the protection scoped to the leaving RegionServer and closes the race where balancing can assign a previously moved region back to the source during unload.

addons/hbase/scripts/hregionserver-member-leave.sh
```diff
@@ -0,0 +1,179 @@
...
+trap 'cleanup_on_exit $?' EXIT
+
+REGIONSERVER_SERVER_NAME="$(get_regionserver_server_name)" || {
+  echo "Failed to resolve HBase ServerName for ${REGIONSERVER_HOST}:${REGIONSERVER_PORT}" >&2
+  exit 1
+}
+
+# ponytail: use per-server decommissioning instead of global balancer toggles; revisit only if HBase admin APIs prove insufficient in production.
+echo "Decommissioning ${REGIONSERVER_SERVER_NAME} before unload..."
+decommission_regionserver "${REGIONSERVER_SERVER_NAME}"
```
leon-ape, line 171: **P1 — the 60-second action deadline can leave this server decommissioned while RegionMover keeps running.** The ComponentDefinition requests 300 seconds, but released KubeBlocks v1.1.0-beta.8 caps every positive ExecAction timeout at 60 seconds. After this remote state change, the script starts RegionMover without an internal bound; HBase defaults it to `Integer.MAX_VALUE`. At the deadline, `exec.CommandContext` kills only the parent Bash, so SIGKILL skips the EXIT recommission trap while the exec'd Java RegionMover child can survive and retain the action pipes. A normal slow drain can therefore leave the live RegionServer excluded from assignments and the operation hung or timed out while an unmanaged mover continues, letting cancellation or retry overlap region movement and disrupt serving.

baojun-zhang: Thanks, this is a valid concern. I checked the actual KubeBlocks v1.1.0-beta.8 implementation and the risk is real: positive lifecycle action timeouts are capped at 60s, so our current `timeoutSeconds: 300` does not take effect there. In the current flow we decommission first, then run RegionMover, and the cleanup depends on the shell EXIT trap. If the action hits the platform timeout, that trap is no longer something we should rely on, which can leave the server decommissioned while the unload path is still in flight.\n\nWe will adjust this along two lines: set the lifecycle timeout to `-1` so memberLeave is not cut off by the platform cap, and add an explicit internal timeout/bounded cleanup path around RegionMover so the decommission state is always reconciled by our own logic rather than by an outer action timeout. This keeps the draining/decommission + RegionMover + empty-check approach, while removing the timeout gap you pointed out.

PR #3101, opened 2026-07-07
feat: retain latest mysql minor versions

xuriwuyun:
## Summary
- keep only the latest two MySQL minor/patch releases per major version in the addon ComponentVersion templates
- update default MySQL 8.0/8.4 ComponentDefinition serviceVersion values
- refresh README supported-version examples

## Validation
- helm template mysql addons/mysql


leon-ape (commented):
Findings:
- This removes previously supported MySQL serviceVersions such as 8.0.33-8.0.44 and 8.4.0-8.4.8 from ComponentVersion and the cluster chart schema; this is a user-visible addon behavior change.
- This changes the default MySQL 8.0/8.4 serviceVersions to 8.0.46 and 8.4.10, but the PR validation only includes `helm template mysql addons/mysql` and has no provision or service-connectivity evidence for the new defaults.
- `pr-label-check` is failing because the PR has no pick/nopick/pick-* label.

shanshanying (approved):

PR #3110, opened 2026-07-08
fix(mongodb): route PBM dataprotection through syncer

gnolong:
## Summary

This PR routes MongoDB PBM physical backup, physical restore, and PITR addon flows through the syncer dataprotection API for both sharding and replicaset topologies.

The current head also addresses the endpoint and termination review findings:

- prefer KubeBlocks' API-provided `DP_DB_HOST`; only synthesize conventional Pod DNS when that value is absent
- consume the mongos host and port directly through `serviceVarRef` for config-server and shard components instead of reconstructing a Service FQDN
- bound backup and restore status polling with positive attempt budgets (`720` attempts at 5 seconds for backup; `7200` attempts at 1 second for restore), plus a validated per-request wall-clock timeout (`30` seconds by default, configurable in `1..300`)
- run each status request in an isolated process group; on timeout, terminate and reap the shell wrapper and `syncerctl` child with bounded TERM-to-KILL escalation, and fail a timed-out restore immediately instead of multiplying the timeout across its attempt budget
- preserve only a non-sensitive PVC token in restore coordination; raw PBM storage configuration remains in a mode-0600 PVC file
- share the PBM backup/restore/PITR wrappers between replicaset and sharded ActionSets

## Exact-head validation

Addon head: `39263f1d429b29494625c5d5ff94d9cd6aea8696`

- review-regression ShellSpec on Bash 3.2: 21 examples, 0 failures
- review-regression ShellSpec on Bash 5.3: 21 examples, 0 failures
- all MongoDB ShellSpec on Bash 3.2: 26 examples, 0 failures
- `helm lint addons/mongodb`: 1 chart, 0 failures
- error-severity ShellCheck for the changed script/specs: clean
- Bash 3.2 and Bash 5.3 syntax checks: clean
- `git diff --check`: clean

The regression probes failed on the reviewed behaviors before each correction:

- `DP_DB_HOST` was replaced by conventional headless DNS when `DP_TARGET_POD_NAME` was present
- rendered config-server and shard definitions retained `MONGOS_INTERNAL_SVC_NAME` plus a synthesized host value
- a persistently running backup exceeded an external 1-second test timeout instead of terminating
- the prior attempt-budget guard accepted an oversized decimal that Bash integer comparison could not evaluate; the new backup and restore entry tests both fail it before the first status call
- a directed mutation that changed both config-server and shard endpoint references away from the rendered mongos ComponentDefinition now fails the tuple regression spec
- the historical V3 19-case spec against exact parent
  `a96a12504886c5ab054be8b581b1a430db629dac` fails 19 examples / 5 cases:
  both simulated two-second status calls exceed the one-second budget without
  interruption, invalid timeout values still reach three status calls, and the
  timeout validator is absent
- deleting the KILL escalation fails because the TERM-ignoring leader and
  descendant exceed the six-second bound
- replacing process-group signals with leader-only signals fails because both
  the descendant PID and process group remain observable
- widening the request-timeout ceiling to 999 fails because 301 must be
  rejected before the first status call
- recording an empty PGID fails the explicit canonical-identity and
  dedicated-group assertions without signaling the ShellSpec runner
- removing job-control isolation fails three cases and reports the shared
  runner PGID without signaling the runner
- current 21-case spec against parent
  `a96a12504886c5ab054be8b581b1a430db629dac` exits normally at 21 examples /
  7 failures; the earlier 19/5 result is retained only as the historical V3
  receipt

## Runtime and image boundary

Prior syncer-test matrix evidence is available at [`apecloud/syncer-test@6e75c5c`](https://github.com/apecloud/syncer-test/tree/6e75c5c/evidence/pr-164/matrix), but it was accumulated on earlier addon revisions and custom PR #164 syncer images. It is not exact-head runtime validation for this addon commit.

The packaged chart still renders `apecloud/syncer:0.6.7`. The required PBM dataprotection implementation remains in open, review-blocked [`apecloud/syncer#164`](https://github.com/apecloud/syncer/pull/164) at `871190b08e76eaa5e14639400de88e1d39c08ab8`; no released syncer semantic version contains that implementation. Therefore this PR is not merge/release ready until syncer #164 is released and the chart is pinned to that released version, followed by exact addon-head runtime validation.


addons/mongodb/templates/cmpd-config-server.yaml
```diff
@@ -330,22 +357,11 @@ spec:
           port:
             name: mongodb
     - name: MONGOS_INTERNAL_SVC_NAME
-      valueFrom:
-        serviceVarRef:
-          name: internal
-          compDef: {{ include "mongos.compDefName" . }}
-          optional: false
-          host: Required
+      value: "$(CLUSTER_NAME)-mongos-internal"
```
leon-ape: This replaces the existing serviceVarRef with a hard-coded Service name, port, namespace suffix, and cluster domain. The rendered endpoint is now tied to KubeBlocks naming conventions instead of the actual internal Service selected by the API.

addons/mongodb/dataprotection/common-scripts.sh
```diff
@@ -212,307 +156,190 @@ backup:
...
 
-function print_pbm_logs_by_event() {
-  local pbm_event=$1
-  # echo "INFO: Printing PBM logs by event: $pbm_event"
-  # shellcheck disable=SC2328
-  local pbm_logs=$(pbm logs -e $pbm_event --tail 200 --mongodb-uri "$PBM_MONGODB_URI" > /dev/null)
-  local purged_logs=$(echo "$pbm_logs" | awk -v start="$PBM_LOGS_START_TIME" '$1 >= start')
-  if [ -z "$purged_logs" ]; then
+function target_syncer_host() {
+  if [ -n "${DP_DB_HOST:-}" ] && [ -z "${DP_TARGET_POD_NAME:-}" ]; then
```
leon-ape: DP_DB_HOST is already populated by KubeBlocks from the selected target Pod's actual hostname/subdomain. Because backup jobs also always set DP_TARGET_POD_NAME, this branch discards that API-provided address and reconstructs it from the `<component>-headless` naming convention, so custom or changed Pod subdomains make every syncer backup target the wrong host.

weicao:
Exact-head review blocker on `e7fb7cf82f68fe6e883e0cb550aabad6f00d80c2`: the new polling change can leave MongoDB backup/restore ActionSet Jobs nonterminal forever.

Observed source path:
- `wait_for_syncer_backup_completion` and `wait_for_syncer_restore_completion` removed their configurable `max_retries` / `attempt` ceilings and now have only `while true`.
- restore treats every nonzero `syncerctl restore status` result (bad/expired request id, auth, endpoint, transport) as `Waiting`, then retries forever.
- `found=false`, unknown/empty status, and persistent `running` also have no terminal path.
- the eight MongoDB DP ActionSet Job actions do not provide an `activeDeadlineSeconds`/timeout fallback. The new ShellSpec covers only `running -> done`, not persistent-running or permanent-error paths.

Deterministic A/B:
- prior `056a5cd...`, with the retry ceiling set to zero, exits `rc=1` with `did not complete after 0 retries`;
- current `e7fb7cf...`, with the same always-running status, continues until an external timeout kills it with `rc=124`.

This also matters with syncer PR164 `40554ff7...`: terminal restore cleanup may make a later status request return not-found/nonzero, which the current addon path classifies as waiting forever.

Please keep a bounded, diagnosable terminal path before runtime acceptance. Either:
1. restore the existing configurable finite ceilings (increase defaults if required), classify persistent status-command errors, and exit 1 diagnostically; or
2. add a real rendered Job/controller deadline plus terminal error classification.

In both cases, add tests for persistent-running and permanent status-error/not-found, not only eventual completion.

Focused artifact: `artifacts/pr164-pr3110-head-delta-2026-07-15-1816.md`, SHA256 `46e547fbaf6c81202e6cf92f3f749f829deb4e24583ac08032232f83e89d74d5`.

Boundary: source/script review only; runtime remains `N=0 / NOT_PROVEN`, and evidence from the previous heads does not migrate.

addons/mongodb/dataprotection/common-scripts.sh
```diff
@@ -212,307 +146,284 @@ backup:
...
+  export RESTORE_STORAGE_CONFIG_FILE RESTORE_STORAGE_CONFIG_TOKEN RESTORE_REQUEST_ACCEPTED RESTORE_COMPLETED
+}
 
-  if [ $exit_code -ne 0 ]; then
-    print_pbm_tail_logs
+function wait_for_syncer_backup_completion() {
+  local backup_name=$1
+  local retry_interval=${SYNCER_PBM_WAIT_INTERVAL_SECONDS:-5}
+  describe_result=""
+  while true; do
```
leon-ape, line 294: These polling loops have no finite ceiling and the rendered ActionSet Jobs have no deadline fallback. A persistent `running`, `found=false`, empty status, or restore status-command error leaves the backup/restore Job nonterminal forever; the current tests cover only eventual completion.

addons/mongodb/templates/cmpd-config-server.yaml
```diff
@@ -329,23 +339,10 @@ spec:
...
     - name: MONGOS_INTERNAL_PORT
-      valueFrom:
-        serviceVarRef:
-          name: internal
-          compDef: {{ include "mongos.compDefName" . }}
-          optional: false
-          port:
-            name: mongos
+      value: "27018"
     - name: MONGOS_INTERNAL_HOST
```
leon-ape, line 348: The target release-1.1 API already provides this endpoint through `serviceVarRef`, but the current head still hard-codes the Service name, port, namespace suffix, and cluster domain. Custom or changed internal Service definitions leave the config-server and shard components pointing at the wrong mongos endpoint.

weicao:
Follow-up on the finite-poll blocker: the design has now passed an independent review, including exact-image probes on the three PBM amd64 images used by this addon.

Approved minimum implementation scope:
- only `addons/mongodb/templates/dataprotection/common-scripts.sh` plus focused ShellSpec;
- finite per-status-call deadline and finite observation/error budgets;
- direct ownership of deadline/grace PIDs;
- one shared bounded TERM->KILL->reap cleanup path for external signals and ordinary `EXIT`;
- preserve original exit authority (`set -e` rc1, explicit rc42, fatal USR1 rc138; TERM/INT/HUP 143/130/129);
- container-local diagnostics with all JSON string values scrubbed, and no PVC writes.

Offline exact-image design probes pass: direct N=3, signal N=18, EXIT N=9, combined foreground N=27, with CLEAN only after the complete PID tree is absent and zero residue. This is design/offline evidence only; product code and runtime remain unproven.

@gnolong, please choose one ownership path explicitly so we do not race the same PR:
1. you implement the approved bounded design on this PR head and request a new exact-head review; or
2. transfer the scoped implementation to me, in which case I will use a separate branch/PR and keep this PR untouched.

No one should start runtime acceptance until a new exact head passes bash syntax/ShellCheck, targeted and full ShellSpec, Helm render, the three-image N=3/N=27 probes, diagnostic canaries, and mount/injection gates.

gnolong:
Companion syncer validation is complete and published for the current addon head.

```text
addon:  e7fb7cf82f68fe6e883e0cb550aabad6f00d80c2
syncer: 162bce18d09f360cb4fb169e409544f4d35a5090
image:  apecloud/syncer:mongodb-pbm-pr164-162bce1
image ID: sha256:2f916721fe12e242c85a710e961a70cd2d8bac24b42f40e09f51ff0dea825c3d
```

[Full matrix report and sanitized logs](https://github.com/apecloud/syncer-test/tree/edcd4de/evidence/pr-164/matrix)

Executed results cover ReplicaSet and sharded full/PITR on MongoDB `8.0.17`, `7.0.28`, `6.0.27`, `5.0.29`, and `4.4.29`, with PBM `2.12`, `2.10`, and legacy `2.5`. Full restores verified documents, catalog data, metadata, and index readback. PITR verified `batch1` present and `batch2` absent. MongoDB `4.0.28` remains excluded because PBM 2.5 rejects it as unsupported.

The exact final-head sharded 8.0.17 full restore passed in `309.18s`, including the coordinated mongod/mongos topology-clock reset barrier. The exact final-head 5.0.29 forced failure-recovery row passed in `143.73s`, with stable `phase=failed`, both members `cleaned`, empty maintenance, and successful pre/post-failure writes.

On the polling review concern: the lack of an addon-wide elapsed-time ceiling is intentional. The addon job follows the backend PBM operation to a terminal state instead of declaring a long valid backup/restore failed and potentially orphaning it. Syncer terminal coordination is retained after token and leader cleanup, so a normal late status query does not become not-found. Transient status command errors are retried; backend `done` and `failed` remain authoritative. This preserves the requested no-wall-clock-cap contract rather than reintroducing the old one-hour/two-hour limits.

All current addon checks are green and the PR is mergeable. Credential scans over the published evidence found no literal credential-bearing MongoDB URI. External review approval remains required.

weicao:
Thanks for the exact-head syncer packet. I checked the published matrix and the cited final syncer head. It closes happy-path compatibility and one forced backend-failure recovery row, but it does not close the addon polling liveness blocker.

Facts on the unchanged addon head `e7fb7cf82f68fe6e883e0cb550aabad6f00d80c2`:

1. `wait_for_syncer_restore_completion` logs every nonzero status-query result and then sleeps/retries forever; there is no consecutive observation-error stop condition. Valid rc0 responses with an unknown/empty nonterminal status also have no stop condition.
2. Both backup and restore execute `syncerctl ... status` without a per-call deadline. At syncer head `162bce18d09f360cb4fb169e409544f4d35a5090`, `ctl/dataprotection.go` passes `context.TODO()`, while `client/httpclient.go` configures only dial/TLS-handshake timeouts and leaves `http.Client.Timeout` unset. A connected request whose headers/body never complete can therefore hold the ActionSet Job indefinitely.
3. The matrix README states that only the sharded 8.0.17 full row and the 5.0.29 forced failure-recovery row are exact-final-head runs. Neither row injects a hung status response, persistent transport/auth failure, rc0 malformed/unknown response, or missing request/op identity, so the packet cannot falsify items 1-2.
4. The release-1.1 addon contract requires negative DataProtection paths to produce a diagnostic error and a stop condition (`docs/addon-api-1.1/12-minimum-acceptance.md:42`); the current loops do not provide one for these observation failures.

I agree that a valid, observable backend `running` state need not be failed by an arbitrary addon-wide wall-clock ceiling. That policy does not require unbounded status calls or unbounded observation failures. The minimal compatible correction is therefore narrower than a blanket operation deadline:

- bound each `syncerctl ... status` call, including TERM/grace/KILL/reap cleanup;
- bound consecutive status observation failures (nonzero, timeout, malformed/invalid response) with a sanitized diagnostic;
- fail explicit terminal `failed/error` and invalid/missing operation identity deterministically;
- reset the observation-error counter after a valid observable nonterminal response;
- keep a valid `found=true,status=running` / valid restore nonterminal phase free of an addon-wide elapsed-time cap if that is the intended product policy;
- add permanent-error, hung-call, malformed/unknown, transient-recovery, signal-cleanup, and PID-residue tests.

So the runtime packet is useful and the global-cap concern is understood, but the exact addon head remains BLOCKED on bounded observability/cleanup. No ownership transfer is implied; the existing PR author can implement this narrow correction on PR3110.

gnolong:
MongoDB 4.0 PBM policy is now explicit: MongoDB `4.0` remains deployable, but PBM-backed backup/restore/PITR in this addon requires MongoDB `4.4` or newer.

PBM `2.5.0` rejected the executed MongoDB `4.0.28` row with `Unsupported MongoDB version. PBM works with v4.4, v5.0, v6.0, v7.0`. Upstream tag `v2.5.0` confirms this in `pbm/version/version.go`; its entry-level support gate also excludes `4.2`.

The two protected BackupPolicyTemplates remain unchanged. KubeBlocks 1.1 does not support a service-version predicate on `BackupMethodTPL`: `versionMapping` resolves env values only, and removing `4.0` from a mapping leaves the method present with the ActionSet default image. Such a change would look restrictive while changing no runtime behavior.

Accordingly, this PR does not claim MongoDB 4.0/4.2 PBM support; unsupported attempts fail closed at PBM's version gate. Conditionally hiding methods needs a future KubeBlocks API capability or separate version-specific templates.

addons/mongodb/values.yaml
```diff
@@ -12,9 +12,6 @@ image:
     backup:
       repository: apecloud/percona-backup-mongodb
       tag: 2.9.1
-  kubectl:
-    repository: apecloud/kubectl
-    tag: 1.29
   syncer:
     repository: apecloud/syncer
     tag: "0.6.7"
```
leon-ape, line 17: The packaged addon still renders `apecloud/syncer:0.6.7`, while all submitted runtime validation uses the custom PR164 image `apecloud/syncer:mongodb-pbm-pr164-162bce1` and PR164 is still open. The published chart therefore does not contain or validate the syncer implementation exercised by these new PBM dataprotection calls.

leon-ape (commented):
Current head `c038b59b` still has the open blockers: `target_syncer_host` discards the KubeBlocks-provided `DP_DB_HOST` whenever `DP_TARGET_POD_NAME` is present; backup and restore status polling remain unbounded with no per-call or job deadline; and the release chart still renders `apecloud/syncer:0.6.7` while the submitted runtime evidence uses unreleased custom PR images and does not cover this exact head.

addons/mongodb/templates/cmpd-config-server.yaml
weicao: Fixed at exact head `89867e9b49da0635c1f3d5cd5f29595d485536ce`. Config-server now consumes both `MONGOS_INTERNAL_HOST` and `MONGOS_INTERNAL_PORT` directly from the `internal` mongos `serviceVarRef`; the intermediate Service-name variable and reconstructed namespace/domain FQDN are removed. The same correction is applied to the shard definition. A render-level regression spec verifies both ComponentDefinitions and passes on Bash 3.2 and 5.3.

addons/mongodb/dataprotection/common-scripts.sh
weicao: Fixed at exact head `89867e9b49da0635c1f3d5cd5f29595d485536ce`. `target_syncer_host` now returns nonempty `DP_DB_HOST` regardless of `DP_TARGET_POD_NAME`; conventional headless DNS is only the compatibility fallback when the API host is absent. The focused parent probe reproduced the wrong conventional DNS, while the corrected spec passes the custom-subdomain case on Bash 3.2 and 5.3.

addons/mongodb/dataprotection/common-scripts.sh
weicao: Fixed at exact head `89867e9b49da0635c1f3d5cd5f29595d485536ce`. Backup polling now has a validated positive ceiling of 720 attempts at 5 seconds; restore has 7200 attempts at 1 second. Both return failure at the ceiling, so the enclosing `set -e` ActionSet script terminates. Regression coverage includes eventual completion, persistent backup running/found=false/empty, persistent restore running/status-command error, and invalid budgets. Focused specs pass 11/0 on Bash 3.2 and 5.3; the parent persistent-running probe was externally killed with `rc=124`.

addons/mongodb/templates/cmpd-config-server.yaml
weicao: Fixed at exact head `89867e9b49da0635c1f3d5cd5f29595d485536ce`. Both config-server and shard definitions now bind `MONGOS_INTERNAL_HOST` directly with `serviceVarRef.name=internal`, mongos `compDef`, and `host: Required`; the port remains a direct `serviceVarRef` to the mongos port. The reconstructed namespace/cluster-domain endpoint is gone. The render-level spec verifies this contract for both definitions.

addons/mongodb/values.yaml
weicao: Confirmed unresolved at addon exact head `89867e9b49da0635c1f3d5cd5f29595d485536ce`. The chart still renders `apecloud/syncer:0.6.7`; required PBM API code is only in open, review-blocked syncer PR #164 at `871190b08e76eaa5e14639400de88e1d39c08ab8`, and no released semantic version contains it. I did not replace the chart tag with a custom PR image or claim the accumulated custom-image matrix as exact-head release evidence. The PR body now marks this as a merge/release blocker: syncer #164 must be released, the chart pinned to that release, and the exact addon head rerun.

addons/mongodb/dataprotection/common-scripts.sh
leon-ape: Finding: The attempt ceiling does not bound a single syncerctl status call. syncer #164 uses context.TODO() and an HTTP client without a request timeout, so an accepted connection can block indefinitely before the loop increments; the backup/restore Job can still remain nonterminal.

weicao: Follow-up hardening at exact head `a96a12504886c5ab054be8b581b1a430db629dac`: attempt budgets are now restricted before the first status call to canonical integers in the explicit shell-safe range `1..9999999`. This closes the oversized-decimal case that made Bash `[` emit `integer expression expected` and bypass the ceiling. Backup and restore each have invalid and oversized entry tests (mock call count remains zero), plus the upper boundary is accepted. Focused correction specs now pass 15/0 on Bash 3.2 and 5.3; all MongoDB specs pass 19/0.

addons/mongodb/templates/cmpd-config-server.yaml
weicao: Follow-up test hardening at exact head `a96a12504886c5ab054be8b581b1a430db629dac`: the render regression now resolves the exact rendered mongos ComponentDefinition and, for config-server and shard and both host/port variables, locks the full `serviceVarRef` tuple (`name: internal`, exact mongos `compDef`, `optional:false`, no direct value), plus `host:Required` / `port.name:mongos`. A directed mutation that points config-server refs to config-server and shard refs to shard now fails 1/1 with `MONGOS_INTERNAL_HOST wrong compDef`; the corrected render passes on Bash 3.2 and 5.3.

addons/mongodb/dataprotection/common-scripts.sh
leon-ape: Finding remains: this validates only the number of loop iterations. It does not bound a single syncerctl status request; syncer #164 still uses context.TODO() and an HTTP client without a request timeout, so the loop can block before incrementing the validated counter.

weicao: Fixed at exact head `3669442e72eb2ad7fbe7687e7ae3582207ae92ba`. Each backup/restore status request now runs in its own process group under a validated wall-clock budget (`SYNCER_STATUS_REQUEST_TIMEOUT_SECONDS`, default 30s, allowed 1..300); timeout reaps the shell wrapper and `syncerctl` child with bounded TERM-to-KILL escalation. Backup and restore both terminate after the first timed-out request, so the 7,200-attempt restore ceiling cannot multiply a hung request. Exact parent `a96a1250` with the new spec is RED 19 examples/5 failures; the corrected head is GREEN 19/0 on Bash 3.2 and 5.3, all MongoDB specs 24/0, Helm lint 1/0, and changed-file ShellCheck/syntax/diff checks are clean. This closes the addon-side nonterminal path; the unreleased syncer #164 image remains a separate merge/runtime blocker.

weicao: Follow-up pushed at exact head `e8913910a288643860c840d004c4bac04d155403`.

The production timeout wrapper is unchanged. This test-only commit closes the
two regression-contract gaps found in the fresh design review:

- the timeout probe now has both the process-group leader and descendant ignore
  TERM, records the descendant PID and PGID, and asserts bounded return, the
  rc-124 diagnostic, and removal of both the descendant and process group;
- timeout value 301 is rejected with the exact `1..300` error before any status
  request.

Directed mutation evidence on Bash 3.2:

- remove KILL: `21 examples / 1 failure` (exceeds the six-second bound);
- signal only the leader: `21 / 1` (descendant PID and PGID remain observable);
- widen the ceiling to 999: `21 / 1` (301 is accepted and performs three
  forbidden status calls).

Exact-head validation:

- focused Bash 3.2: `21 / 0`
- focused Bash 5.3: `21 / 0`
- all MongoDB ShellSpec on Bash 3.2: `26 / 0`
- Helm lint, error-severity ShellCheck, Bash 3.2/5.3 syntax, and
  `git diff --check`: clean

The release/runtime boundary is unchanged: the chart still renders syncer
`0.6.7`, while required syncer PR #164 exact
`871190b08e76eaa5e14639400de88e1d39c08ab8` remains open and unreleased.

weicao: Runner-safe regression follow-up pushed at exact head
`39263f1d429b29494625c5d5ff94d9cd6aea8696`.

The production status-timeout script is byte-identical to production head
`3669442e72eb2ad7fbe7687e7ae3582207ae92ba`. This test-only follow-up:

- captures the ShellSpec runner PGID;
- requires canonical child PID, child PGID, and runner PGID;
- requires the child PGID to differ from the runner PGID;
- prohibits process-group cleanup for invalid or shared identities;
- permits direct child fallback only when the live PID still resolves to the
  recorded dedicated PGID;
- explicitly asserts valid recorded identity and dedicated process-group
  isolation.

Bash 3.2 RED receipts all exit through ShellSpec rather than SIGKILL:

- final 21-case spec against parent `a96a1250...`: `21 / 7`
- empty PGID: `21 / 1`
- remove `set -m`: `21 / 3`
- remove KILL: `21 / 1`
- signal leader only: `21 / 1`
- widen timeout ceiling to 999: `21 / 1`

Exact-head GREEN remains focused Bash 3.2 `21 / 0`, focused Bash 5.3
`21 / 0`, all MongoDB ShellSpec `26 / 0`, with Helm lint,
error-severity ShellCheck, dual-shell syntax, and diff-check clean.

The earlier `19 / 5` parent result is a historical V3-spec receipt; the current
21-case parent receipt is the normal `21 / 7` result above.

The release boundary is unchanged: the chart still renders syncer `0.6.7`,
while required syncer PR #164 exact `871190b0...` remains open and unreleased.

weicao (changes requested):
Current-head release dependency review on `39263f1d429b29494625c5d5ff94d9cd6aea8696`: changes requested.

Direct evidence:
- `addons/mongodb/values.yaml` still renders `apecloud/syncer:0.6.7`.
- required syncer PR apecloud/syncer#164 is still OPEN / BLOCKED / REVIEW_REQUIRED at `871190b08e76eaa5e14639400de88e1d39c08ab8`; no released semantic version containing that implementation is available from this dependency chain.
- the submitted MongoDB PBM runtime matrix used custom PR images, so it does not validate the release chart rendered by this addon head.

Pre-flight contract gates:
- root-cause/release scope: the addon introduces PBM-through-syncer calls, but the published dependency does not contain the exercised implementation;
- release-1.1 evidence boundary: source/unit/custom-image evidence cannot prove the released chart path;
- current-head boundary: accumulated custom-image runs do not migrate to this release artifact.

Design-contract classes 1–8, scoped to this release-dependency comment:
1. **Blocker** — the release chart silently retains the old `0.6.7` image instead of enforcing the required implementation.
2. No separate empty-field issue; the tag is non-empty but contract-incompatible.
3. **Blocker** — commit/artifact discontinuity: addon head, syncer PR head, and tested image are not one released exact state.
4. No sentinel-value issue found in this dependency binding.
5. No conditional cleanup pre-state issue introduced by the values binding.
6. No NotFound short-circuit issue in this dependency binding.
7. No terminating-vs-absent decision in this dependency binding.
8. No operator-precedence expression in this dependency binding.

Required closure:
1. merge and release syncer #164 (or an equivalent reviewed implementation) as an official version;
2. pin this chart to that released version, not a custom PR image;
3. rerun the relevant backup/restore/PITR and negative-path acceptance on the resulting exact addon head and released image, with exact imageID evidence.

Until those three steps close, this PR is not release-mergeable. I did not modify the human-owned PR branch or substitute a custom image.

weicao:
Dependency blocker confirmed on exact addon head `39263f1d429b29494625c5d5ff94d9cd6aea8696`.

Facts:

- `addons/mongodb/values.yaml` still renders `docker.io/apecloud/syncer:0.6.7` (published index digest `sha256:832ab6140d17c425d5df751ce44a4bf4d33043c68d00014237ff8e8ed79a7d7e`).
- The MongoDB Backup/Restore/PITR implementation validated with this PR is in syncer PR apecloud/syncer#164, current exact head `871190b08e76eaa5e14639400de88e1d39c08ab8`; that PR is still OPEN, MERGEABLE, REVIEW_REQUIRED.
- The newer published `docker.io/apecloud/syncer:0.7.7` is not a substitute: its index digest is `sha256:80a9bbfafe420035b9512d8a61277fe5fb465551929663d6014f6d2c270c1626`, embedded revision is `6f6af72f9a69dc8f4a4bf93524366d2d5964b0be`, and PR164 head is not an ancestor of that revision.
- The custom validation image therefore proves only the pinned PR candidate; it is not a released chart dependency.

Resolution order:

1. finish human review and merge syncer PR164;
2. publish a semantic syncer image from a merged revision containing exact PR164 functionality, with immutable multi-arch digest and embedded revision;
3. update this release-1.1 chart from `0.6.7` to that released tag;
4. add a render contract that asserts every MongoDB ComponentVersion/init-syncer reference uses the released tag, then rerun focused/full ShellSpec and Helm gates on the new exact addon head.

I am not pinning the release chart to `mongodb-pbm-pr164-*` or another custom PR image. Until steps 1-3 exist, this PR remains dependency-blocked and makes no release/runtime-ready claim.

weicao (changes requested):
P1 recheck on exact addon head `39263f1d429b29494625c5d5ff94d9cd6aea8696`: the release dependency and runtime closure are still missing.

Current direct evidence:
- `addons/mongodb/values.yaml` still pins `apecloud/syncer:0.6.7`.
- required `apecloud/syncer#164` is still OPEN / REVIEW_REQUIRED (current head `6859b1d0da3e09274cc4696862ec778d793ae9dd`), so the PBM API implementation is not present in an official released Syncer selected by this chart.
- the addon PR head has not changed since the prior requested-changes review; there is still no exact-head acceptance result using an official released image selected by the chart.

Therefore the prior release-contract blockers remain: class 1 old-image fallback and class 3 artifact/current-head discontinuity. CI success does not close either blocker.

Required closure remains: merge/release the Syncer PBM dependency, pin this chart to that official release, then provide exact addon-head + released-image backup/restore/PITR and negative-path terminal evidence with exact imageID. Until then this PR remains not release-mergeable.

weicao (changes requested):
P1 dependency recheck on exact addon head `39263f1d429b29494625c5d5ff94d9cd6aea8696`: still blocked.

Direct current evidence:
- the chart still selects `apecloud/syncer:0.6.7` in `addons/mongodb/values.yaml`;
- required `apecloud/syncer#164` has advanced to `51f3b0c16c93d45bd68b367cca89f379285faf2a`, but remains OPEN / REVIEW_REQUIRED and is not an official released Syncer selected by this chart;
- no new addon head or exact-head released-image backup/restore/PITR plus negative-path terminal evidence is present.

Pre-flight/current-head boundary: green CI validates the submitted source/chart checks, but does not prove the missing release dependency or its runtime contract. Classes 1–8 were rechecked for this dependency boundary: class 1 old-image fallback and class 3 artifact/commit discontinuity remain blockers; classes 2 and 4–8 add no separate blocker in the image binding itself.

Required closure is unchanged: merge and officially release the Syncer PBM API dependency, pin this chart to that release, then provide exact addon-head + selected released-image acceptance terminals (backup/restore/PITR and negative paths) with exact imageID. Until all three close, this PR remains not release-mergeable.

weicao:
# PR3110 exact-head dependency recheck (2026-08-04)

## Result

`BLOCKED`: addon PR3110 still selects a released Syncer artifact that does not
contain the PBM API implementation exercised by the PR's runtime evidence.

This is a release dependency/current-artifact discontinuity, not an addon CI
failure and not a new runtime product failure.

## Exact identities

- Addon PR: `apecloud/kubeblocks-addons#3110`
- Addon head: `39263f1d429b29494625c5d5ff94d9cd6aea8696`
- Addon base branch current tip: `release-1.1@1c8380ba8aad98a1282ff176d6d622468d47c41e`
- Addon head merge base: `36eb2a36504c8f1e4dd0f18d98dd583ec6ef8a2a`
- Addon base is 16 commits ahead of that merge base.
- Addon chart selection: `addons/mongodb/values.yaml` ->
  `apecloud/syncer:0.6.7`
- Syncer PR: `apecloud/syncer#164`
- Syncer PR head: `9156dbeb8ee84ff0936258354bab36db6cd94e3a`
- Syncer PR state: `OPEN`, `REVIEW_REQUIRED`; current checks are terminal green.
- Syncer main: `266913b270f501579e2c686dd01ac846ebb03666`

## Current official images

Docker Hub was read directly on 2026-08-04.

### `apecloud/syncer:0.7.8`

- OCI index digest:
  `sha256:bd546bd085fd88811a2d279f214d7462373b8f417e31d534eb5830bfeb241c6d`
- amd64 manifest:
  `sha256:33038b9467fa95fb0bc570c1346bde0da3b0057d354171ba46fa511a647834df`
- arm64 manifest:
  `sha256:12dbb12d006305343b78970fb0cc37d5d500a6c8afd835d32021cf6c4b871e03`
- embedded revision:
  `266913b270f501579e2c686dd01ac846ebb03666` (current Syncer main)

`git merge-base --is-ancestor 9156dbeb... 266913b...` returned rc=1. A
direct `git grep -i 'pbm|percona backup for mongodb' origin/main -- '*.go'`
returned no matches. Therefore `0.7.8` is not a released artifact containing
PR164's PBM implementation.

### `apecloud/syncer:0.7.7`

- OCI index digest:
  `sha256:ee96fcea4543d1f8636a41c4457e7e39b9c539149a0fa52a1df543e805a8572b`
- embedded amd64 revision:
  `fd2f95c269a2c001e05db9a33b236a5b42456ed1`

`git merge-base --is-ancestor 9156dbeb... fd2f95c...` also returned rc=1.

## Runtime and CI boundary

- PR164's newest MongoDB 4.4/PBM 2.5 report is useful candidate evidence:
  8/8 scenarios passed on exact Syncer head `9156dbeb...` and candidate image
  ID `sha256:42cae6a008ca02ce26d12f030812af358294de58cdffbc062cc30e9442ba99ca`.
- That image is not the official image selected by PR3110, so the result does
  not migrate to addon head `39263f1d...` plus chart-selected `0.6.7`.
- PR3110's source/chart checks are terminal: eight pass and four review-trigger
  jobs skip. This closes static CI only; it cannot prove the missing released
  dependency.
- MongoDB runner allowlist is empty under the current STOP directive. No new
  package, runner, Job, retry, or runtime acceptance was started by this audit.

## Required closure

1. Split/review and merge the required Syncer PBM functionality, or merge an
   equivalent independently reviewable implementation.
2. Publish a semantic multi-arch Syncer image whose immutable digest and
   embedded revision contain that merged implementation.
3. Update PR3110 to select that official release and add an exact render
   contract for every MongoDB Syncer reference.
4. Run the permitted exact addon-head plus released-image acceptance and post
   exact imageID/terminal/cleanup evidence.

Until then, changing PR3110 from `0.6.7` to `0.7.7` or `0.7.8`, or pinning a
custom PR image, would not close the dependency contract.

weicao (changes requested):
P1 dependency/current-head recheck remains blocked.

Direct current evidence:
- addon head is unchanged at `39263f1d429b29494625c5d5ff94d9cd6aea8696` and `addons/mongodb/values.yaml` still selects `apecloud/syncer:0.6.7`;
- required `apecloud/syncer#164` has drifted again to `23d171123d59c129978d998afe0dac405a5b2fa8`, but remains OPEN / REVIEW_REQUIRED / CONFLICTING, and its `Push Pre-Check` is failing;
- because that dependency head is neither merged nor officially released and selected by this chart, candidate-branch PBM evidence does not establish the chart's backup/restore/PITR contract;
- no new addon head or exact addon-head + selected released-image backup/restore/PITR and negative-path terminal evidence exists.

Pre-flight/current-head boundary: addon CI success is static evidence only. Classes 1–8 were rechecked for this dependency binding: class 1 old-artifact fallback and class 3 artifact/commit discontinuity remain blockers; classes 2 and 4–8 add no separate blocker in the image binding itself.

Required closure remains: merge/review the PBM Syncer implementation, publish an official immutable multi-arch release containing it, update the chart to select that release, then provide exact addon-head + released-image terminal and cleanup evidence with exact imageID. Until then this PR remains not release-mergeable.

weicao:
# PR3110 dependency gate recheck: Syncer `23d1711` (2026-08-05)

## Result

`BLOCKED`: there is still no released, approved Syncer artifact containing the
PBM backup/restore/PITR implementation selected by addon PR3110.

The current Syncer Push Pre-Check failure is independently classified as CI
infrastructure. It is not a source-test failure, but closing it would still not
close the conflict, reviewability, release, or exact-artifact blockers.

## Exact state

- Addon PR3110 head:
  `39263f1d429b29494625c5d5ff94d9cd6aea8696`
- Addon state: `OPEN`, `MERGEABLE`, `CHANGES_REQUESTED`
- Chart selection: `apecloud/syncer:0.6.7`
- Syncer PR164 head:
  `23d171123d59c129978d998afe0dac405a5b2fa8`
- Syncer state: `OPEN`, `CONFLICTING`, `REVIEW_REQUIRED`
- Syncer main: `266913b270f501579e2c686dd01ac846ebb03666`
- New commits since the prior `9156dbeb...` audit:
  - `aab2c5e` `refactor(mongodb): simplify restore coordination`
  - `23d1711` `fix(mongodb): handle empty cluster role as replicaset`

`git merge-tree --write-tree origin/main HEAD` returned rc=1 with one content
conflict: `config/config.go`. The other named paths auto-merged.

## Push Pre-Check failure classification

- Workflow run: `30911876794`
- Failed job: `92000444488`
- Job URL:
  <https://github.com/apecloud/syncer/actions/runs/30911876794/job/92000444488>
- Failure occurred while GitHub built the pinned `crate-ci/typos@v1.13.14`
  action container, before repository preflight logic ran.
- On each of three build attempts, `apt-get update` could not connect to
  `archive.ubuntu.com:80` and `security.ubuntu.com:80`; the action Dockerfile
  then could not locate `wget`, `git`, or `jq` and exited rc=100.
- Same exact-head workflow results: `make-test` PASS, coverage PASS,
  trigger/label PASS, image check SKIP.

Independent source gates on a clean detached exact-head worktree:

- `git diff --check origin/main...HEAD`: rc=0
- `go test ./...`: rc=0
- `go vet ./...`: rc=0

Thus the hosted failure is infrastructure-only. This classification does not
convert PR164 to approved, conflict-free, merged, or released status.

## Official artifact check

Docker Hub newest official semantic tag remains `apecloud/syncer:0.7.8`:

- OCI index:
  `sha256:bd546bd085fd88811a2d279f214d7462373b8f417e31d534eb5830bfeb241c6d`
- embedded revision:
  `266913b270f501579e2c686dd01ac846ebb03666`

No official image has been published after that tag. Syncer PR head
`23d171123...` is not contained by the official main revision. Therefore
bumping PR3110 from `0.6.7` to `0.7.8` would still select an implementation
without the required PBM surface.

## Runtime boundary

Previously published custom/candidate-image E2E remains useful only for the
exact candidate it exercised. It cannot migrate to addon head `39263f1d...`
plus chart-selected `0.6.7`, or to official `0.7.8`.

The MongoDB runner allowlist is empty under the active STOP directive. This
audit started no package, runner, Job, retry, or runtime acceptance.

## Required closure

1. Resolve/split PR164 into independently reviewable capabilities and close
   the exact `config/config.go` conflict.
2. Obtain review approval and merge the required PBM implementation.
3. Publish a semantic multi-arch image whose immutable digest and embedded
   revision contain the merged implementation.
4. Pin PR3110 to that official release, add exact render contracts, and run
   permitted exact-head released-image acceptance with terminal/cleanup
   evidence.

No addon-side source change is valid before steps 1-3.

weicao (changes requested):
P1 exact-dependency recheck on unchanged addon head `39263f1d429b29494625c5d5ff94d9cd6aea8696`: still blocked.

Current direct evidence:
- `addons/mongodb/values.yaml` still selects released `apecloud/syncer:0.6.7`.
- required `apecloud/syncer#164` advanced from `23d1711...` to `4416b90a50e28b608e0563c11d52c94466a3eadf` via `refactor(dcs): consolidate dp coordination` (10 paths, +1290/-1522), but remains OPEN / REVIEW_REQUIRED / CONFLICTING.
- exact merge-tree against Syncer main `266913b...` still has the content conflict in `config/config.go`.
- clean exact-head gates pass locally: `git diff --check origin/main...HEAD`, `go test ./...`, and `go vet ./...`.
- its Push Pre-Check red is again infrastructure-only: the pinned `crate-ci/typos` container build cannot reach Ubuntu apt mirrors and exits rc100 before repository preflight. That does not close conflict/review/release.
- newest official semantic image remains `syncer:0.7.8`, index `sha256:bd546bd085fd88811a2d279f214d7462373b8f417e31d534eb5830bfeb241c6d`, embedding Syncer main `266913b...`, not PR164.

So class 1 old-artifact fallback and class 3 artifact/current-head discontinuity remain. The new Syncer refactor is source-green but neither merged nor released; no safe addon pin exists. Required closure remains: resolve/split/review/merge the PBM Syncer implementation, publish an official immutable multi-arch release containing it, pin the addon chart to that release with an exact render contract, then run permitted exact-head released-image acceptance and cleanup evidence. Until then PR3110 remains not release-mergeable.

weicao (changes requested):
P1 exact-dependency recheck on unchanged addon head `39263f1d429b29494625c5d5ff94d9cd6aea8696`: still blocked, with a new source-gate red in the required dependency.\n\n- The chart still selects released `apecloud/syncer:0.6.7`.\n- Syncer #164 advanced from `4416b90a...` to `3ed19a41a40d9f048d1090ffc8f69e96b2edb510` via a 52-file PBM service refactor, but remains OPEN / CONFLICTING / REVIEW_REQUIRED.\n- Exact hosted `make-test` now fails `make lint`: `engines/mongodb/config.go:279:6: func normalizedNames is unused`. Exact-head grep finds only its declaration/comments. This is a repository failure; the separate Push Pre-Check red remains Ubuntu apt infrastructure-only.\n- The merge-tree still has the `config/config.go` content conflict. Newest official semantic image remains `syncer:0.7.8` (`sha256:bd546bd...`) and does not contain this PR head.\n\nTherefore class 1 old-artifact fallback and class 3 artifact/current-head discontinuity remain, and the candidate dependency now also has an exact source-lint blocker. Required closure is unchanged: fix and review/split/merge Syncer #164, publish an official immutable multi-arch release containing it, pin this chart to that release, then provide permitted exact addon-head + released-image terminal/cleanup evidence. No safe addon pin exists now.

weicao (changes requested):
P1 exact-dependency recheck on unchanged addon head `39263f1d429b29494625c5d5ff94d9cd6aea8696`: still blocked after another Syncer head drift.

Direct current evidence:
- `addons/mongodb/values.yaml` still selects released `apecloud/syncer:0.6.7`.
- required `apecloud/syncer#164` advanced from `3ed19a41...` to `e2af8c5090f2712e028f16459dd2c6ca265c3a9f`, but remains OPEN / CONFLICTING / REVIEW_REQUIRED and is not an official release selected by this chart.
- exact hosted `make-test` still fails `make lint` at `engines/mongodb/config.go:279:6`: `normalizedNames` is unused (exit 2); `Push Pre-Check` is now green, so the remaining red is repository-source evidence.
- no exact addon-head + chart-selected released-image backup/restore/PITR, negative-path, and cleanup terminal evidence exists.

Pre-flight/current-head boundary: the new dependency commit changes 29 MongoDB PBM/model paths (+1297/-875), so evidence for its prior head cannot be carried forward. Addon static CI does not prove this unreleased runtime dependency.

Classes 1–8 rechecked for the release binding: class 1 old-artifact fallback and class 3 artifact/current-head discontinuity remain blockers; classes 2 and 4–8 add no separate blocker in the image binding itself.

Required closure remains: make Syncer #164 source-green, resolve/review/merge it, publish an official immutable multi-arch release containing the exact implementation, update this chart to select that release, then provide permitted exact addon-head + selected released-image terminal/cleanup evidence with exact imageID. No safe addon pin exists now.

weicao (changes requested):
P1 exact-dependency recheck on unchanged addon head `39263f1d429b29494625c5d5ff94d9cd6aea8696`: still blocked after another material Syncer rewrite.

Direct current evidence:
- `addons/mongodb/values.yaml` still selects released `apecloud/syncer:0.6.7`.
- required `apecloud/syncer#164` advanced from `e2af8c50...` to `5c08be0e00b6114f77b25ba9ae04163b30237eaa`, but remains OPEN / CONFLICTING / REVIEW_REQUIRED and is not an official release selected by this chart.
- exact hosted `make-test` fails `make lint` with three repository-source errors: unused `normalizedNames`, `detachTestDBServices`, and `updateRestoreState` (exit 2).
- `Push Pre-Check` also fails, but its log stops in the checker-image `apt-get` step (exit 100); that infrastructure red does not replace or weaken the direct source-lint failure.
- no exact addon-head + chart-selected released-image backup/restore/PITR, negative-path, and cleanup terminal evidence exists.

Pre-flight/current-head boundary: the new dependency commit changes 30 paths (+2388/-3428), including DCS coordination and Mongo restore state machines, so prior-head evidence cannot carry forward. Addon static CI cannot prove this unreleased runtime dependency.

Classes 1–8 rechecked for the release binding: class 1 old-artifact fallback and class 3 artifact/current-head discontinuity remain blockers; classes 2 and 4–8 add no separate blocker in the image binding itself.

Required closure remains: make Syncer #164 source-green, resolve/review/merge it, publish an official immutable multi-arch release containing the exact implementation, update this chart to select that release, then provide permitted exact addon-head + selected released-image terminal/cleanup evidence with exact imageID. No safe addon pin exists now.

weicao:
Additional required metadata fix: the current title, `fix: route MongoDB PBM dataprotection through syncer`, omits the addon scope required for a single-addon change. Please rename it to `fix(mongodb): route PBM dataprotection through syncer`.

This title-only correction does not close the existing Syncer source/review/release or exact released-image runtime blockers.

PR #3112, opened 2026-07-08
fix(postgresql): guard pg-basebackup restore data directory

### LINKED ISSUE #3111: PostgreSQL pg-basebackup restore should fail fast on missing or misplaced data payload
## Problem

PostgreSQL `pg-basebackup` restore prepareData can currently finish with `done!` even when no supported basebackup artifact is present on the compatibility fallback path. It also does not assert that the restored payload is a PostgreSQL data directory under the runtime data path `/home/postgres/pgdata/pgroot/data`.

That can let the restore controller treat prepareData as successful while the target volume is empty, incomplete, or populated at the wrong path. The next PostgreSQL startup may then initialize a fresh data directory instead of using restored data.

## Constructed test scenario

Add focused ShellSpec coverage for `addons/postgresql/dataprotection/pg-basebackup-restore.sh`:

1. Modern `${DP_BACKUP_NAME}.tar.zst` artifact extracts into `${VOLUME_DATA_DIR}/pgroot/data` and leaves the volume root clean.
2. `DATA_DIR=${VOLUME_DATA_DIR}` fails before pulling data.
3. No supported artifact fails and does not print `done!`.
4. Extracted payload without `PG_VERSION`, `base/`, `global/`, or `pg_wal/` fails.
5. Existing PostgreSQL files at volume root fail with diagnostics.

## Reproducibility

Reproducible with script-level stubs for `datasafed`, `tar`, `gunzip`, and `pg_waldump`; no Kubernetes cluster is needed for the script contract. Runtime restore still needs target-side validation on the fixed vcluster before the PR can be marked ready.

## Expected fix direction

Make the ActionSet expose `VOLUME_DATA_DIR=/home/postgres/pgdata` and `DATA_DIR=$(VOLUME_DATA_DIR)/pgroot/data`, then make `pg-basebackup-restore.sh` fail fast unless the final payload is a PostgreSQL data directory in that exact subdirectory and not at the volume root.

weicao:
## Summary

Fixes #3111.

This PR tightens PostgreSQL `pg-basebackup` restore prepareData so it cannot silently succeed with an empty, misplaced, or incomplete payload:

- `postgresql-basebackup` ActionSet now declares `VOLUME_DATA_DIR=/home/postgres/pgdata` and derives `DATA_DIR=$(VOLUME_DATA_DIR)/pgroot/data`.
- `pg-basebackup-restore.sh` now fails before pulling when `DATA_DIR` does not match the PostgreSQL runtime data subdirectory.
- After extraction, the script requires `PG_VERSION`, `base/`, `global/`, and `pg_wal/` under `pgroot/data`, and rejects PostgreSQL data files at the volume root.
- The old compatibility fallback now fails if neither `base.tar.gz` nor `base.tar` exists instead of printing `done!`.
- Added focused ShellSpec coverage for the success path and fail-fast cases.

## Evidence

Local checks on branch `easton/pg-restore-path-fix` commit `d586e60fa`:

- `shellspec --load-path ./shellspec --default-path addons/postgresql/scripts-ut-spec/pg_basebackup_restore_spec.sh --shell bash` -> 5 examples, 0 failures.
- `find addons/postgresql/dataprotection addons/postgresql/scripts addons/postgresql/scripts-ut-spec -type f \( -name '*.sh' -o -name '*.bash' \) -print0 | xargs -0 -n1 bash -n && git diff --check` -> pass.
- `helm dependency build addons/postgresql` then `helm template pg addons/postgresql` -> rendered `postgresql-basebackup` has `VOLUME_DATA_DIR=/home/postgres/pgdata` and `DATA_DIR=$(VOLUME_DATA_DIR)/pgroot/data`.

Full PostgreSQL shellspec suite was also attempted locally, but the existing `pitr_restore_spec` fails on this macOS host because the local environment lacks GNU `ls -I` compatibility under Bash 3.2. The new pg-basebackup spec is independent and passes.

## Runtime boundary

Draft until target-side restore validation runs on the fixed vcluster. This PR only proves the script/data-dir contract locally; it does not claim PostgreSQL restore PASS, full backup-restore acceptance, or release readiness.

### DIFF
--- a/addons/postgresql/dataprotection/pg-basebackup-restore.sh
+++ b/addons/postgresql/dataprotection/pg-basebackup-restore.sh
@@ -5,12 +5,82 @@ export PATH="$PATH:$DP_DATASAFED_BIN_PATH"
 export DATASAFED_BACKEND_BASE_PATH="$DP_BACKUP_BASE_PATH"
 
 function remote_file_exists() {
-    local out=$(datasafed list $1)
+    local out
+    out=$(datasafed list "$1")
     if [ "${out}" == "$1" ]; then
         echo "true"
         return
     fi
     echo "false"
+}
+
+function validate_data_dir_contract() {
+    if [[ -z "${DATA_DIR:-}" ]]; then
+        echo "ERROR: DATA_DIR is required for pg-basebackup restore" >&2
+        exit 1
+    fi
+
+    DATA_DIR="${DATA_DIR%/}"
+    if [[ -n "${VOLUME_DATA_DIR:-}" ]]; then
+        VOLUME_DATA_DIR="${VOLUME_DATA_DIR%/}"
+        local expected_data_dir="${VOLUME_DATA_DIR}/pgroot/data"
+        if [[ "${DATA_DIR}" != "${expected_data_dir}" ]]; then
+            echo "ERROR: pg-basebackup restore DATA_DIR must be ${expected_data_dir}, got ${DATA_DIR}" >&2
+            echo "ERROR: refusing to restore PostgreSQL files into the volume root or an unexpected subdirectory" >&2
+            exit 1
+        fi
+    fi
+}
+
+function log_restore_layout() {
+    echo "restore layout: VOLUME_DATA_DIR=${VOLUME_DATA_DIR:-<unset>} DATA_DIR=${DATA_DIR}"
+    if [[ -n "${VOLUME_DATA_DIR:-}" && -d "${VOLUME_DATA_DIR}" ]]; then
+        find "${VOLUME_DATA_DIR}" -maxdepth 3 -mindepth 1 -print 2>/dev/null \
+            | while IFS= read -r path; do printf '%s\n' "${path#${VOLUME_DATA_DIR}/}"; done \
+            | sort | head -200 || true
+    elif [[ -d "${DATA_DIR}" ]]; then
+        find "${DATA_DIR}" -maxdepth 2 -mindepth 1 -print 2>/dev/null \
+            | while IFS= read -r path; do printf '%s\n' "${path#${DATA_DIR}/}"; done \
+            | sort | head -200 || true
+    fi
+}
+
+function assert_no_pgdata_at_volume_root() {
+    if [[ -z "${VOLUME_DATA_DIR:-}" || "${VOLUME_DATA_DIR}" == "${DATA_DIR}" || ! -d "${VOLUME_DATA_DIR}" ]]; then
+        return
+    fi
+
+    local misplaced=()
+    for path in PG_VERSION base global pg_wal; do
+        if [[ -e "${VOLUME_DATA_DIR}/${path}" ]]; then
+            misplaced+=("${path}")
+        fi
+    done
+
+    if [[ ${#misplaced[@]} -gt 0 ]]; then
+        echo "ERROR: PostgreSQL data files were found at volume root ${VOLUME_DATA_DIR}: ${misplaced[*]}" >&2
+        echo "ERROR: restore payload must live under ${DATA_DIR}" >&2
+        log_restore_layout >&2
+        exit 1
+    fi
+}
+
+function assert_pgdata_restored() {
+    local missing=()
+    for path in PG_VERSION base global pg_wal; do
+        if [[ ! -e "${DATA_DIR}/${path}" ]]; then
+            missing+=("${path}")
+        fi
+    done
+
+    if [[ ${#missing[@]} -gt 0 ]]; then
+        echo "ERROR: invalid PostgreSQL data directory after pg-basebackup restore: ${DATA_DIR}" >&2
+        echo "ERROR: missing required entries: ${missing[*]}" >&2
+        log_restore_layout >&2
+        exit 1
+    fi
+
+    assert_no_pgdata_at_volume_root
 }
 
 function save_backup_end_lsn() {
@@ -54,33 +124,46 @@ function save_backup_end_lsn() {
     fi
 }
 
-mkdir -p ${DATA_DIR};
+function finish_restore() {
+    assert_pgdata_restored
+    save_backup_end_lsn
+    echo "done!";
+    exit 0
+}
+
+validate_data_dir_contract
+mkdir -p "${DATA_DIR}";
 
 if [ $(remote_file_exists "${DP_BACKUP_NAME}.tar.zst") == "true" ]; then
   datasafed pull -d zstd-fastest "${DP_BACKUP_NAME}.tar.zst" - | tar -xvf - -C "${DATA_DIR}/"
-  save_backup_end_lsn
-  echo "done!";
-  exit 0
+  finish_restore
 fi
 
 # for compatibility
 if [ $(remote_file_exists "${DP_BACKUP_NAME}.tar.gz") == "true" ]; then
   datasafed pull "${DP_BACKUP_NAME}.tar.gz" - | gunzip | tar -xvf - -C "${DATA_DIR}/"
-  save_backup_end_lsn
-  echo "done!";
-  exit 0
+  finish_restore
 fi
 
 # NOTE: restore from an old version backup, will be removed in 0.8
+restored_base_archive=false
 if [ $(remote_file_exists "base.tar.gz") == "true" ]; then
   datasafed pull "base.tar.gz" - | tar -xzvf - -C "${DATA_DIR}/"
+  restored_base_archive=true
 elif [ $(remote_file_exists "base.tar") == "true" ]; then
   datasafed pull "base.tar" - | tar -xvf - -C "${DATA_DIR}/"
+  restored_base_archive=true
 fi
 if [ $(remote_file_exists "pg_wal.tar.gz") == "true" ]; then
+  mkdir -p "${DATA_DIR}/pg_wal"
   datasafed pull "pg_wal.tar.gz" - | tar -xzvf - -C "${DATA_DIR}/pg_wal/"
 elif [ $(remote_file_exists "pg_wal.tar") == "true" ]; then
+  mkdir -p "${DATA_DIR}/pg_wal"
   datasafed pull "pg_wal.tar" - | tar -xvf - -C "${DATA_DIR}/pg_wal/"
 fi
-save_backup_end_lsn
-echo "done!";
+if [[ "${restored_base_archive}" != "true" ]]; then
+  echo "ERROR: no supported pg-basebackup artifact found for ${DP_BACKUP_NAME}" >&2
+  echo "ERROR: expected ${DP_BACKUP_NAME}.tar.zst, ${DP_BACKUP_NAME}.tar.gz, base.tar.gz, or base.tar" >&2
+  exit 1
+fi
+finish_restore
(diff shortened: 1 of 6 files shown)


weicao:
## Review: REQUEST CHANGES — split the PR; the guards are right, the bootstrap-hook half has an incomplete lifecycle

**The good half (land as-is)**: `validate_data_dir_contract` + fail-before-`datasafed pull` + post-extract payload assertion (`PG_VERSION/base/global/pg_wal` under pgroot/data) is the correct minimal root-cause fix for #3111 — no more `done!` with zero artifacts, no more silent restore-to-wrong-directory. Data-dir derivation matches in-repo precedent exactly (PITR ActionSet's `VOLUME_DATA_DIR`+`$(VAR)` pattern, `cmpd.yaml:372-386`).

**The problem half — restore-hook/Patroni bootstrap injection (HIGH)**: the PR copies the wal-g signal/hook pattern but omits its essential step. `wal-g-restore.sh:156-158` documents the contract: data is moved to `${DATA_DIR}.old` **because Patroni expects an empty data directory and `kb_restore.sh` moves it back during bootstrap**. This PR leaves restored data *in* DATA_DIR, so on the primary success path Patroni **adopts** the existing dir, `bootstrap.method: kb_restore` never executes, and `kb_restore.signal` persists forever. Consequences:
1. `generate_patroni_yaml.py` re-injects bootstrap/replica config on every restart of that pod;
2. a later empty-data-dir event on the pod (patronictl reinit, PVC replacement) fires the **stale** kb_restore bootstrap, whose completeness check exits 1 → bootstrap retry loop instead of normal recovery.

The shellspec invokes the hook manually so this is invisible to the PR's tests.

**MEDIUM — cross-flow blast radius**: `kb_restore.signal` is also created by `wal-g-restore.sh:66` and `wal-g-archive-restore.sh:34`; the unconditional `bootstrap.method='kb_restore'` in the shared `generate_patroni_yaml.py` now applies to those flows too — wal-g full restore survives only because the PITR block later overwrites `method`, an order-dependent untested coincidence.

**Recommendation**: split. PR A = ActionSet env + restore-script guards (approve immediately). PR B = the bootstrap machinery, which must either adopt the full wal-g `.old` contract (hook actually runs and clears the signal) or drop bootstrap injection entirely — with runtime validation on BOTH basebackup and wal-g flows before ready.

This matters operationally right now: this PR is queued as restore-chain layer 3 validation after the clean-epoch decision — in its current shape the validation could pass superficially (adoption path works) while planting the stale-signal trap for every restored cluster's future restarts.

weicao:
## Re-review of 2e342395 (delta from a22c2a1f) — HIGH blocker resolved

Static verdict on the restore/bootstrap half: **the adoption-path finding is fixed**. Four checkpoints:

1. **Adoption path closed.** `finish_restore` now stages the payload in `DATA_DIR.old` and recreates an empty `DATA_DIR`, so Patroni cannot adopt a pre-populated PGDATA — it must run bootstrap, and the `kb_restore` hook always executes (`pg-basebackup-restore.sh` finish_restore + hook).

2. **Hook is fail-closed.** Validates `.old` exists and is complete (PG_VERSION/base/global/pg_wal), requires `DATA_DIR` empty (`find -mindepth 1 -maxdepth 1` also catches dotfiles), and clears `kb_restore.signal` only *after* the commit `mv`. The stale-signal landmine from my previous review now surfaces as an explicit ERROR + failed bootstrap instead of a silent stale re-bootstrap — diagnosable, not latent.

3. **prepareData is retry-safe.** A Job re-run after a completed pass re-pulls into the (empty, recreated) `DATA_DIR` and re-stages; `rm -rf ${DATA_DIR}.old` before the `mv` prevents the nested-`.old/data` corruption class (which, note, still exists on main in `wal-g-restore.sh` — separate issue incoming). The crash windows I traced (between hook-write / mv / mkdir) all converge on re-run.

4. **Cross-flow (my earlier MEDIUM) resolved by analysis.** The new explicit `bootstrap.method=kb_restore` injection in `generate_patroni_yaml.py` is overridden by the PITR block (`local_config['bootstrap'].update(pitr_config)`) whenever `pgdata/conf/recovery.conf` exists — and every wal-g flavor writes that file (`wal-g-restore.sh:150`, `wal-g-archive-restore.sh:104`), so wal-g chains keep the `kb_restore_from_time` wiring from main unchanged; the new method only takes effect for the basebackup flavor, as intended. The dropped `if 'postgresql' not in local_config` guard was hoisted to L49-51 — safe.

Bonus: the `restored_base_archive` guard also fixes the legacy empty-artifact fall-through (a restore that finds no artifact no longer "succeeds" into an empty initdb) — this independently closes a finding from my fresh main-head review.

**Scope + gate:** this verdict covers the restore/bootstrap file family (pg-basebackup-restore.sh, generate_patroni_yaml.py, configmap.yaml, actionsets + their specs). The switchover/reloader/auth changes bundled here stand under the earlier review pass. With the fix landed coherently in place I withdraw the hard split request. PR should stay draft until a fresh deploy/runtime run on 2e342395 — please include one wal-g restore in that round to empirically confirm checkpoint 4.

weicao:
## Re-review of 67b8f4a20 (marker protocol delta from 2e342395) — static CLEAR

Crash-window walk of the new handoff protocol, all states converge:

- **Crash after `mv`, before signal-clear** → next invocation takes the marker path: `verify_pgdata_complete` on `DATA_DIR`, `sync`, clear signal then marker, exit 0. Self-healing, closes the class-3 commit-boundary blocker.
- **Marker placement is correct**: touched into `.old` *before* `chown -R` (ownership covered) and *before* the `mv` (travels with the payload); `.old` and `DATA_DIR` share a parent, so the rename is atomic on the same filesystem — "marker present but payload torn" is not reachable, and `verify_pgdata_complete` guards it anyway.
- **Post-move `sync`** provides the durability barrier Derek required.
- **Cleanup order (signal before marker) is the safe direction**: the reverse order would strand a signal with no marker and hard-fail the retry.
- **Non-marker non-empty `DATA_DIR` hard fail** preserves the fail-closed stance for genuinely foreign data.

Independent verification on my machine: `pg_basebackup_restore_spec.sh` **11 examples, 0 failures**, and the case list covers all five required behaviors (hook-fail-before-commit signal retention ×3 variants, replica cleanup, moved-before-signal-clear convergence, missing-`.old`+non-marker hard fail). `generate_patroni_yaml_spec.sh` failed 2 examples locally — traced to my host missing PyYAML, not to the PR (Easton/Derek both ran it green in complete envs).

Two non-blocking nits, no action required before runtime:

1. If the hook crashes between signal-clear and marker-clear, the `.kb_restore_handoff` dotfile stays in PGDATA forever (the hook is never invoked again once the signal is gone). Benign — postgres ignores unknown files and no future flow misreads it — but worth a `rm -f` in postReady cleanup someday.
2. `generate_patroni_yaml_spec.sh` fails rather than skips when `python3 -c 'import yaml'` is unavailable; a skip-guard would make the suite portable.

**Verdict: static gate CLEAR from my side at 67b8f4a20.** Runtime round may launch on this head; please keep one wal-g restore in the round to empirically confirm the cross-flow analysis from my previous comment.

weicao:
## Integrated-tree deep review at 67b8f4a20 — new critical: the staging protocol breaks the pg-basebackup → archive-wal PITR chain

The standalone static clear (marker protocol, retry safety, cross-flow via generate_patroni_yaml) stands. This finding is at a level none of the three standalone reviews looked at: **composition with the Continuous (archive-wal) prepareData that runs after this script in a PITR restore**.

**Mechanism:**

- `pg-basebackup-restore.sh` now stages unconditionally: `finish_restore` always does `mv ${DATA_DIR} ${DATA_DIR}.old; mkdir -p ${DATA_DIR}` and writes the hook + `kb_restore.signal`. There is **no `DP_RESTORE_TIMESTAMP`/`DP_RESTORE_TIME` guard anywhere in the script** (grep: zero hits).
- The wal-g family guards exactly this composition: `wal-g-restore.sh:55-57` exits early (no staging, no hook) when `DP_RESTORE_TIMESTAMP` is set, leaving the Continuous prepareData to finish the job.
- `postgresql-pitr-restore.sh:7-16` (the archive-wal Continuous prepareData) assumes the Full restore left data **in `DATA_DIR`**: its retry branch requires `! -d ${DATA_DIR}` (false now — `DATA_DIR` exists, empty), then `ls ${DATA_DIR}/pg_wal` on the empty dir → empty `latest_wal` → `basename` with no operand exits 1 under the actionset's `set -e` (`actionset-postgresql-pitr.yaml:39`). **Every retry fails identically: point-in-time restore from a pg-basebackup base is permanently broken.**

**Companion (major): the leaked signal.** `write_restore_hook` creates `kb_restore.signal`; `postgresql-pitr-restore.sh:26-28` then overwrites the hook with the legacy 3-line version that (a) ignores `--replica` and exits 0 with no data staged — so Patroni's `restore_data` create-replica method "succeeds" producing nothing instead of falling through to basebackup — and (b) never removes the signal, so the restore-bootstrap branch re-arms on every container start. On main this chain never created a signal, so the branch never fired.

**Fix:** mirror the wal-g pattern — when `DP_RESTORE_TIMESTAMP`/`DP_RESTORE_TIME` is set, skip staging + hook/signal writing and leave the extracted data in `DATA_DIR` for the Continuous stage. Note this hands the chain back to the (pre-existing) main-head defect #3119 in `postgresql-pitr-restore.sh` — that issue's fix remains required independently; ideally land them coordinated. Please also add a combined-state spec case (`.old` populated + `DATA_DIR` empty at Continuous entry) — #3072's `pitr_restore_spec.sh:118-124` currently encodes the pre-#3112 state.

**Minor (hygiene):** `.backup_log` / `.backup_end_lsn` are written into `DATA_DIR` (lines 92-121), ride the staging into the live PGDATA, and nothing consumes `.backup_end_lsn`; now that the marker protocol makes stray-file hygiene in `DATA_DIR` load-bearing, route these to stderr / `RESTORE_SCRIPT_DIR`.

weicao:
Fix pushed (06de2d1d) per the review: `finish_restore` now mirrors wal-g — when `DP_RESTORE_TIMESTAMP`/`DP_RESTORE_TIME` is set it verifies and exits with the data left in `DATA_DIR` (no staging, no hook, no signal), so the archive-wal Continuous prepareData finds `pg_wal` where it expects and no signal leaks to the legacy hook. Added the combined-state spec case (timestamp set ⇒ unstaged, hook/signal absent). Local spec 12/12; integrated-tree suite 93/93.

weicao:
## Delta review 67b8f4a20 → 5003f0515 — pin refreshed, static clear stands

The three commits since my last clear:

1. **Permission normalization** (`chmod 00700 "${DATA_DIR}.old"` in the hook): postgres refuses a PGDATA that is group/other-accessible or carries setgid, and a basebackup tar can preserve such modes. The leading `00` also clears special bits. Correctly placed after `chown`, **before** the `sync` durability barrier — the commit-marker ordering contract is preserved. Covered by the new 0775/2775 spec cases.
2. **`recovery_conf: {}` removed from the injected `kb_restore` bootstrap method**: an empty recovery_conf made Patroni treat the plain basebackup bootstrap as a recovery bootstrap; a full restore has nothing to recover — the hook already strips `standby.signal`/`recovery.signal`. The wal-g/PITR chains are unaffected (their method + recovery_conf still comes from `kb_pitr.conf` via the PITR block, which overrides this injection). Spec added.

Both changes were also swept in my 19-PR integration verification (suite green apart from two known cross-PR spec follow-ons unrelated to this PR). Static verdict for head `5003f0515`: **clear** — R1l's runtime evidence can count against this exact pin.

weicao:

PR #3115, opened 2026-07-10
fix(postgresql): accept wal_init_zero as boolean

weicao:
## Problem

apecloud/apecloud#19428 reports that changing PostgreSQL `configuration.wal_init_zero` fails.

Dev evidence from the linked KBE issue:

- A fresh `postgresql-17` cluster accepted a direct KB OpsRequest with `wal_init_zero=on`; the ConfigMap changed to `wal_init_zero = on`, and runtime `SHOW wal_init_zero` returned `on`.
- The same cluster rejected a direct KB OpsRequest with `wal_init_zero=true` because the addon CUE constraint only allowed string `"on" | "off"`.

PostgreSQL official docs define `wal_init_zero` as a boolean parameter, and PostgreSQL boolean settings may be written as `on/off`, `true/false`, `yes/no`, or `1/0`.

## Fix

- Change `wal_init_zero` CUE constraints for PostgreSQL pg12/14/15/16/17/18 from string enum `"on" | "off"` to the existing addon boolean pattern: `bool & false | true`.
- Keep the default template value as `wal_init_zero = off`, which is a valid PostgreSQL boolean token and preserves existing rendered config behavior.
- Add a shellspec guard to keep all supported PostgreSQL majors aligned.

## Validation

- `shellspec --load-path ./shellspec --default-path addons/postgresql/scripts-ut-spec/wal_init_zero_spec.sh --shell bash`
- `shellspec --load-path ./shellspec --default-path addons/postgresql/scripts-ut-spec --shell bash`
- `helm template pg addons/postgresql` and checked rendered ParametersDefinitions contain `wal_init_zero?: bool & false | true` while templates still render `wal_init_zero = off`.
- `git diff --check`

Related: apecloud/apecloud#19428
Replaces metadata-only attempt: apecloud/apecloud#19492

### DIFF
--- a/addons/postgresql/config/pg12-config-constraint.cue
+++ b/addons/postgresql/config/pg12-config-constraint.cue
@@ -1037,7 +1037,7 @@
 	wal_writer_flush_after?: int & >=0 & <=2147483647 @storeResource(8KB)
 
 	// If set to on (the default), this option causes new WAL files to be filled with zeroes. On some file systems, this ensures that space is allocated before we need to write WAL records. However, Copy-On-Write (COW) file systems may not benefit from this technique, so the option is given to skip the unnecessary work. If set to off, only the final byte is written when the file is created so that it has the expected size.
-	wal_init_zero?: string & "on" | "off"
+	wal_init_zero?: (bool & (false | true)) | (string & ("on" | "off"))
 
 	// (kB) Sets the maximum memory to be used for query workspaces.
 	work_mem?: int & >=64 & <=2147483647 @storeResource(1KB)
--- a/addons/postgresql/config/pg12-config-constraint.cue
+++ b/addons/postgresql/config/pg12-config-constraint.cue
@@ -1037,7 +1037,7 @@
 	wal_writer_flush_after?: int & >=0 & <=2147483647 @storeResource(8KB)
 
 	// If set to on (the default), this option causes new WAL files to be filled with zeroes. On some file systems, this ensures that space is allocated before we need to write WAL records. However, Copy-On-Write (COW) file systems may not benefit from this technique, so the option is given to skip the unnecessary work. If set to off, only the final byte is written when the file is created so that it has the expected size.
-	wal_init_zero?: string & ("on" | "off" | "true" | "false")
+	wal_init_zero?: string & "on" | "off" | "true" | "false"
 
 	// (kB) Sets the maximum memory to be used for query workspaces.
 	work_mem?: int & >=64 & <=2147483647 @storeResource(1KB)
--- a/addons/postgresql/config/pg12-config-constraint.cue
+++ b/addons/postgresql/config/pg12-config-constraint.cue
@@ -1037,7 +1037,7 @@
 	wal_writer_flush_after?: int & >=0 & <=2147483647 @storeResource(8KB)
 
 	// If set to on (the default), this option causes new WAL files to be filled with zeroes. On some file systems, this ensures that space is allocated before we need to write WAL records. However, Copy-On-Write (COW) file systems may not benefit from this technique, so the option is given to skip the unnecessary work. If set to off, only the final byte is written when the file is created so that it has the expected size.
-	wal_init_zero?: (bool & (false | true)) | (string & ("on" | "off"))
+	wal_init_zero?: string & ("on" | "off" | "true" | "false")
 
 	// (kB) Sets the maximum memory to be used for query workspaces.
 	work_mem?: int & >=64 & <=2147483647 @storeResource(1KB)
--- a/addons/postgresql/config/pg14-config-constraint.cue
+++ b/addons/postgresql/config/pg14-config-constraint.cue
@@ -1106,7 +1106,7 @@
 	work_mem?: int & >=64 & <=2147483647 @storeResource(1KB)
 
 	// If set to on (the default), this option causes new WAL files to be filled with zeroes. On some file systems, this ensures that space is allocated before we need to write WAL records. However, Copy-On-Write (COW) file systems may not benefit from this technique, so the option is given to skip the unnecessary work. If set to off, only the final byte is written when the file is created so that it has the expected size.
-	wal_init_zero?: string & "on" | "off"
+	wal_init_zero?: (bool & (false | true)) | (string & ("on" | "off"))
 
 	// Sets how binary values are to be encoded in XML.
 	xmlbinary?: string & "base64" | "hex"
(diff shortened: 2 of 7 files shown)


leon-ape (commented):
Finding: Target KubeBlocks validates boolean parameter assignments with `strconv.ParseBool`, which accepts `true`/`false` and `1`/`0` but not `on`/`off`. Changing `wal_init_zero` from the `"on" | "off"` string enum to `bool` therefore makes the previously valid `wal_init_zero=on/off` OpsRequests fail validation.

weicao:
PostgreSQL owner check: I agree with Leon's review; this PR should not merge as-is.

Reason: changing `wal_init_zero` from the existing string enum `"on" | "off"` to CUE `bool` changes the user-facing OpsRequest contract. KubeBlocks' bool validation path uses Go `strconv.ParseBool`, whose accepted values are `true/false`, `1/0`, and `t/f` variants, not PostgreSQL-style `on/off` (see https://pkg.go.dev/strconv#ParseBool). So existing valid reconfigure inputs like `wal_init_zero=on` or `wal_init_zero=off` would start failing validation.

Recommended direction: keep the addon constraint compatible with the existing PostgreSQL-style values (`on/off`), or add an explicit compatibility layer that accepts/migrates `on/off` before switching the public schema to bool. The current ShellSpec only checks that the files were changed to bool; it does not cover the compatibility regression Leon identified.

weicao:
Updated head to `3faa114fa6047019320418bed3027afa9295a5d6` to address Leon's compatibility finding.

Change in this head:
- `wal_init_zero` is no longer a pure CUE `bool`.
- It now accepts both forms:
  - bool `true` / `false`, for the Cloud metadata path that currently sends boolean values;
  - string `on` / `off`, preserving the existing PostgreSQL-style OpsRequest/config contract.
- Six PostgreSQL versions remain covered: pg12/14/15/16/17/18.
- Template defaults remain `wal_init_zero = off`.

Validation on new head:
- `shellspec --load-path ./shellspec --default-path addons/postgresql/scripts-ut-spec/wal_init_zero_spec.sh --shell bash` => 6 examples, 0 failures
- `shellspec --load-path ./shellspec --default-path addons/postgresql/scripts-ut-spec --shell bash` => 26 examples, 0 failures
- `helm template pg addons/postgresql >/tmp/postgresql-pr3115-rendered.yaml` => rendered `constraints=6 defaults=6`
- `git diff --check`

So the intended contract is compatibility, not a breaking public-schema switch to pure bool.

weicao:
## Review: REQUEST CHANGES — the union type does not fix the on/off regression (empirically verified)

**Decisive finding**: the current head's `(bool & (false|true)) | (string & ("on"|"off"))` disjunction is classified **BoolType** by KubeBlocks' CUE visitor — `pkg/parameters/validate/cue_visitor.go:57-66` classifies by bitmask on `IncompleteKind()` and the BoolKind case precedes StringKind, so the string branch is unreachable at the transform layer. Verified by running the production `ValidateConfigWithCue` path against this PR's exact schema:
- `wal_init_zero=on` → `REJECTED: strconv.ParseBool: parsing "on": invalid syntax`
- `off` → REJECTED; `true/false/1/0` → accepted

**Second critical**: the templates retain `wal_init_zero = off` (e.g. `pg14-config.tpl:280`) and validation transforms the whole rendered file — under this schema the probe rejected a file containing `wal_init_zero=off`, meaning **any reconfigure touching postgresql.conf on these versions would fail validation** after merge.

**Recommended fix** (consistent with the addon's 12 existing on/off string-enum params — `archive_mode`, `huge_pages`, `constraint_exclusion`, etc. — and PostgreSQL-native practice):
```cue
wal_init_zero?: string & ("on" | "off" | "true" | "false")
```
StringType → no ParseBool; on/off OpsRequests stay valid; the Cloud path's true/false accepted; tpl default stays valid. Do not go the pure-bool direction: no mixed bool|string disjunction exists anywhere in this repo, and bool breaks existing on/off users by definition.

Also: `scripts-ut-spec/wal_init_zero_spec.sh` is grep-on-file-text only — it cannot catch this pipeline-level regression and gave false confidence; suggest replacing with a case table over the widened enum.

weicao:
## Deep review at head 3faa114fa (integrated-tree pass) — REQUEST CHANGES, now merge-blocking

The rework did not implement the agreed string-enum approach, and the new head is a **regression vs main**.

**1. `on`/`off` are still rejected — the union is still classified BoolType.**
All six cue files now carry:

```cue
wal_init_zero?: (bool & (false | true)) | (string & ("on" | "off"))
```

`IncompleteKind()` for this disjunction is `BoolKind|StringKind`, and `pkg/parameters/validate/cue_visitor.go:57-66` checks BoolKind **before** StringKind → field classified `BoolType` → `strconv.ParseBool` transform (cue_visitor.go:154) → `on`/`off` fail, exactly as at the previous head. `value_transformer.go` doesn't rescue Properties-format files (`needValueTransformer`, value_transformer.go:63-67). Net effect: `true/false` now pass, but `on/off` — PostgreSQL's canonical spelling, accepted on main — now fail.

**2. The pre-existing template default now breaks every major.**
All six `pg*-config.tpl` ship `wal_init_zero = off` (e.g. pg14:280). On main the CUE type was a pure string enum, so this validated. With the new BoolType classification, whole-rendered-file validation (`ValidateConfigWithCue`) fails on `ParseBool("off")` — **initial config render and every postgresql.conf reconfigure fail on all six majors**. This is why the integrated review flags it as merge-blocking, not just parameter-spelling.

**3. The new spec asserts the broken text, not behavior.**
`scripts-ut-spec/wal_init_zero_spec.sh:5-9` greps for the exact cue string and the `= off` default — it passes precisely because both halves of the broken pair are present. A behavioral test (run the value through the validator, or at minimum assert the type is a pure string disjunction) would have caught this.

**Fix (same as the original review recommendation):**

```cue
wal_init_zero?: string & ("on" | "off" | "true" | "false")
```

StringKind-only → StringType → no ParseBool; all four spellings valid to PostgreSQL; the `= off` tpl default stays valid. Please also convert the spec to assert behavior, and consider extracting this as a shared `#PgBool` so the ~80 other bool-typed params can adopt the same pattern later.

weicao:
Fix pushed (ad9587c8) per the review: pure string enum `string & ("on" | "off" | "true" | "false")` in all six cue files (StringKind-only → no ParseBool; the `= off` tpl default stays valid), and the spec now asserts the type contract + guards against any bool-branch regression instead of grepping the old text. Local spec 6/6; integrated-tree suite 93/93.

PR #3126, opened 2026-07-10
fix(etcd): honor inject image pull policy

### LINKED ISSUE #3125: etcd inject-bash ignores images.pullPolicy override
## Problem

The etcd addon declares `images.pullPolicy`, but the inject-bash init container template passes Helm's `default` arguments in reverse order:

```gotemplate
{{ default .Values.images.pullPolicy "IfNotPresent" }}
```

Because the given value is the non-empty literal `IfNotPresent`, the rendered policy is always `IfNotPresent`; setting `images.pullPolicy=Never` or `Always` is silently ignored.

This blocks deterministic local-sideload validation in IDC/vcluster environments, where the image policy must match the supplied local image contract.

## Constructed test scene

- `kubeblocks-addons/main@554fd848cecbf7943da6d8aa206c673bcbbfcc74`
- addon `addons/etcd`, chart `1.2.0-alpha.2`
- render once with defaults and once with both etcd image policies set to `Never`

## Reproducibility

Deterministic on every Helm render of the pinned source.

## Reproduction steps

```bash
helm template etcd addons/etcd --dependency-update \
  --set images.pullPolicy=Never \
  --set image.pullPolicy=Never > /tmp/etcd.yaml

grep -n 'imagePullPolicy:' /tmp/etcd.yaml
```

Observed before the fix in the ComponentDefinition:

```yaml
initContainers:
  - name: inject-bash
    imagePullPolicy: IfNotPresent
containers:
  - name: etcd
    imagePullPolicy: Never
```

## Expected

Helm's `default` call should use `IfNotPresent` as the fallback and `.Values.images.pullPolicy` as the given value. Defaults should still render both policies as `IfNotPresent`; an explicit `Never` override should render both as `Never`.

weicao:
## Summary

Fix the Helm `default` argument order so the etcd `inject-bash` init container honors `images.pullPolicy` while retaining `IfNotPresent` as the default.

Fixes #3125.

## Validation

```text
default policies:
  inject-bash: IfNotPresent
  etcd:        IfNotPresent

images.pullPolicy=Never,image.pullPolicy=Never:
  inject-bash: Never
  etcd:        Never
```

- `helm template addons/etcd --dependency-update` for default and explicit Never values
- `helm lint addons/etcd`: 1 chart linted, 0 failed

## Scope

One template expression only. No image reference, runtime command, lifecycle action, or backup/restore behavior changes.

### DIFF
--- a/addons/etcd/templates/cmpd.yaml
+++ b/addons/etcd/templates/cmpd.yaml
@@ -13,7 +13,7 @@ spec:
   runtime:
     initContainers:
       - name: inject-bash
-        imagePullPolicy: {{default .Values.images.pullPolicy "IfNotPresent"}}
+        imagePullPolicy: {{ default "IfNotPresent" .Values.images.pullPolicy }}
         volumeMounts:
           - name: scripts
             mountPath: /scripts


weicao:
Runtime/deploy evidence update for current head `5ac558cf11a62cb1542c0646c001efca1ba29ff3`:

- Exact chart artifact SHA256: `90be8842b04e653febe06412dd5764b4f21761db158146d7cbf95d0fbb5d461b`.
- Explicit `images.pullPolicy=Never` render produced `Never` for both the etcd container and `inject-bash`; the chart was installed as Helm revision 1.
- Live definitions present: ComponentDefinition `etcd-3-1.2.0-alpha.2`, ComponentVersion `etcd`, ActionSet `etcdctl-br`, and BackupPolicyTemplate `etcd-backup-policy-template`.
- No Addon CR or etcd business Cluster/Pod was created during this deploy-only gate.
- CI is now complete/green for chart checks, generation, shellcheck, and shellspec.

Evidence boundary: this closes the definitions-deploy and explicit-pull-policy path only. Backup/restore runtime has not started because the shared idc4 host is at the max-Pod environment gate (current nodes `112/110`, `110/110`, `109/110`); that is Layer 1 environment capacity, not an addon runtime verdict. Runtime evidence will follow only after stable capacity is available.

leon-ape (approved):
LGTM

PR #3132, opened 2026-07-11
fix(redis): fail fast on PITR archive errors

weicao:
## Summary

Fixes #3131.

This patch closes the Redis PITR fail-open paths reported in the static review:

- enable `set -e -o pipefail` in `pitr-backup.sh`, so failed `datasafed push` / archive pipelines stop before remote cleanup or local AOF/base-file deletion;
- enable `set -e -o pipefail` in `pitr-restore.sh`, so failed archive pull/extract exits non-zero;
- make restore return non-zero when no archive matches the requested restore time;
- make restore return non-zero for unknown PITR object types;
- fix the backup purge timestamp typo `currentUnix` -> `current_unix`;
- add focused Redis ShellSpec coverage for these failure contracts with stubbed `datasafed`/`redis-cli`/`redis-check-rdb`.

## Verification

Local checks:

```bash
bash -n addons/redis/dataprotection/pitr-backup.sh addons/redis/dataprotection/pitr-restore.sh addons/redis/scripts-ut-spec/pitr_dataprotection_spec.sh
shellspec --load-path ./shellspec --shell /opt/homebrew/bin/bash addons/redis/scripts-ut-spec/pitr_dataprotection_spec.sh
shellspec --load-path ./shellspec --shell bash addons/redis/scripts-ut-spec/pitr_dataprotection_spec.sh
git diff --check
```

Results:

- Bash 5 focused ShellSpec: 4 examples, 0 failures
- macOS Bash 3 default path: guarded skip, 0 examples, 0 failures
- `bash -n`: PASS
- `git diff --check`: PASS

Not run locally:

- `shellcheck`: unavailable on this machine (`shellcheck: command not found`)
- live Redis PITR runtime: not run yet; this PR remains draft pending CI and runtime acceptance

## Boundary

This PR only covers the PITR backup/restore fail-fast blocker. It does not address the separate sentinel fallback/split-brain blocker from the same review; that should be handled in a separate P0 PR.

PR #3134, opened 2026-07-11
fix(redis): fail closed on sentinel restart without primary

### LINKED ISSUE #3133: Redis sentinel startup should not default-primary fallback on restart data
## Problem
Redis sentinel-mode startup currently falls back to the default/min-lexicographical primary when every configured Sentinel fails to return primary information. That fallback is only safe before the first bootstrap. If Redis has already persisted data, choosing a default primary during a Sentinel outage can start a stale pod as primary and create a split-brain/data-loss window.

Affected scripts:

- `addons/redis/scripts/redis-start.sh`
- `addons/redis/scripts/redis5-start.sh`

## Constructed test scenario
Add focused Redis ShellSpec coverage for both Redis 6/7 and Redis 5 startup scripts:

1. Sentinel is configured.
2. All sentinel queries fail or return empty primary info.
3. The Redis data directory contains persisted data such as `dump.rdb`.
4. Startup must exit non-zero and must not call `get_default_initialize_primary_node`.

A separate first-bootstrap case keeps compatibility:

1. Sentinel is configured.
2. All sentinel queries fail.
3. The Redis data directory has no persisted Redis data, only ignored startup/account marker files.
4. Startup may use the default primary for first bootstrap.

## Reproducibility
Reproducible with local ShellSpec stubs; no live Sentinel cluster is required for the failure contract.

## How to reproduce
Run the focused Redis start specs after applying the regression tests:

```bash
shellspec --load-path ./shellspec --shell /opt/homebrew/bin/bash addons/redis/scripts-ut-spec/redis_start_spec.sh addons/redis/scripts-ut-spec/redis5_start_spec.sh
```

Expected behavior after the fix:

- restart/data-present case exits non-zero and does not fallback to default primary;
- first-bootstrap/data-empty case still permits default-primary fallback;
- both Redis 6/7 and Redis 5 startup scripts follow the same contract.

weicao:
## Summary

Fixes #3133.

This patch closes the Redis sentinel fallback blocker while preserving first-bootstrap behavior:

- when Sentinel is configured but no Sentinel returns primary info, startup now checks whether Redis already has persisted data;
- if persisted Redis data exists, startup exits non-zero instead of falling back to the default/min-lexicographical primary;
- if the Redis data directory has no persisted Redis data, startup may still use the default primary for first bootstrap;
- applies the same contract to Redis 6/7 `redis-start.sh` and Redis 5 `redis5-start.sh`;
- adds focused ShellSpec coverage for restart/data-present fail-closed and first-bootstrap/data-empty fallback cases.

## Verification

Local checks:

```bash
bash -n addons/redis/scripts/redis-start.sh addons/redis/scripts/redis5-start.sh addons/redis/scripts-ut-spec/redis_start_spec.sh addons/redis/scripts-ut-spec/redis5_start_spec.sh
shellspec --load-path ./shellspec --shell /opt/homebrew/bin/bash addons/redis/scripts-ut-spec/redis_start_spec.sh addons/redis/scripts-ut-spec/redis5_start_spec.sh
shellspec --load-path ./shellspec --shell bash addons/redis/scripts-ut-spec/redis_start_spec.sh addons/redis/scripts-ut-spec/redis5_start_spec.sh
git diff --check
```

Results:

- Bash 5 focused ShellSpec: 72 examples, 0 failures
- macOS Bash 3 default path: 72 examples, 0 failures, 11 guarded skips
- `bash -n`: PASS
- `git diff --check`: PASS

Not run locally:

- `shellcheck`: unavailable on this machine (`shellcheck: command not found`)
- live sentinel restart runtime: not run yet; this PR remains draft pending CI and runtime acceptance

## Boundary

This PR only covers the sentinel fallback/split-brain blocker. The PITR backup/restore fail-fast blocker is handled separately in #3132.

Testing handoff boundary per #all direction: Redis dev side specifies the contract and focused test scenario here; Redis test owner should execute runtime validation and collect pod/log/evidence artifacts against this exact head.

### DIFF
--- a/addons/redis/scripts-ut-spec/redis5_start_spec.sh
+++ b/addons/redis/scripts-ut-spec/redis5_start_spec.sh
@@ -531,5 +531,83 @@ Describe "Redis5 Start Bash Script Tests"
         The stdout should include "Error: Required environment variable SENTINEL_POD_FQDN_LIST is not set."
       End
     End
+
+    Context "when sentinel is configured but no sentinel returns primary info"
+      setup() {
+        export REDIS_DATA_DIR="./redis5-data"
+        export SENTINEL_COMPONENT_NAME="redis-sentinel"
+        export SENTINEL_POD_FQDN_LIST="sentinel-0.redis-sentinel-headless,sentinel-1.redis-sentinel-headless"
+        mkdir -p "$REDIS_DATA_DIR"
+        echo "persisted" > "$REDIS_DATA_DIR/dump.rdb"
+      }
+      Before "setup"
+
+      un_setup() {
+        unset SENTINEL_COMPONENT_NAME
+        unset SENTINEL_POD_FQDN_LIST
+        unset REDIS_DATA_DIR
+        rm -rf ./redis5-data
+      }
+      After "un_setup"
+
+      It "fails closed instead of falling back to default primary when Redis data exists"
+        Skip if "shell type and version unmatch, please check!" should_skip_when_shell_type_and_version_invalid
+        getent() {
+          echo "10.0.0.1 $2"
+        }
+        retry_get_master_addr_by_name_from_sentinel() {
+          return 1
+        }
+        get_default_initialize_primary_node() {
+          echo "SHOULD_NOT_FALLBACK"
+        }
+        When run init_or_get_primary_from_redis_sentinel
+        The status should be failure
+        The stdout should include "Failed to retrieve primary info from sentinel: sentinel-0.redis-sentinel-headless"
+        The stdout should include "Failed to retrieve primary info from sentinel: sentinel-1.redis-sentinel-headless"
+        The stdout should include "Error: no primary node found from all redis sentinels and Redis data already exists; refusing to use default primary while sentinel is configured."
+        The stdout should not include "SHOULD_NOT_FALLBACK"
+      End
+    End
+
+    Context "when no sentinel returns primary info before first bootstrap"
+      setup() {
+        service_port="6379"
+        export REDIS_POD_NAME_LIST="redis-1,redis-0"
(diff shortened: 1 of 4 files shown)


weicao (commented):
Reviewer pass (Helios, original B1 finding author) on head `ad7e927` — verified against PR branch source.

**B1 (silent fallback to min-lex-pod-as-primary on sentinel outage) — FIXED, verified, and with the right nuance.** `redis-start.sh:196-202`: when no sentinel returns primary info, it now branches on `redis_data_dir_has_existing_data` — data present ⇒ `exit 1` (no fallback, container restarts, no stale second-primary); data-empty first bootstrap ⇒ default-primary fallback still allowed. This is exactly the bootstrap-vs-restart distinction, not a blanket fail-closed that would break first provisioning.

**The data-dir helper is the crux and it's correct** (`:217-229`): it excludes precisely the addon's own non-user-data files (`users.acl`, `users.acl.bak`, `.fixed_pod_ip_enabled`, `.kb_redis_start_initialized`) via `find -mindepth 1 -maxdepth 1 ! -name ... -print -quit`, so an RDB/AOF presence triggers fail-closed while a genuinely-fresh dir still bootstraps. `-print -quit` also keeps it cheap. Applied consistently to `redis5-start.sh`.

Scope clean (sentinel startup only; PITR stays in #3132). 72/0 focused ShellSpec on Bash 5 is solid coverage for the branch logic. Residual is the stated gap — no live sentinel-restart runtime yet — so keeping draft pending a runtime that: (a) persists data, kills all sentinels, restarts a pod, asserts non-zero exit + no `replicaof` written; (b) confirms data-empty first bootstrap still converges. That runtime is the only thing between this and merge from a contract standpoint. No blockers from my side.

PR #3108, opened 2026-07-08
fix(mariadb): add bounded Galera ordered shutdown

weicao:
## Summary

- keep Galera `podManagementPolicy: Parallel` and replace the inline preStop block with `/scripts/galera-prestop.sh`
- add bounded shutdown ordering: lower ordinals wait for higher ordinals to stop, then still perform a bounded local shutdown after the ordering budget expires
- measure the wait with real elapsed time so peer probe latency is included in the 120-second termination budget
- distinguish a refused connection (listener stopped) from timeout, DNS failure, and other unreachable states; uncertain peers remain pending until the budget expires
- validate pod/peer identity and timeout configuration instead of silently treating malformed input as the highest ordinal
- persist degraded shutdown evidence in the data directory and mirror it to the main container log; if both evidence sinks fail, finish the local shutdown attempt and return a failed preStop hook
- publish `${DATA_DIR}/.galera-shutting-down` before the first Galera desync mutation, then run local cleanup as `wsrep_desync=ON`, `wsrep_on=OFF`, and bounded `mysqladmin shutdown`
- bump the MariaDB chart to `1.2.0-alpha.30` because earlier immutable ComponentDefinition names have already been used

This supersedes PR #3105 as the selected route. PR #3105 remains useful only as OrderedReady trade-off evidence and should not be merged.

## Cross-PR integration note

PR #3144's Galera shutdown watcher consumes `.galera-shutting-down` to avoid sending SIGKILL while this preStop hook is deliberately waiting or shutting down. This PR now publishes that marker before `wsrep_desync=ON`; #3144 clears a stale marker when the server starts. The marker write is harmless without #3144, but together they close the watcher/preStop race.

Both PRs modify the MariaDB ComponentDefinition/preStop area and may conflict mechanically. Preserve this ordering and the #3144 stale-marker cleanup when resolving or stacking them.

## Validation

Predecessor candidate: `7da449a056a8b6ede479c09c46d5e21ba8f85957`, based on the prior reviewed candidate `a141076bdfdf651d6d0301a7b8e05845bf599110`.

- focused `galera-prestop.sh` ShellSpec: 23 examples, 0 failures
- related parallel-shutdown template ShellSpec: 5 examples, 0 failures
- full local MariaDB ShellSpec: 700 examples, 0 failures, 9 existing pendings
- `bash -n`, ShellCheck severity=error, and `git diff --check`: pass
[...]

PR #3144, opened 2026-07-11
fix(mariadb): defer galera bootstrap on hard-crash grastate; backslash-safe sql_quote

weicao:
## Problem

Galera bootstrap treated DNS/TCP timeout or refusal as proof that a peer was absent. During a partition, a node carrying `safe_to_bootstrap=1` could therefore bootstrap a second writable Primary while the existing Primary was hidden.

Tightening those observations to `uncertain` closed that split-brain path, but exposed the orderly full-stop successor: all stopped peers are unreachable, so even the elected `safe_to_bootstrap=1` node deferred and the cluster could not restart.

## Fix

- Require explicit peer observations. Primary always wins; uncertainty defers; only an all-explicit-non-Primary set permits ordinary automatic bootstrap.
- Keep empty FQDN, DNS/TCP timeout or refusal, unreadable SQL state, and hard-crash `seqno=-1` fail-closed.
- Persist the existing preStop `.galera-shutting-down` marker on the same PVC and combine it with exact `safe_to_bootstrap=1` as the only orderly-full-stop uncertainty override. Marker-only, safe=0, no-marker, crash, and partition paths remain fail-closed.
- Atomically promote prior-generation `.galera-shutting-down` evidence to the distinct `.galera-bootstrap-pending` retry authority. This separates election evidence from the current generation's preStop/self-heal guard.
- Preserve retry-pending authority through election, defer/wait, both `exec` paths, entrypoint, and MariaDB startup. A failed launch therefore reuses the same exact safe-bit-plus-marker authority on the next container attempt.
- Complete retry-pending authority only after the watcher directly observes `wsrep_local_state=4` and `wsrep_cluster_status=Primary`. Missing or unremovable evidence fails closed; removal failure is reported and retried on the next watcher tick.
- Collapse a duplicate current-generation shutdown marker only while durable retry-pending authority remains present, with fail-closed deletion verification.
- Require `_wait_for_primary_peer` to observe both predicate rc=0 and `GALERA_PEER_OBSERVATION=primary` on the first probe and every retry.

This is a source-contract and unit-evidence fix. No live partition, package, runner, or Kubernetes recovery result is claimed here; those remain Test-owned.

## Evidence

- Post-launch lifecycle TDD RED: 50 examples, 8 failures before the source fix.
- Focused `galera_start_spec.sh`: Bash 3.2 `52/52`; Bash 5.3 `52/52`.
- Galera ShellSpec group: Bash 3.2 `77/77`; Bash 5.3 `77/77`.
- MariaDB scripts full discovery: 729 examples, 7 failures, 9 pending. All 7 failures are the independently reproduced baseline `replication_merged_replication_mode_env_wire_spec.sh` failures; changed-test delta is 0.
- `git diff --check`, Bash 3.2 syntax, and Bash 5.3 syntax: green.
- ShellCheck severity-error scan of the product script: green.
- Exact-head CI for the identity below: 14 SUCCESS, 4 expected SKIPPED, pending 0, red 0.

## Exact review identity

- head `550cbaa24ef71ef24ee30f0dcceddb14e74735c2`
- tree `85fb1b51c8d4f125e5bbe8ab4a4292ccd17e26d0`
- sole parent `999ba7e92c3d6a6148805cd06119850a354c1302`
- merge-base with `origin/main` `5fa53f531defe335cd044723ebb854e6f4de4dbe`

Relationship: #3108 covers ordered graceful shutdown; this PR covers crash/partition bootstrap admission and the retry-safe orderly-full-stop successor.


weicao:
Superseded by #3276. This PR's current head a408efd93fa194637324f19f734bdc6dc6085f40 is included in the consolidated candidate 7978812cf8ca066cfc13ac6baf0ddd78505e0965; further review and testing should use #3276.

weicao:

weicao (commented):
_any_peer_alive treats DNS failures and TCP timeouts/refusals as evidence that no peer is alive. During a partition, a node with safe_to_bootstrap=1 can start --wsrep-new-cluster while another Primary still exists, producing two writable clusters and divergent data.

leon-ape (commented):
[P1] Orderly full-cluster restart cannot elect the safe Galera node

`_any_peer_alive` reports no peers only when every remote is already accepting SQL and explicitly reports non-Primary. After an orderly full shutdown the peers are stopped or unreachable, so even the node marked `safe_to_bootstrap=1` defers as uncertain, while the other nodes wait for a Primary before starting MariaDB. The cluster therefore remains in restart/wait loops and cannot recover automatically after a clean full shutdown.

addons/mariadb/scripts/galera-start.sh
```diff
@@ -334,8 +508,14 @@ main() {
...
   ) &
 
+  local bootstrap_mode=0
   if should_bootstrap; then
+    bootstrap_mode=1
+  fi
+
+  if [ "${bootstrap_mode}" = "1" ]; then
     echo "Starting Galera cluster bootstrap (--wsrep-new-cluster)..."
+    _consume_prior_graceful_shutdown_evidence
```
leon-ape: **P1 — a failed bootstrap consumes the only retry authority.** This deletes `.galera-shutting-down` before `docker-entrypoint.sh` has created a Primary. If the entrypoint or MariaDB fails after the deletion, the next container still sees `safe_to_bootstrap=1` and stopped/uncertain peers but no marker, so it refuses to bootstrap; the orderly cluster returns to permanent no-Primary restart loops and requires manual recovery.

weicao: Fixed at exact head 550cbaa24ef71ef24ee30f0dcceddb14e74735c2 (tree 85fb1b51c8d4f125e5bbe8ab4a4292ccd17e26d0). Prior-generation `.galera-shutting-down` is now atomically promoted to durable `.galera-bootstrap-pending`; neither launch path consumes it. The watcher completes it only after directly observing `wsrep_local_state=4` and `wsrep_cluster_status=Primary`, so entrypoint/MariaDB failure preserves the same safe-bit-plus-marker retry authority. Duplicate preStop evidence and fail-closed cleanup paths are covered. Focused: Bash 3.2 52/52, Bash 5.3 52/52; Galera group 77/77 on both; full MariaDB 729 with the same 7 unrelated baseline failures and 9 pending. Exact-head CI is running.

leon-ape (approved):
LGTM

Issue #3148, opened 2026-07-11
[postgresql] postgres-setup.sh swallows generate_patroni_yaml.py failures — spilo launches with empty SPILO_CONFIGURATION (restore silently becomes empty initdb)


weicao:
## Problem

`scripts/postgres-setup.sh:119-123` launches spilo even when patroni-config generation fails:

```bash
python3 /kb-scripts/generate_patroni_yaml.py $tmp_patroni_yaml
# SPILO_CONFIGURATION is defined by spilo image
SPILO_CONFIGURATION=$(cat $tmp_patroni_yaml)
export SPILO_CONFIGURATION
exec /launch.sh init
```

No exit-code check, no `set -e` in this function. If the generator exits non-zero (malformed conf input → `yaml.safe_load` → `None` → `dict.update` TypeError; transient read error; future generator bug), the tmp file doesn't exist, `cat` fails, and spilo starts with `SPILO_CONFIGURATION=""` — silently dropping `custom_conf` (the entire KB parameter template), the `pg_hba` template, the `kb_restore`/`kb_restore_from_time` bootstrap method, and all `bootstrap.dcs` tuning.

## Effect

- On a **restore** pod: patroni falls back to `initdb` — the restore "succeeds" as a fresh empty database (silent data loss, the worst failure shape).
- On a normal pod: config drift (wrong max_connections, missing hba entries) that surfaces far from the root cause.

Every future generator bug becomes an undiagnosable config-drop instead of a clean CrashLoop. The integrated PR set made the generator load-bearing for restore correctness (it injects `bootstrap.method=kb_restore`), raising the blast radius.

## Fix

```bash
python3 /kb-scripts/generate_patroni_yaml.py "$tmp_patroni_yaml" || { echo "FATAL: patroni config generation failed" >&2; exit 1; }
[ -s "$tmp_patroni_yaml" ] || { echo "FATAL: generated patroni config is empty" >&2; exit 1; }
```

Fail-closed: the pod CrashLoops with an attributable log line instead of running misconfigured.

Contract anchor: silent-fallback class per addon design-contract review guidance; docs otherwise silent. Found in the 2026-07-12 convergence sweep (pre-exists on main). Severity: **medium** (latent; catastrophic shape when triggered on restore paths).

Issue #3146, opened 2026-07-11
[postgresql] standby (DR) cluster replication auth broken by construction: remote credentials collected into dead vars, spilo uses the local password


weicao:
## Problem

Cross-cluster standby (DR) provisioning fails replication auth **by construction**: the remote cluster's credentials are collected but never used.

- `templates/cmpd.yaml:163-174` declares `KB_PGUSER_STANDBY` / `KB_PGPASSWORD_STANDBY` from the `remote-instances` serviceRef — **no script, template, or env consumes them** (repo-wide grep: declarations only).
- Meanwhile `templates/cmpd.yaml:414-417` wires spilo's actual standby auth to the **local** random password:

```yaml
- name: PGUSER_STANDBY
  value: standby
- name: PGPASSWORD_STANDBY
  value: $(POSTGRES_PASSWORD)
```

- The cluster chart dutifully feeds remote credentials into the ServiceDescriptor (`addons-cluster/postgresql/templates/remote-service-descriptor.yaml` `auth.username/password` from `standbyAccountSecret`) — they flow into the two dead vars and are silently dropped.

## Effect

Standby cluster B pointing at primary cluster A (different random passwords, which is the default) loops on `password authentication failed for user "standby"` during basebackup/streaming, never bootstraps, never publishes a `primary`/`standby_leader` role, and the cluster never becomes Available. Any DR/standby acceptance case reds at provision.

## Fix

Consume the vars: when `STANDBY_HOST` is set and `KB_PGPASSWORD_STANDBY` is non-empty, export `PGPASSWORD_STANDBY=$KB_PGPASSWORD_STANDBY` (and `PGUSER_STANDBY=$KB_PGUSER_STANDBY`) in `scripts/postgres-setup.sh` before `exec /launch.sh` — overriding the local-password default. Alternatively remove the dead vars and document the cross-cluster same-password requirement (strictly worse).

## Contract anchor

addon-docs `06-variables-and-services.md` (serviceRefVarRef semantics); docs silent on cross-cluster credential requirements.

Found in the 2026-07-12 convergence sweep (pre-exists on main). Severity: **high** for any DR-lane acceptance case.

PR #3152, opened 2026-07-11
fix(postgresql): postgres-setup.sh hardening — pending-restart starvation, standby DR auth, config-generation fail-closed

### LINKED ISSUE #3145: [postgresql] pending-restart loop starves secondaries for all but 6 whitelisted parameters — restart-class reconfigure never completes on multi-replica clusters
## Problem

The background pending-restart loop in `scripts/postgres-setup.sh` — the only mechanism that restarts **secondaries** after a restart-class reconfigure (the reconfigure action targets the primary only: `_helpers.tpl` `targetPodSelector: Role / matchingKey: primary`, and patroni `POST /restart` is member-local) — has three defects. The first one makes most restart-class reconfigures on multi-replica clusters permanently incomplete.

### 1. [high] Six-parameter whitelist starves every other restart-class parameter

`scripts/postgres-setup.sh:16`:

```bash
sql="select setting from pg_settings where name in ('max_connections','max_locks_per_transaction','max_worker_processes','max_prepared_transactions','wal_level','track_commit_timestamp')"
```

and the gate at `:62-72`: once the leader is no longer `pending_restart`, a secondary restarts **only if one of those six probed values differs from the primary**. For every other parameter in `reloader/restart-parameter.yaml` (`shared_buffers`, `max_wal_senders`, `max_replication_slots`, `archive_mode`, `shared_preload_libraries`, `logging_collector`, `huge_pages`, …) the probes are equal, the loop `continue`s forever, and the secondary keeps the old value with `pending_restart: true` indefinitely.

**Repro:** 2+ replica cluster → Reconfigure `shared_buffers` → primary shows the new value; every secondary shows the old value and stays pending-restart until an unrelated restart. Any suite asserting the new value on all pods (or no pod pending-restart) reds.

**Fix:** drop the whitelist comparison; use the node's own signal — `select count(*) from pg_settings where pending_restart` (PG ≥9.5) or patroni's member `pending_restart` flag — keeping the existing leader-first ordering.

### 2. [low] Both secondaries restart simultaneously

Both pass the gate in the same ~5s polling window → a window with zero live replicas (and blocked commits if synchronous mode is ever enabled). Fix: serialize by ordinal or check no sibling is mid-restart.

### 3. [low] The loop's restart call has no timeout

`postgres-setup.sh:79` `curl -XPOST .../restart` without `-m`: a hung patroni wedges the loop permanently and secondary auto-restart silently stops. One-line fix.

## Contract anchors

- addon-docs `10-day2-operations.md:65` ("Reconfigure 成功，不代表所有 Pod 的动态参数一致")
- addon-docs `08-parameters-and-config.md:101` ("不能用'一个 Pod reload 成功'代表整个 Component 参数一致")

Found in the 2026-07-12 convergence sweep (integrated tree main 554fd848 + 11 open PRs; defect pre-exists on main). Severity: **high** (deterministic acceptance red for multi-replica restart-class reconfigure).

weicao:
Fixes #3145, #3146, #3148 — three defects in the same script, one commit per concern.

- **#3145**: the pending-restart loop compared a 6-parameter whitelist against the primary and skipped the restart otherwise, starving every other restart-class parameter (shared_buffers etc.) — secondaries stayed pending_restart with the old value forever. Patroni's own per-member pending_restart flag (already checked) is the ground truth; the whitelist gate is removed. Bonus in the same commit: member restarts are serialized (no zero-live-replica window) and the restart POST is bounded (-m 30).
- **#3146**: standby (DR) clusters exported remote credentials into two dead vars while spilo authenticated with the local random password — DR provisioning failed auth by construction. The remote credentials now override when present.
- **#3148**: generate_patroni_yaml.py failures were swallowed and spilo launched with an empty SPILO_CONFIGURATION (on restore pods: silent empty initdb reported as success). Now fails closed with an attributable FATAL line.

postgres_setup_spec.sh 7/7 green (bash 5.3); bash -n clean. Draft until runtime-tested per team convention.

### DIFF
--- a/addons/postgresql/scripts-ut-spec/postgres_setup_spec.sh
+++ b/addons/postgresql/scripts-ut-spec/postgres_setup_spec.sh
@@ -117,7 +117,7 @@ Describe "PostgreSQL Initialization Script Tests"
     End
   End
 
-  Describe "need_restart_for_pending()"
+  Describe "pending restart candidate selection"
     setup() {
       CURRENT_POD_NAME="pg-cluster-postgresql-0"
     }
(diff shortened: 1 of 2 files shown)


weicao:
## `pending_restart` concurrent follower restart fix

Observed failure: with one leader and two replicas both reporting `pending_restart=true` in the same arbitration window, the previous script let both replicas independently POST their local Patroni `/restart`. The deterministic pre-fix test ran both follower decisions and recorded two callers (`pg-cluster-postgresql-1`, `pg-cluster-postgresql-2`), so this is a reproduced concurrency defect rather than preventive hardening.

Fix at exact head `f1f5fcc42433da996a3667013cca1b7fcde3d873`:
- select one cluster-wide candidate from the Patroni snapshot: pending healthy leader first, otherwise healthy pending replicas in stable member-name order;
- select no candidate while any member is already `restarting`;
- after the existing 5-second arbitration delay, re-read the whole cluster and recompute the candidate;
- POST only when the current pod is that non-empty candidate.

A/B test boundary using kubeblocks-tests PR #816 exact head `2aed946f2eaf45c2e655b976731f531b41d196d0`:
- old product `a730c41eb406efae0a5c9384186c65c83e3f722d`: 5/5 RED, each run `PASS=2 FAIL=1 SKIP=0`, both followers decided and both POSTed;
- new product `f1f5fcc42433da996a3667013cca1b7fcde3d873`: 5/5 GREEN, each run `PASS=3 FAIL=0 SKIP=0`, both followers decided, only `pg-cluster-postgresql-1` POSTed;
- independent macOS Bash 3.2 reproduction: old 3/3 RED, new 3/3 GREEN, with zero jq/temp-file/TSV corruption noise.

Product verification: focused ShellSpec 10/0, all PostgreSQL script specs 23/0, `bash -n` and diff checks pass. Fresh PR CI is terminal green across shell-check, ShellSpec, generated files, chart checks, and the superseding `nopick` label check (8/8 current checks).

Scope boundary: this proves the offline three-member arbitration contract and product script/unit checks. It is not a full Kubernetes runtime regression or release-readiness claim.

weicao:

PR #3155, opened 2026-07-11
fix(mysql): fail closed on orchestrator API ambiguity

weicao:
## Problem

Static review found fail-open and contract defects in the MySQL Orchestrator integration:

- direct cluster-alias registration constructed `http://${ORC_ENDPOINTS}` although the API port is supplied separately;
- orchestrator-client accepted HTTP 4xx/5xx and JSON `null` as successful responses;
- preTerminate forgot by a short Pod name, continued after Orchestrator errors, and could report success from an empty or unreachable readback.

## Changes

- normalize the Orchestrator API base and use bounded fail-on-HTTP-error alias requests;
- require curl HTTP success plus non-null JSON in both packaged orchestrator-client copies;
- make preTerminate bounded and alias-based, prove Orchestrator reachability, and verify the cluster name/alias is absent before success;
- add focused ShellSpec negative and positive cases.

## Validation

- ShellSpec: 8 examples, 0 failures
- `bash -n` on touched scripts: PASS
- `git diff --check`: PASS
- full chart render: PASS
- exact rebased-head PR CI: terminal green

## Evidence Boundary

The branch is rebased onto current `main` at exact head `64b9be0c6abe270b125de0a9055ae4f9956165dd`. This is source, unit, syntax, and render evidence. Runtime N=0; it does not establish MySQL acceptance, full-suite PASS, or release readiness.

### DIFF
--- a/addons/mysql/orc-tools/orchestrator-client.sh
+++ b/addons/mysql/orc-tools/orchestrator-client.sh
@@ -275,7 +275,7 @@ function api {
   api_call_result=0
   if [[ ${curl_auth_params} != "401 Unauthorized" ]]; then
     for sleep_time in 0.1 0.2 0.5 1 2 2.5 5 0 ; do
-      api_response=$(curl ${curl_auth_params} -s "$uri" | jq '.')
+      api_response=$(curl ${curl_auth_params} -fsS "$uri" | jq -ce 'select(. != null)')
       api_call_result=$?
       [ $api_call_result -eq 0 ] && break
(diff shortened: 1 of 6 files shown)


weicao:
Rebased directly onto latest main b55707f0. New head: 64b9be0c. No conflict. Post-rebase bash syntax on all touched scripts, helm lint, full chart render, and git diff --check pass. The PR base is now main; new CI is the remaining automated gate.

PR #3157, opened 2026-07-11
fix(mysql): harden PITR binlog cleanup state

weicao:
## Problem

The MySQL PITR cleanup integration contains three safety/availability defects:

1. under actionset-wide `set -e`, a standalone `synced_binlogs=$(get_synced_binlogs)` exits the archiver before the intended fail-closed guard can skip purge;
2. `min_synced_file` leaks across repeated cleanup calls, allowing a later empty replica discovery to reuse stale state;
3. uploaded-file detection filters by `DP_TARGET_POD_NAME`, which can remain on the old primary after switchover, hiding all uploaded files and disabling cleanup.

## Changes

- place the failing assignment in an `if ! ...` condition;
- reset `min_synced_file` locally on every invocation;
- filter uploaded files by the live `LOG_PREFIX` derived from MySQL `log_bin_basename`.

Commits are separate for focused review and rollback:
- `d315627b` set-e and stale-state guard
- `591594e2` live-prefix matching

## Validation

- `bash -n addons/mysql/dataprotection/mysql-pitr-backup.sh`: PASS
- set-e regression harness: emits `guarded` then `alive`
- uploaded-prefix harness: old-primary path excluded, `actual-writer-bin.000002` selected
- `git diff --check`: PASS

## Evidence boundary

Static/render evidence only; focused runtime N=0. This does not establish PITR runtime PASS, fullsuite PASS, or release readiness.

### DIFF
--- a/addons/mysql/dataprotection/mysql-pitr-backup.sh
+++ b/addons/mysql/dataprotection/mysql-pitr-backup.sh
@@ -193,6 +193,10 @@ cleanup_mysql_binlogs() {
 
     # Get synced binlog files from all replicas
     function get_synced_binlogs() {
+        # Reset per call: this is invoked repeatedly in the long-lived archiver
+        # process, and a leaked value from a previous call would let a purge
+        # proceed on a stale synced position when replica discovery fails.
+        local min_synced_file=""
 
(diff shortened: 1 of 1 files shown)


weicao:
Superseded by main-based PR #3088. The set-e and stale-state guards were already present there; the remaining live-prefix matching change was appended at head 774786e5. This integration-based PR is being closed.

PR #3159, opened 2026-07-11
fix(mysql): make xtrabackup restore retries explicit

weicao:
## Problem

Both full and incremental xtrabackup restore scripts always enter download/prepare/`--move-back`, even when a DataProtection action is retried. A retry after success fails unnecessarily; a retry after a partial move-back can operate on a non-empty target without classifying the interrupted state.

## Contract

- completion marker present: idempotent success, keep existing data;
- empty target (ignoring filesystem `lost+found`): proceed;
- non-empty target without marker: fail closed and preserve the partial-restored evidence.

The guard runs before download/prepare/move-back in both scripts.

## Validation

- ShellSpec: 7 examples, 0 failures (full + incremental; empty/completed/partial/lost+found)
- `bash -n` touched scripts: PASS
- `git diff --check`: PASS

## Evidence boundary

Static/unit/render evidence only; restore runtime N=0. This does not establish restore PASS, fullsuite PASS, or release readiness.

### DIFF
--- a/addons/mysql/dataprotection/restore.sh
+++ b/addons/mysql/dataprotection/restore.sh
@@ -3,6 +3,28 @@ set -e
 set -o pipefail
 export PATH="$PATH:$DP_DATASAFED_BIN_PATH"
 export DATASAFED_BACKEND_BASE_PATH="$DP_BACKUP_BASE_PATH"
+
+restore_preflight() {
+  local residue
+
+  mkdir -p "${DATA_DIR}"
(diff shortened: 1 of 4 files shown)


weicao:
Rebased directly onto latest main b55707f0. New head: 8f0ed519. No conflict. Post-rebase bash syntax, helm lint, full chart render, and git diff --check pass. The PR base is now main; new CI is the remaining automated gate.

PR #3160, opened 2026-07-11
fix(mysql): make redo capacity constraint optional

weicao:
## Problem

The integration schema requires `innodb_redo_log_capacity`, while the MySQL 8.0 template only renders it when a container memory limit is available. A no-memory-limit deployment can therefore fail configuration validation before startup.

## Change

Keep the type/range validation for user-supplied values, but make the field optional. This matches the template contract: memory-known derives a value; memory-unknown omits it.

## Validation

- machine check: exactly one optional declaration and no required declaration
- default no-memory render omits `innodb_redo_log_capacity`
- `git diff --check`: PASS

## Evidence boundary

Static/render evidence only; no-memory-limit runtime N=0. This does not establish cluster startup PASS, fullsuite PASS, or release readiness.

### DIFF
--- a/addons/mysql/config/mysql8-config-constraint.cue
+++ b/addons/mysql/config/mysql8-config-constraint.cue
@@ -583,8 +583,8 @@
 	// Controls encryption of redo log data for tables encrypted using the InnoDB tablespace encryption feature.
 	innodb_redo_log_encrypt: string & "OFF" | "ON" | *"OFF"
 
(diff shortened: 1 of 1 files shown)


weicao:
Superseded by main-based PR #3084. The optional redo-capacity constraint from this PR was appended to #3084 at head e59e2216, so this integration-based PR is being closed.

PR #3161, opened 2026-07-11
fix(postgresql): share boolean parameter schema

### LINKED ISSUE #3124: [postgresql] Minor-findings batch from 2026-07-11 main-head review (restore robustness / CUE fidelity / HA lifecycle)
## Scope

Batch memo for the minor findings from the 2026-07-11 main-head review (head 554fd848), to be split into focused PRs by root cause. Each item is independently fixable; none is release-blocking on its own.

### Restore robustness

1. **pgdump restore: artifact pull failure is not fatal.** `dataprotection/pgdump-restore.sh` sets only `set -o pipefail` (no `set -e`), so a mid-stream failure of `datasafed pull | tar -xf` continues into database/schema creation and `pg_restore` against a truncated `-Fd` dir; a truncated-but-toc-complete dir yields a partial restore maskable by the `errors ignored on restore` grep (lines 76-78). Fix: `set -e` before the pull (drop to `set +e` only around the conflict-policy section, as already structured at line 68). Anchor: `09a-backup-basic.md:111`.

2. **Restore-path constants inconsistently derived.** `templates/actionset-postgresql-pitr.yaml:11` hardcodes `/home/postgres/pgdata` while `actionset-wal-g-pitr.yaml:12` templates `{{ .Values.dataMountPath }}`; `wal-g-restore.sh:147` / `wal-g-archive-restore.sh:101` hardcode `WALG_DIR=/home/postgres/pgdata/wal-g` while the same scripts derive `walg_dir=${VOLUME_DATA_DIR}/wal-g`. Consistent today only because `values.yaml:153` pins the path; overriding `dataMountPath` breaks the rendered `restore_command`. Fix: derive everywhere.

3. **`kb_restore.signal` protocol inconsistent across restore flavors.** wal-g flavors create the signal + `--replica`-aware hook; `postgresql-pitr-restore.sh` never does, so `postgres-setup.sh` chown gating and `replica_restore.conf` injection are skipped for that flavor — two different marker protocols for the same bootstrap hook. Fix: emit the same signal + replica-aware hook from the PITR flavor (single shared heredoc).

### CUE schema fidelity

4. **Boolean-typed params reject PostgreSQL's `on/off/yes/no/1/0` spellings.** ~80 params per file are plain `bool` (pg14:47 `autovacuum`, :302 `fsync`, :1085 `wal_recycle`), while `wal_init_zero` (pg14:1109) is `string & "on" | "off"` — the same inconsistency PR #3115 hit, in reverse. pg14's own template default uses a third spelling (`pg14-config.tpl:279 wal_compression = 'True'`). Fix: one shared `#PgBool` union across all six files.

5. **`vacuum_cost_delay` / `autovacuum_vacuum_cost_delay` typed `int`** (pg14:1022/:68 and identical in all majors) although PG12+ made them floating-point precisely so users can set the documented `0.5`–`2ms` values. Fix: `number & >=0 & <=100` (and `>=-1` for the autovacuum variant).

6. **`geqo_pool_size` carries `@storeResource()`** (pg14:317) although it is a unitless count (genetic-population individuals) — a size annotation invites `4KB`-style unit conversion silently distorting the value. Fix: drop the annotation in all six files. (All other ~40 unit annotations spot-checked clean.)

### HA / lifecycle

7. **No `memberLeave` action and no documented "not needed" statement.** `templates/cmpd.yaml:235-267` has only roleProbe/switchover/accountProvision; the RBAC comment at :280-281 shows `patronictl remove` was contemplated but never wired. With `synchronous_mode` on, scaling in the current sync standby blocks commits on the primary up to ttl/loop_wait (30s/10s), and etcd-DCS mode leaves a stale member key until TTL expiry. Fix: add a memberLeave action (rotate sync state via patroni REST, `patronictl remove` for etcd mode) — or an explicit README statement. Anchors: `10-day2-operations.md:39,60`, `05c-lifecycle-action-contracts.md:30`.

8. **Demoted-primary exit path uploads leftover WAL with no overwrite guard.** `wal-g-archive.sh:211-218` / `postgresql-pitr-backup.sh:128-135`: the role-mismatch interlock is correct, but the exit-path `uploadMissingLogs` can collide with the new primary archiving the same-named final segment of the old timeline (no `WALG_PREVENT_WAL_OVERWRITE`; the pitr flavor's `datasafed push` overwrites unconditionally) — last writer wins, potentially poisoning the PITR chain at the failover boundary. Fix: set `WALG_PREVENT_WAL_OVERWRITE=true` and make the exit-path upload skip already-present segments. Anchor (partial): `09b-backup-extensions.md:43`.

### Already resolved elsewhere

- *pg_basebackup legacy branch falling through to `exit 0` with no artifact (empty cluster masquerading as restored)* — *fixed by the `restored_base_archive` guard in PR #3112 head 2e342395; listed here only for review-trail completeness.*

weicao:
## Problem

Issue #3124 item 4 found that PostgreSQL boolean settings are represented inconsistently in the CUE constraints:

- many PostgreSQL boolean GUCs are plain `bool` or `bool & false | true`;
- `wal_init_zero` and several extension booleans are string-only `on/off` enums;
- PostgreSQL accepts boolean spellings such as `on/off`, `true/false`, `yes/no`, and `1/0`, case-insensitively.

Using a CUE union such as `bool | string` is not sufficient here: the KB parameter validator sees the bool branch first and classifies the field as BoolType, so `on/off/yes/no` still fail. The shared type must therefore stay StringKind-only.

## Fix

- Add a shared StringKind-only `#PgBool` regex to each PostgreSQL major constraint file: `string & =~"(?i)^(on|off|true|false|yes|no|0|1)$"`.
- Convert CUE boolean fields across pg12/14/15/16/17/18 to `#PgBool`.
- Convert exact `string & "on" | "off"` boolean fields to `#PgBool`.
- Convert schema defaults like `#PgBool | *true` to quoted string defaults like `#PgBool | *"true"` so the default does not poison the kind mask back to BoolType.
- Preserve non-boolean three-valued enums such as `huge_pages = on/off/try` and `pgtle.enable_password_check = on/off/require`.
- Add a ShellSpec guard that rejects a bool branch, rejects unquoted bool defaults, and spot-checks existing template defaults such as `wal_init_zero = off`, `cron.log_statement = 'on'`, `index_adviser.enable_log = 'on'`, and `remove_temp_files_after_crash = 'on'`.

## Validation

- `shellspec --load-path ./shellspec --shell bash addons/postgresql/scripts-ut-spec/pg_bool_schema_spec.sh`
- `shellspec --load-path ./shellspec --default-path addons/postgresql/scripts-ut-spec --shell bash`
- `helm template pg addons/postgresql >/tmp/pg-bool-schema-render-v2.yaml`
- `helm lint addons/postgresql`
- `git diff --check`

## Runtime status

Draft: schema/render/static checks only. No runtime reconfigure acceptance yet.

Fixes #3124
Related #3115

### DIFF
--- a/addons/postgresql/config/pg12-config-constraint.cue
+++ b/addons/postgresql/config/pg12-config-constraint.cue
@@ -1,9 +1,11 @@
 // PostgreSQL parameters: https://postgresqlco.nf/doc/en/param/
+#PgBool: string & =~"(?i)^(on|off|true|false|yes|no|0|1)$"
+
 #PGParameter: {
 	// Allows tablespaces directly inside pg_tblspc, for testing, pg version: 15
-	allow_in_place_tablespaces?: bool
+	allow_in_place_tablespaces?: #PgBool
 	// Allows modification of the structure of system tables as well as certain other risky actions on system tables. This is otherwise not allowed even for superusers. Ill-advised use of this setting can cause irretrievable data loss or seriously corrupt the database system.
-	allow_system_table_mods?: bool
+	allow_system_table_mods?: #PgBool
 	// Sets the application name to be reported in statistics and logs.
 	application_name?: string
 	// Sets the shell command that will be called to archive a WAL file.
@@ -15,13 +17,13 @@
 	// (s) Forces a switch to the next xlog file if a new file has not been started within N seconds.
 	archive_timeout: int & >=0 & <=2147483647 | *300 @timeDurationResource(1s)
 	// Enable input of NULL elements in arrays.
-	array_nulls?: bool
+	array_nulls?: #PgBool
 	// (s) Sets the maximum allowed time to complete client authentication.
 	authentication_timeout?: int & >=1 & <=600 @timeDurationResource(1s)
 	// Use EXPLAIN ANALYZE for plan logging.
-	"auto_explain.log_analyze"?: bool
+	"auto_explain.log_analyze"?: #PgBool
 	// Log buffers usage.
-	"auto_explain.log_buffers"?: bool & false | true
+	"auto_explain.log_buffers"?: #PgBool
 	// EXPLAIN format to be used for plan logging.
 	"auto_explain.log_format"?: string & "text" | "xml" | "json" | "yaml"
 
@@ -29,22 +31,22 @@
 	"auto_explain.log_min_duration"?: int & >=-1 & <=2147483647 @timeDurationResource()
 
 	// Log nested statements.
-	"auto_explain.log_nested_statements"?: bool & false | true
+	"auto_explain.log_nested_statements"?: #PgBool
 
 	// Collect timing data, not just row counts.
-	"auto_explain.log_timing"?: bool & false | true
+	"auto_explain.log_timing"?: #PgBool
 
 	// Include trigger statistics in plans.
-	"auto_explain.log_triggers"?: bool & false | true
+	"auto_explain.log_triggers"?: #PgBool
 
 	// Use EXPLAIN VERBOSE for plan logging.
-	"auto_explain.log_verbose"?: bool & false | true
+	"auto_explain.log_verbose"?: #PgBool
 
 	// Fraction of queries to process.
 	"auto_explain.sample_rate"?: float & >=0 & <=1
 
 	// Starts the autovacuum subprocess.
-	autovacuum?: bool
+	autovacuum?: #PgBool
 
 	// Number of tuple inserts, updates or deletes prior to analyze as a fraction of reltuples.
 	autovacuum_analyze_scale_factor: float & >=0 & <=100 | *0.05
@@ -104,7 +106,7 @@
 	bytea_output?: string & "escape" | "hex"
 
 	// Check function bodies during CREATE FUNCTION.
-	check_function_bodies?: bool & false | true
+	check_function_bodies?: #PgBool
 
 	// Time spent flushing dirty buffers during checkpoint, as fraction of checkpoint interval.
 	checkpoint_completion_target: float & >=0 & <=1 | *0.9
@@ -149,10 +151,10 @@
 	"cron.database_name"?: string
 
 	// Log all jobs runs into the job_run_details table
-	"cron.log_run"?: string & "on" | "off"
+	"cron.log_run"?: #PgBool
 
 	// Log all cron statements prior to execution.
-	"cron.log_statement"?: string & "on" | "off"
+	"cron.log_statement"?: #PgBool
 
 	// Maximum number of jobs that can run concurrently.
 	"cron.max_running_jobs": int & >=0 & <=100 | *5
@@ -170,22 +172,22 @@
 	datestyle?: string
 
 	// Enables per-database user names.
-	db_user_namespace?: bool & false | true
+	db_user_namespace?: #PgBool
 
 	// (ms) Sets the time to wait on a lock before checking for deadlock.
 	deadlock_timeout?: int & >=1 & <=2147483647 @timeDurationResource()
 
 	// Indents parse and plan tree displays.
-	debug_pretty_print?: bool & false | true
+	debug_pretty_print?: #PgBool
 
 	// Logs each queries parse tree.
-	debug_print_parse?: bool & false | true
+	debug_print_parse?: #PgBool
 
 	// Logs each queries execution plan.
-	debug_print_plan?: bool & false | true
+	debug_print_plan?: #PgBool
 
 	// Logs each queries rewritten parse tree.
-	debug_print_rewritten?: bool & false | true
+	debug_print_rewritten?: #PgBool
 
 	// Sets the default statistics target.
 	default_statistics_target?: int & >=1 & <=10000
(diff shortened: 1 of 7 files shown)


weicao:
## REQUEST CHANGES — the bool branch defeats the PR's purpose and regresses provisioning; third occurrence of the same validator trap

**1. `#PgBool: bool | string & (...)` is still classified BoolType — `on/off/yes/no` remain rejected for all ~120 converted params.**

This is the exact mechanism that failed PR #3115 twice: `IncompleteKind()` for a bool|string disjunction is `BoolKind|StringKind`, and `pkg/parameters/validate/cue_visitor.go:57-66` checks BoolKind **before** StringKind → field classified `BoolType` → `strconv.ParseBool` transform → `on`/`off`/`yes`/`no` all error. The string enum branch is unreachable dead weight. The PR's goal (accept PostgreSQL's native spellings) is not achieved for a single parameter.

**2. Worse: converting previously-correct pure-string params breaks whole-file validation → provisioning regression.**

On main these were `string & "on" | "off"` (StringKind-only, worked): `wal_init_zero`, `cron.log_run`, `cron.log_statement` (verified; please also audit `index_adviser.enable_log`, `remove_temp_files_after_crash` and any other conversion whose tpl default is `on`/`off`-spelled). Their rendered defaults exist in the templates:

```
pg14-config.tpl:83   cron.log_statement = 'on'
pg14-config.tpl:280  wal_init_zero = off
pg14-config.tpl:134  index_adviser.enable_log = 'on'
pg14-config.tpl:313  remove_temp_files_after_crash = 'on'
```

With BoolType classification, `ParseBool("on")` fails and **whole-rendered-file validation fails on every major that renders any of these defaults** — cluster creation and every reconfigure break. Same regression shape as #3115's second head.

**3. Defaults carry their own bool literal:** e.g. pg14:1063 `wal_compression: #PgBool | *true` — if `#PgBool` is fixed to a pure string type, a `*true` (bool) default re-poisons the kind mask for that field. Defaults must be string literals (`*"true"` / `*"on"`).

**4. Capitalization coverage:** templates also use `'True'`/`'False'` spellings (dozens of lines, e.g. pg14:48 `auto_explain.log_timing = 'True'`). ParseBool accepted them; a lowercase-only enum won't. Any pure-string design must cover the case variants.

## Suggested fix (covers all four points)

```cue
#PgBool: string & =~"^(?i)(on|off|true|false|yes|no|0|1)$"
```

- StringKind-only → StringType → no ParseBool; CUE regex validates.
- `(?i)` covers `True/False/ON/Off/...` — everything PostgreSQL itself accepts (PG bools are case-insensitive), including every spelling currently in the six templates.
- Replace every `#PgBool | *<bool-literal>` default with a string literal default.
- Make `pg_bool_schema_spec.sh` assert **no bool branch exists** (regression guard — grep `#PgBool:.*\bbool\b` must fail) and behaviorally cross-check that every bool-family tpl default in all six majors matches the `#PgBool` pattern, instead of pinning the current (broken) definition text.

## Sequencing

PR #3115 (head ad9587c8) already fixes `wal_init_zero` as a pure string enum on the same lines this PR touches — guaranteed conflict. Suggest #3115 lands first (it is the minimal fix already three-way reviewed) and this PR rebases and absorbs it into `#PgBool`.

Happy to re-review quickly on the next head — the `(?i)` regex form kills this trap class permanently, which is exactly the value of doing the shared pattern.

weicao:
Addressed the REQUEST CHANGES from review comment 4948010384 in new head `66ae2b1d`:

- Changed `#PgBool` to StringKind-only: `string & =~"(?i)^(on|off|true|false|yes|no|0|1)$"`.
- Removed the bool branch entirely so KB CUE visitor cannot classify these fields as BoolType.
- Changed schema defaults from `*true` / `*false` to quoted `*"true"` / `*"false"` so defaults do not poison the kind mask.
- Strengthened `pg_bool_schema_spec.sh` to reject a bool branch, reject unquoted bool defaults, and spot-check existing template defaults (`wal_init_zero = off`, `cron.log_statement = 'on'`, `index_adviser.enable_log = 'on'`, `remove_temp_files_after_crash = 'on'`).

Local validation rerun:
- `shellspec --load-path ./shellspec --shell bash addons/postgresql/scripts-ut-spec/pg_bool_schema_spec.sh` -> 6/0
- `shellspec --load-path ./shellspec --default-path addons/postgresql/scripts-ut-spec --shell bash` -> 17/0 with existing bash-version skips
- `helm template pg addons/postgresql >/tmp/pg-bool-schema-render-v2.yaml`
- `helm lint addons/postgresql`
- `git diff --check`

CI note: added `nopick`; fresh label-check passed. Shellspec CI is still pending at this readback.

weicao:
## Re-review at 66ae2b1d — static APPROVE, all REQUEST-CHANGES points addressed

Independently verified on the new head:

1. `#PgBool: string & =~"(?i)^(on|off|true|false|yes|no|0|1)$"` in all six cue files — pure StringKind, no bool branch anywhere (repo-grep confirms no remaining `: bool` fields outside the definition).
2. Zero bool-literal defaults; 13-14 quoted-string defaults per file (`| *"true"` form).
3. Multi-value enums correctly preserved and NOT converted: `huge_pages` (on/off/try), `pgtle.enable_password_check` (on/off/require), pg15+ `wal_compression` (pglz/lz4/zstd/on/off).
4. The previously-regressed template defaults now validate: `wal_init_zero = off`, `cron.log_statement = 'on'`, `index_adviser.enable_log = 'on'`, `remove_temp_files_after_crash = 'on'` all match the (?i) pattern, as do the `'True'/'False'` family.
5. `pg_bool_schema_spec.sh` ran 6/6 on my machine, and its guards are behavioral (forbids bool branch + bool-literal defaults, asserts enum preservation and template spot-checks) — no more asserting broken text.

This kills the validator trap class for the whole file family — exactly what the shared pattern should have done. Remaining notes (unchanged): sequencing vs PR #3115 (same wal_init_zero lines; #3115 first then rebase, or coordinate merge order), and draft-until-runtime per team convention (one live reconfigure of an on/off spelling through update-parameter.sh would be the ideal acceptance probe). Reference doc for future reviewers: kubeblocks-addon-docs docs/design/addon-cue-bool-string-union-classification-trap-guide.md.

weicao:

PR #3158, opened 2026-07-11
fix(mysql): fail close ORC prechecks within action budgets

weicao:
## Problem

Two ORC lifecycle paths can report success without trustworthy evidence or exceed the kbagent action window:

1. The switchover master precheck merges orchestrator-client retry stderr into stdout. A transient curl/jq error can become the parsed master token and make the script exit 0 as "not the master" without attempting switchover.
2. memberLeave runs three unbounded orchestrator-client calls. A hung client can be killed by the 60s clamp mid-action, and the old `instance ... || true` verification cannot distinguish a genuinely absent instance from a failed query.

### M1: fail closed on noisy master prechecks

- preserve stdout/stderr separately in the timeout fallback
- extract only the last non-empty `host:port` token from stdout
- use the same resolver in the initial master guard and post-switchover candidate discovery
- retain raw stdout/stderr/rc diagnostics when resolution fails

### M2: bound memberLeave and prove absence

- bound each orchestrator-client call to 10s with a 1s kill grace
- set memberLeave action timeout to 50s; worst case is 3 x 11s + 3s settle = 36s, leaving 14s for diagnostics
- fail rc=1 on forget, reachability, or verification timeout/error
- replace ambiguous `instance ... || true` with a successful bounded `all-instances` read and exact hostname absence check
- emit per-action phase and retry-safe diagnostics to action stderr and `/tmp/member-leave.log`

## Tests

- M1: retry stderr + final successful stdout; stderr-only noise; last valid stdout token
- M2: positive absent closure; forget timeout; reachability timeout; list verification timeout; instance still present; missing leaving-member identity

Local gates:

- MySQL ShellSpec: 46 examples, 0 failures (bash 3.2.57)
- focused M1+M2 ShellSpec: 19 examples, 0 failures
- bash-n: PASS
- shellcheck `--severity=warning`: PASS
- `helm dependency build addons/mysql`: PASS
- `helm lint addons/mysql`: PASS
- render: memberLeave `timeoutSeconds=50`, client timeout `10`, settle `3`, stderr mirrored
- `git diff --check`: PASS
- merge-tree: PASS, tree `6d86e28f3cb01fd7ca23f7f13251d6ae59aa6afd`

## Boundary

This PR is a draft child of #3117 and changes only M1/M2. It does not include PITR cleanup, ORC client/init/preTerminate, M3, or M4 work owned elsewhere. Evidence is code/static only; runtime N=0 and no acceptance or release-readiness claim is made.

PR #3162, opened 2026-07-11
fix(mysql): close consolidated acceptance review blockers

weicao:
## Base and scope

Exact base: `helios/mysql-review-consolidated-20260711@f163740e` (main + 12-fix integration + #3117).

This composition preserves owner commits as separate units and closes the deep-review findings before runtime acceptance:

- myloader restore uses DP-injected credentials;
- ProxySQL forwards termination signals without spawn race or double wait;
- PITR cleanup survives `set -e`, resets per-call state, and matches uploaded files by live binlog prefix;
- `innodb_redo_log_capacity` is optional when memory is unknown;
- ORC alias/API responses fail closed on wrong endpoint, HTTP errors, JSON null, and unreachable preTerminate;
- ORC switchover rejects noisy master tokens;
- ORC memberLeave uses bounded calls and positive all-instances absence proof;
- full/incremental xtrabackup retries are idempotent after completion and fail closed on partial targets.

## Local combined gate

Exact head: `a2a8796c`.

- MySQL full ShellSpec: **69 examples, 0 failures**
- focused/changed scripts `bash -n`: PASS
- changed scripts ShellCheck severity=error: PASS
- `git diff --check f163740e..HEAD`: PASS
- rendered memberLeave budget: 50s action, 3x10s clients + kill grace/settle <=36s

## Runtime boundary

Runtime N=0 on this combined exact head. This PR is not MySQL acceptance PASS, fullsuite PASS, or release readiness. Test execution must use the exact head and developer-specified focused scripts after this code/static gate is independently reviewed.

### DIFF
--- a/addons/mysql/config/mysql8-config-constraint.cue
+++ b/addons/mysql/config/mysql8-config-constraint.cue
@@ -583,8 +583,8 @@
 	// Controls encryption of redo log data for tables encrypted using the InnoDB tablespace encryption feature.
 	innodb_redo_log_encrypt: string & "OFF" | "ON" | *"OFF"
 
-	// The total capacity in bytes of the redo log.
-	innodb_redo_log_capacity: int & >= 8388608
+	// Optional because the template only derives it when a memory limit is available.
+	innodb_redo_log_capacity?: int & >= 8388608
 
 	// The replication thread delay (in ms) on a slave server if innodb_thread_concurrency is reached.
 	innodb_replication_delay?: int & >=0 & <=4294967295
--- a/addons/mysql/dataprotection/mysql-myloader.sh
+++ b/addons/mysql/dataprotection/mysql-myloader.sh
@@ -37,5 +37,5 @@ fi
 echo "parameters: $params"
 
 datasafed pull -d zstd-fastest "${DP_BACKUP_NAME}.mydumper.zst" - | myloader --stream  \
-  -h ${DP_DB_HOST} -u ${MYSQL_ADMIN_USER} -p ${MYSQL_ADMIN_PASSWORD} -P ${DP_DB_PORT} --regex '^(?!(kubeblocks\.kb_health_check$))' \
+  -h ${DP_DB_HOST} -u ${DP_DB_USER} -p ${DP_DB_PASSWORD} -P ${DP_DB_PORT} --regex '^(?!(kubeblocks\.kb_health_check$))' \
   ${params}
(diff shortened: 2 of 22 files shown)


weicao:
Independent composition review for Henry-owned M1/M2 scope at exact head `a2a8796ce95353b0c244fd3132d57571cd8d9950`: code/static composition gate CLOSED (runtime remains N=0). GitHub does not allow this shared account to approve its own PR, so recording the independent review as a comment.

Composition identity:
- `2957aa3c` (#3117 contract) is an ancestor of the consolidated head.
- M1/M2/M1-noise commits have stable patch IDs matching #3158 exactly.
- The six M1/M2 code/spec/template/value files are byte-identical between #3158 head `0bee0760` and this consolidated head.

Contract checks:
- stdout/stderr remain separated; only stdout is parsed.
- parser accepts the last K8s `hostname:port` token and rejects colon-shaped noise; missing/invalid token fails closed with rc/stdout/stderr.
- initial master guard and verify candidate lookup share the same resolver.
- memberLeave bounds forget/clusters/all-instances independently, fails closed on each call error, and closes only after a successful all-instances list proves exact hostname absence.
- rendered values are action timeout 50s, client timeout 10s, settle 3s; three 10s calls + three 1s kill-grace windows + 3s settle = 36s worst case, leaving 14s diagnostics budget.

Independent gates on this exact head:
- focused M1/M2 ShellSpec: 20/20 PASS
- MySQL full ShellSpec: 69/69 PASS
- changed bash-n, ShellCheck error-level, diff-check: PASS
- helm dependency build, lint, template: PASS
- GitHub: 7/7 SUCCESS, MERGEABLE/CLEAN

XP design-contract classes 1-8 walked: class 1 no silent fallback; class 2 non-empty master/leaving-member identities enforced; class 3 no stale multi-read composition change; class 4 empty/invalid master sentinel checked fail-closed; classes 5-7 no cleanup/NotFound/release path in this scope; class 8 no unsafe mixed boolean precedence introduced. No M1/M2 composition blocker found.

weicao:
Closing this PR because its base is `helios/mysql-review-consolidated-20260711`, not `main` or a release branch. Per repository PR policy, consolidated/integration branches may be used for local validation but not as formal PR bases. The composed commits remain preserved on the head branch; any still-needed change will be split or re-created from an allowed main/release baseline with the correct scope and evidence.

Issue #3149, opened 2026-07-11
[postgresql] reconfigure pipeline: multi-pair args dropped after the first, dynamic params misrouted to restart, no post-apply verification


weicao:
## Scope

Three defects in the reconfigure pipeline (`reloader/update-parameter.sh` + `reloader/restart-parameter.yaml` + the action wrapper), same subsystem, found in the 2026-07-12 convergence sweep (integrated tree; all pre-exist on main or in open-PR heads).

### 1. [medium] Multi-parameter reconfigure applies only the first key/value pair

The action wrapper passes exactly `"$1" "$2"` (`_helpers.tpl:231-235`) and `update-parameter.sh:53-55` consumes exactly one pair. Per the reconfigure parameter-passing contract, the controller may append multiple `[k,v]` pairs to the command; any pair beyond the first is never PATCHed into patroni — the value only reaches the pod via ConfigMap projection, which patroni does not watch, so it never applies until an unrelated restart. `SHOW param2` stays old → suite red.

Needs one runtime confirmation of the multi-pair dispatch shape, but the fix is correct under both semantics and cheap:

```bash
while [ $# -ge 2 ]; do collect "$1" "$2"; shift 2; done
# one combined PATCH /config; restart if ANY collected param is restart-class
```

### 2. [medium] Dynamic parameters misrouted into `restart-parameter.yaml` → gratuitous primary restart

Confirmed misroutes (PostgreSQL context ≠ postmaster): `wal_keep_size` (sighup), `wal_keep_segments` (sighup), `wal_compression` (superuser), `session_preload_libraries` (superuser), `autovacuum_max_workers` (sighup since PG17 — needs per-major split). Reconfiguring any of these triggers `POST /restart` on the primary: full connection drop and write outage for a parameter PostgreSQL applies via reload. Suites asserting "dynamic param applies without restart" or running traffic during reconfigure flake red.

Fix: remove sighup/superuser-context params from the restart list; split per-major where context changed across majors.

### 3. [low] Reload path never verifies the value applied

Success = PATCH 2xx + POST /reload 2xx; patroni applies DCS config on its HA loop (≤ loop_wait 10s). The action can report success while `SHOW` still returns the old value. Fix: bounded post-apply poll of `SHOW <param>` (~2× loop_wait) before exiting 0.

## Contract anchors

- addon-docs `addon-reconfigure-exec-parameter-passing-guide.md:40,60` (controller appends Arguments array elements)
[...]

PR #3165, opened 2026-07-11
fix(postgresql): remove reload-applicable parameters from restart-parameter.yaml

weicao:
Addresses the restart-misrouting item of #3149 (independent of the update-parameter.sh items, which land after #3025 merges to avoid same-file conflicts).

Removes `wal_keep_size`, `wal_keep_segments` (sighup), `wal_compression`, `session_preload_libraries` (superuser) from the restart routing — PostgreSQL applies all four via reload in every supported major, and the forced primary restart was pure downtime. `autovacuum_max_workers` stays with an explanatory comment (sighup only since PG17; shared file; PG17+ cost is an unnecessary restart, never a wrong value).

YAML validated; 4 removals confirmed programmatically. Draft until a reconfigure of one of these params is runtime-verified to reload without restart.

### DIFF
--- a/addons/postgresql/reloader/restart-parameter.yaml
+++ b/addons/postgresql/reloader/restart-parameter.yaml
@@ -1,6 +1,9 @@
 #parameters:
 - archive_mode
 - autovacuum_freeze_max_age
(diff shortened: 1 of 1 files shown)


weicao:

PR #3167, opened 2026-07-12
feat(valkey): restore sharded backups with source slots

weicao:
## Summary

- persist per-shard slot ownership and snapshot RDB SHA-256 in cluster backup metadata
- restore same-shard-count archives by rebuilding primary membership and reclaiming archived slots
- resolve restored peer FQDNs to validated IPv4 addresses before `CLUSTER MEET`
- require every observed cluster member to map uniquely to the configured KubeBlocks shard/pod roster
- require every designated restored primary to have a unique `CLUSTER MYID` before any meet or slot write
- prepare every restored non-first replica offline before Valkey accepts clients
- let each restored replica validate its PVC-local authorization and attach itself
- preserve a validated per-pod formed tombstone so repeated `All` iterations are zero-write successes

## Current exact head

- head: `fffda798a6bef4aa92cc1fc30d8167649e7f032e`
- tree: `db1bd981cae917c12a0889087006c1c03da33bdd`
- parent/base dependency: `c04f666c66ec2e98340b354600ce9f41e295924c` on `athena/valkey-cluster-topology`
- state: Draft and stacked; the parent is not in `main`

## First-red evidence

The parent head completed Backup but rejected cluster archives during Restore. Frozen packet outer SHA-256: `b03b7cba...98ab`.

Head `666ba2269b19351d237c80d7cc4258825e61ad9c` restored all six pods but passed peer FQDNs directly to `CLUSTER MEET`; Valkey rejected them and left three isolated primaries. Frozen packet outer SHA-256: `8757149e77e06e6adb587159c3844bc361b226dbdcaf014fee145bedac68a8e1`.

Head `4a890f6d90059652b44842e306505724960bbcf0` converged the three primaries, then exposed that restored ordinal-1 pods were still data-bearing singleton masters. Frozen packet outer SHA-256: `beffd04d6d033e8d9e4efe2a39b1c65f1ecbdc8bdb380f496fb23f7a05c79f84`.

Read-only follow-up found matching per-shard ordinal-0/1 base RDB and manifest SHA-256 values, with all six incremental AOF files at 0 bytes. Corrected packet outer SHA-256: `2683fb79794f44bb4ba3eca35d216fad4bbe111b5617bbcdb535692e3105cca5`.

Head `1c058ac5df103f2dd7d85f0b758efb5b462f132e` completed the engine path in one fresh CT11 run: 6/6 Ready, six-node membership, 16384 slots, replicas bound to intended parents, matching per-shard DBSIZE, and 60/60 data readback. The strict terminal gate still failed because only 3/6 PVCs reached `phase=formed`.

Head `007ad7e459dba8971de96aca38dec5e31f468d6a` selected all pods, then focused review proved a fixed-order deadlock: primaries waited for replica actions that could not run after the first retryable error.

Head `fc9971c3b78185e8a1086a1374a19471bf2d1d3b` removed the ordering dependency. Primaries close only primary duties; replicas validate, self-attach, and close only their own binding.

Head `8e146dd528de694d7b91974e1765cec677c9acb8` added configured-roster validation after a focused RED showed that identical primary views containing the same foreign slotless node could otherwise reach `ADDSLOTSRANGE`.

The second-pass RED on `8e146dd5` modeled three isolated primaries sharing one `CLUSTER MYID`. The old code skipped every `MEET`, reached `ADDSLOTSRANGE`, and produced one recorded slot write. Head `062f281a` requires the designated primary IDs to be unique before reading the coordinator view or performing any cluster mutation. The corrected example returns `restore-membership`, `retry_safe=no`, and records zero writes.

A solo subagent review then found that the generic already-formed fast path could commit and delete local restore metadata without revalidating archived ownership. The pre-fix probe returned `RC=0 META_EXISTS=no RESTORE_VALIDATION=0`. Current head `fffda798` routes every existing `cluster-meta` through the local primary-ownership or restored-replica preparation path before the generic formed fast path.

## Safety contract

The restore job validates a canonical non-root data directory and safe append names before writing. It removes attempt-owned partial payload after extraction or validation failure. Cluster archives atomically persist `.kb-cluster-restore-state` as `phase=prepared`, bound to the exact `cluster-meta` SHA-256.

Before server start, a restored non-first pod must prove:

- exact regular non-symlink metadata, RDB, restore state, offline markers, and multipart AOF files
- `dump.rdb` matches the backup-recorded SHA-256
- multipart AOF is exactly the seeded base RDB, empty incremental AOF, and exact manifest with no extra files
- coexisting offline markers across a crash boundary are exact and bound to this pod and metadata

It then persists an offline prepare marker, deletes only the verified redundant RDB/AOF payload, persists the prepared marker, and starts Valkey. No restore path issues online `FLUSHALL`, `CLUSTER RESET`, or destructive online cleanup.

Before `MEET` or slot assignment, the current head requires:

- the source and target shard counts to match
- every designated primary to answer and return a non-empty, unique engine node ID
- every observed node address to map to exactly one configured shard/pod FQDN
- every primary view to contain every designated primary ID
- archived slot ranges to be valid, non-overlapping, and owned only by the intended restored primary

Configured replicas already attached during convergence are allowed. Foreign members, duplicate configured-address mappings, duplicate designated primary IDs, malformed views, slot overlap, and slot ownership outside the archived ranges fail closed before writes.

## Scope and boundary

This supports sharded restore only when source and target shard counts match. Different shard counts remain a hard failure. Peer resolution is intentionally IPv4-only. Cluster archives created before the `rdb_sha256` metadata field fail closed and must be recreated.

Runtime acceptance is not claimed for head `fffda798a6bef4aa92cc1fc30d8167649e7f032e`. Runtime evidence is N=0 on this exact head. Older CT11 runs are diagnostic evidence only and do not migrate to this head.

Solo subagent-assisted focused review is complete with PASS / 0 remaining blocker. Exact-head CI is 7/7 SUCCESS. This PR is not merge-ready because exact-head runtime remains N=0, and it remains stacked on a parent branch that has not landed in `main`.

## Verification

- pre-fix duplicate-primary-ID contract: 40 examples, 1 failure; recorded `ADDSLOTSRANGE` write
- pre-fix formed-fast-path probe: `RC=0 META_EXISTS=no RESTORE_VALIDATION=0`
- exact-head restore-slot contract on Bash 3.2: 40 examples, 0 failures
- exact-head restore-slot contract on Bash 5.3: 40 examples, 0 failures
- affected restore/manage/start specs on Bash 3.2: 176 examples, 0 failures
- affected restore/manage/start specs on Bash 5.3: 176 examples, 0 failures
- full Valkey suite on Bash 3.2: 274 examples, 0 failures
- full Valkey suite on Bash 5.3: 527 examples, 0 failures
- Bash syntax checks, scoped ShellCheck, and `git diff --check`: PASS
- exact-head GitHub CI: 7/7 SUCCESS
- solo subagent-assisted focused review: PASS / 0 blocker

### DIFF
--- a/addons/valkey/dataprotection/backup.sh
+++ b/addons/valkey/dataprotection/backup.sh
@@ -194,10 +194,21 @@ if [ "${_cluster_enabled}" = "1" ]; then
     echo "ERROR: cluster mode — invalid slot ranges '${_shard_slot_ranges}' (migration markers, overlap, malformed or out-of-domain); refusing backup metadata." >&2
     exit 1
   fi
+  _rdb_sha256=$(sha256sum ./dump.rdb 2>/dev/null | awk '{print $1}')
+  case "${_rdb_sha256}" in
+    ''|*[!0-9a-fA-F]* )
+      echo "ERROR: cluster mode — cannot compute dump.rdb SHA-256 for restore identity." >&2
+      exit 1 ;;
+  esac
+  [ "${#_rdb_sha256}" -eq 64 ] || {
+    echo "ERROR: cluster mode — invalid dump.rdb SHA-256 length." >&2
+    exit 1
+  }
   {
     printf 'source_shards=%s\n' "${_source_shards}"
     printf 'shard_master_id=%s\n' "${_shard_master_id}"
     printf 'shard_slot_ranges=%s\n' "${_shard_slot_ranges}"
+    printf 'rdb_sha256=%s\n' "${_rdb_sha256}"
   } > ./cluster-meta
   backup_files+=("./cluster-meta")
   # cluster-meta is backup metadata, not engine data: it must not remain
--- a/addons/valkey/dataprotection/backup.sh
+++ b/addons/valkey/dataprotection/backup.sh
@@ -135,13 +135,35 @@ fi
 backup_files=("./dump.rdb")
 [ -f "./users.acl" ] && backup_files+=("./users.acl")
 
+validate_cluster_slot_ranges() {
+  awk -v raw="$1" 'BEGIN {
+    if (raw == "") exit 1
+    n = split(raw, parts, ",")
+    for (i = 1; i <= n; i++) {
+      token = parts[i]
+      if (token ~ /^[0-9]+$/) {
+        start = token + 0; end = start
+      } else if (token ~ /^[0-9]+-[0-9]+$/) {
+        split(token, bounds, "-")
+        start = bounds[1] + 0; end = bounds[2] + 0
+      } else {
+        exit 1
+      }
+      if (start < 0 || end > 16383 || start > end) exit 1
+      for (slot = start; slot <= end; slot++) {
+        if (seen[slot]++) exit 1
+      }
+    }
+  }'
+}
+
 # Cluster (sharding) mode: embed the source shard count as engine truth
 # (master lines in CLUSTER NODES) so restore can verify the same-shard-count
 # v1 boundary. Sentinel/standalone targets skip this (cluster_enabled:0).
 _cluster_enabled=$("${connect_base[@]}" INFO cluster 2>/dev/null | grep "^cluster_enabled:" | tr -d "\\r" | cut -d: -f2)
 if [ "${_cluster_enabled}" = "1" ]; then
   _source_shards=$("${connect_base[@]}" CLUSTER NODES 2>/dev/null | tr -d "\r" | awk '$3 ~ /master/' | grep -c . || true)
-  if [ -z "${_source_shards}" ] || [ "${_source_shards}" -lt 1 ]; then
+  if [ -z "${_source_shards}" ] || [ "${_source_shards}" -lt 3 ] || [ "${_source_shards}" -gt 32 ]; then
     echo "ERROR: cluster mode detected but could not count source shards from CLUSTER NODES." >&2
     exit 1
   fi
@@ -166,6 +188,10 @@ if [ "${_cluster_enabled}" = "1" ]; then
   _shard_slot_ranges=$(echo "${_master_line}" | cut -d' ' -f9- | tr ' ' ',')
   if [ -z "${_shard_slot_ranges}" ]; then
     echo "ERROR: cluster mode — this shard's master ${_shard_master_id} owns no slot ranges; refusing to write empty metadata." >&2
+    exit 1
+  fi
+  if ! validate_cluster_slot_ranges "${_shard_slot_ranges}"; then
+    echo "ERROR: cluster mode — invalid slot ranges '${_shard_slot_ranges}' (migration markers, overlap, malformed or out-of-domain); refusing backup metadata." >&2
     exit 1
   fi
   {
--- a/addons/valkey/dataprotection/restore.sh
+++ b/addons/valkey/dataprotection/restore.sh
@@ -72,35 +72,83 @@ seed_multipart_aof_from_rdb() {
   echo "INFO: Seeded multipart AOF manifest from restored dump.rdb."
 }
 
-# Cluster (sharding) restore: NOT SUPPORTED in v1 — fail fast.
-#
-# Same shard count is NOT a sufficient safety condition (review, PR #3044):
-# restore re-forms the cluster (phase B) with an even slot split, but the
-# SOURCE layout may differ (any post-rebalance / scale history). Keys
-# restored onto a shard that does not own their hash slots are silently
-# unreachable, and archive->target-shard positional mapping is itself
-# unverified. Until a slot-aware restore is designed (cluster-meta already
-# records source_shards + shard_slot_ranges for exactly that), a cluster
-# archive is refused outright — refusal beats silent misplacement.
-refuse_cluster_restore() {
-  local meta="${DATA_DIR}/cluster-meta" source_shards
-  [ -f "${meta}" ] || return 0
-  source_shards=$(grep '^source_shards=' "${meta}" | cut -d= -f2)
-  # Remove everything THIS run extracted before exiting: the emptiness
-  # guard proved DATA_DIR held no user data at entry, so all entries here
-  # are our own extraction output. Without this, a retried prepareData
-  # pod trips the '/data is not empty' guard over our leftovers and the
-  # TRUE refusal reason is masked on every retry (live finding, CT11
-  # focused rerun).
+# Cluster archives retain cluster-meta until postProvision has rebuilt the
+# engine membership with the archived slot ownership. prepareData validates
+# the local shard metadata here; cluster-wide overlap/gap checks happen while
+# slots are assigned by valkey-cluster-manage.sh.
+cleanup_restored_payload() {
   find "${DATA_DIR}" -mindepth 1 -maxdepth 1 ! -name ".kb-data-protection" ! -name "lost+found" -exec rm -rf {} +
-  echo "ERROR: this archive is a Valkey CLUSTER (sharding) backup (source_shards=${source_shards})." >&2
-  echo "  Cluster datafile restore is NOT supported in v1: restored slot layouts cannot yet be" >&2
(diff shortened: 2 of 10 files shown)


weicao:
## Independent deep review @ exact head `007ad7e459dba8971de96aca38dec5e31f468d6a` (base `c04f666c`)

Scope: full read of `backup.sh`, `restore.sh`, `valkey-cluster-manage.sh`, `valkey-cluster-server-start.sh`, `backuppolicytemplate.yaml`, `cmpd-valkey-cluster.yaml` at head, plus KB `main` controller source for the declared blocker. Static + source review only; runtime N=0 boundary unchanged.

### Declared blocker — CONFIRMED at controller source, and it is sharper than stated

KB main `pkg/controller/lifecycle/kbagent.go:372-403` (`callActionWithSelector`): pods run **sequentially** and the loop **returns on the first pod error** (`if len(rsp.Error) > 0 { return nil, formatError(...) }`); the `// TODO: impl - back-off to retry - timeout` comment confirms this is known-incomplete.

Combined with this PR's state machine, the failure is not "can starve" but a **deterministic deadlock**: every pod structurally returns rc=1 at least once before global formation — the coordinator defers after the MEET round (`valkey-cluster-manage.sh:719-722`), after slot assignment (`:773-774`), and while observing attach (`:791-800`); a replica returns rc=1 **even after a successful self-attach** (`:639-640`). Under first-error-return, the first deferring pod in list order prevents every pod behind it from running that round; replicas can only attach inside their own invocations; therefore formation is unreachable and `maxRetries: 20 × retryInterval: 5` burns in ~2 min → provision fails every time. The KB-side fix (attempt every selected pod, aggregate errors) is the correct one; an addon-side workaround (rc=0 once the local obligation is positively closed, moving `formed` commit to a later hook) would weaken the metadata-cleanup invariant and needs its own crash analysis — recommend staying on the KB lane.

### Should fix in-PR before acceptance (cheap, local)

1. **MAJOR — no per-call timeout on any `valkey-cli` invocation.** `build_cli`/`build_cluster_cli` (`valkey-cluster-manage.sh:79-101`) wrap nothing; these are called in loops over all shards (`:613`, `:651`, `:663`, `:680`, `:728`) and the action declares no `timeoutSeconds`. One black-holed peer (rescheduled pod, SYN drop) blocks a single PING for minutes → kbagent 60s clamp kills the attempt → all 20 retries burn while the peer was merely restarting. Fix: `timeout "${VALKEY_CLI_TIMEOUT:-5}"` in both builders + `timeoutSeconds: 50` on the action.
2. **MAJOR — `CLUSTER REPLICATE` error replies are swallowed.** `valkey-cli` exits 0 on `(error)` replies; MEET (`:713-716`) and ADDSLOTS (`:768-771`) correctly case-check `'(error)'*|'ERR '*`, but REPLICATE (`:1352-1356`) checks only rc. A persistent `ERR Unknown node`/`ERR To set a master the node must be empty` logs `repaired ... now replicating` and livelocks retries under a success log and a wrong phase label (`restore-replica-converge`). Fix: add the same error-reply case; classify `attach-replicate no`.
3. **MAJOR — backup hash-vs-tar TOCTOU can produce a green backup that can never restore.** `backup.sh:197` hashes `./dump.rdb`, `:221` re-reads it by name into tar. The target is a live secondary with engine-default `save` thresholds (neither config template sets `save`), so an auto-BGSAVE or full-resync rename-replace between hash and tar archives a *different* rdb than the recorded `rdb_sha256`. Backup completes green; every restore fails closed at `restore.sh:179-183`. Fix: copy/hardlink to a staging name, hash the copy, tar the copy — or re-hash after tar and fail the backup on drift.
4. **MAJOR — `LASTSAVE` fallback fails open to a stale restore point.** `backup.sh:77` `|| _lastsave_before=0`: on a transient LASTSAVE failure, any historical lastsave satisfies the freshness gate (`:108`), and a BGSAVE that was merely *scheduled* during an AOF rewrite can result in archiving the pre-existing stale `dump.rdb` with its (matching) stale SHA — silent data loss relative to the claimed backup point. The sibling fallback at `:107` correctly fails closed. Fix: hard-fail the backup if the LASTSAVE baseline can't be read.
5. **MAJOR — restore is not re-entrant after SIGKILL/OOM/eviction.** `restore.sh` has no EXIT/signal trap (grep: none); "removes attempt-owned partial payload" covers only in-script failure branches. A prepareData pod killed mid-`tar -x` (realistic for multi-GB archives) leaves partial payload; the retried job hits the emptiness guard (`:41-45`) and hard-fails until someone manually wipes the PVC. Safe fix using the existing placeholder ordering (`.kb-data-protection` is removed **last**, `:223`): on entry, `placeholder present AND other entries present` ⇒ prior incomplete attempt ⇒ `cleanup_restored_payload` and proceed; keep the hard failure for `no placeholder + non-empty` (live data).

### Minor (fix or note)

- `cluster_nodes_of` (`:127-131`) pipes through `tr`, so its rc is always 0 — several `|| classify` guards are dead, and an unreachable node during replica-shape validation is misclassified as deterministic `restore-replica-shape no` (`:1423-1428`) instead of a transient defer. Fail direction stays closed; taxonomy is wrong.
- Marker/state "atomic" writes (`server-start:260,278`, `manage:507`, `restore.sh:213-215`) rename before syncing the tmp file's data; a power-loss can leave a truncated marker whose mismatch then permanently refuses startup (fail-closed brick, no data loss). Sync the tmp file before `mv`.
- `sort` for coordinator/first-pod election is bare in `manage.sh:359,386` but `LC_ALL=C` in `server-start:210` — same roster could in principle order differently under different locales; make them uniform.
- `source_shards` counts `master,fail` and slotless masters (`backup.sh:165`) — inflated count permanently refuses a legitimate same-shard-count restore. Count only slot-owning healthy masters.
- Missing `getent` in the image → classified `restore-dns yes` (transient) though deterministic (`manage:414-417`); `host_resolves` (`:1731-1734`) handles it deliberately — mirror that.
- Archive-existence probe `datasafed list | grep -qF` (`restore.sh:56`) is substring-match + hides backend errors as "not found"; fail-closed today, but simplest is to rely on `datasafed pull`'s own failure.

### Verified clean (so you don't re-check)

Shard-count mismatch is a genuine hard failure non-bypassable by empty values (validated `''|*[!0-9]*` + range check before compare, `manage:536-543,607-624`); SHA binding order correct (verify before use/seed, `restore.sh:179-194`); two-stage offline-prepare marker protocol is crash-safe in all orderings I could construct (`server-start:214-284` — commit-before-delete, idempotent re-entry); `phase=formed` committed before metadata removal with clean re-entry on every partial-crash permutation (`manage:477-518`); IPv4 resolution fails closed, never falls back to hostname, validates dotted-quad before MEET (`manage:413-430`); coordinator restore mode is genuinely observation-only (`manage:1325-1334`); no online FLUSHALL/CLUSTER RESET anywhere; bash 3.2 compatibility confirmed (no 4+ constructs); DP env contract uses only DP-injected vars + ActionSet `DATA_DIR` (no component-only-env bug); XP classes 1–8 walked — class 1 hits are exactly items 4 (LASTSAVE) and the `cluster_nodes_of` taxonomy; class 4 (sentinel values) clean; class 7 (terminating-vs-absent) clean via the placeholder ordering.

### Done well

The PVC-local data-safety layer is strong: symmetric strict slot-range validation on producer and consumer with overlap/migration-marker rejection; positive-observation gating on every rc=0 (state=ok + 16384 + exact id-set equality, never absence-of-error); the two-stage offline prepare protocol; and honest evidence discipline in the PR body (first-red packets per head, explicit N=0, self-declared controller dependency).

**Verdict:** direction and design quality are high; no data-loss path found. Blocked exactly where the author says, now confirmed at `kbagent.go` source level with a sharper "deterministic deadlock" characterization. Items 1–5 are cheap, local, and should land in this PR before any runtime attempt; the controller `All`-aggregation fix is the gate for acceptance. Static review only — nothing here claims runtime PASS.

PR #3172, opened 2026-07-13
fix(qdrant): harden rollout provisioning and memberLeave reentry

shanshanying:
- fix https://github.com/apecloud/kubeblocks/issues/10528
## Summary
- derive Qdrant bootstrap/control endpoints from runtime component identity instead of static pod lists
- harden memberLeave to be replay-safe, bounded, and to wait for shard drain before peer removal
- add focused ShellSpec coverage for setup and memberLeave replay/idempotency behavior

## Validation
- shellspec addons/qdrant/scripts-ut-spec/qdrant_member_leave_spec.sh
- shellspec addons/qdrant/scripts-ut-spec/qdrant_setup_spec.sh
- shellcheck --severity=error on changed Qdrant scripts/specs
- helm lint addons/qdrant
- helm lint addons-cluster/qdrant
- live kind-kbv12 memberLeave reentry validation: first run failed while remaining=1/transfers=1, retry completed drain/removal, replay reported already-left, final count=8004


leon-ape (commented):
Finding: `qdrant_start_mode` treats one failed 3-second ComponentService probe as proof that an empty ordinal-0 is the initial node. A recreated ordinal-0 with no local `raft_state.json` will self-bootstrap a second Raft cluster whenever all surviving endpoints are temporarily unready or unreachable.

Finding: `qdrant_current_pod_fqdn` and `qdrant_bootstrap_service_host` reconstruct KubeBlocks Service DNS names from controller naming conventions even though the target `apps.kubeblocks.io/v1` API exposes the resolved ComponentService and headless hosts through `serviceVarRef`. The lifecycle path therefore depends on manually duplicated Service naming and DNS behavior.

Finding: Peer selection uses `contains($name)` against peer URIs. When the leaving pod is ordinal `-1` and ordinal `-10` exists, both URIs match, so `leave_peer_id` contains two IDs and memberLeave fails before draining or removing the intended peer.

Finding: Replacing Qdrant `1.15.4` with `1.15.5` removes a previously published serviceVersion from the rendered ComponentVersion. Existing Clusters pinned to `1.15.4` no longer have a matching release after the addon definition is updated.

Finding: This PR exposes Qdrant `1.15.5` and `1.18.2`, but the validation only reports memberLeave reentry. It does not include current-head provisioning and service-connectivity evidence for either newly exposed version.

leon-ape (commented):
Finding: qdrant_start_mode still treats failure to reach the ComponentService for ten probes as proof that an empty ordinal-0 is the initial node. A recreated ordinal-0 with no local raft_state.json can bootstrap a second cluster during a transient service or network outage.\n\nFinding: values.yaml publishes Qdrant 1.15.5 and 1.18.2, but both required release-registry checks fail because apecloud/qdrant:v1.15.5 and apecloud/qdrant:v1.18.2 do not exist in the Docker Hub and Aliyun release registries.\n\nFinding: Current-head validation covers memberLeave reentry only. It does not provide provisioning and service-connectivity evidence for the newly exposed Qdrant 1.15.5 and 1.18.2 releases.

weicao:
Qdrant owner triage on current head `969224305858fdf71f0ee6bc5204240fc1b3a1cc`: the three current-head review findings are confirmed and remain merge blockers.

1. `qdrant_start_mode` still returns `bootstrap` for an empty ordinal-0 after the bounded ComponentService probes are exhausted. The unit test at `qdrant_setup_spec.sh` explicitly expects this. That does not distinguish first-cluster creation from a recreated ordinal-0 during a temporary service/network outage, so it can start a second Raft cluster. This path needs a positive bootstrap authority/identity signal; probe exhaustion alone must not authorize bootstrap.
2. The exact-head release-chart checks for both Docker Hub and Aliyun are red. `values.yaml` exposes `1.15.5` and `1.18.2`, but the required `apecloud/qdrant:v1.15.5` and `apecloud/qdrant:v1.18.2` release images are unavailable in those registries. Please remove or replace those entries until the release images exist.
3. Current-head evidence covers script/unit behavior only. Before re-review, please attach live provisioning and service-connectivity evidence for every newly exposed serviceVersion that remains in the PR.

@shanshanying please update this owner branch for these three items. I am not treating the current CI/unit results as runtime PASS.

leon-ape (commented):
Finding: qdrant_start_mode still authorizes bootstrap for an empty ordinal-0 solely after the ComponentService probes are exhausted. A recreated ordinal-0 can therefore start a second Raft cluster during a temporary service or network outage.\n\nFinding: memberLeave.timeoutSeconds is 30, but the script can spend 20 seconds in its drain window after multiple 5-second control calls and then up to another 15 seconds in peer removal. KubeBlocks can terminate the action before the script reaches its own bounded result.\n\nFinding: The Docker Hub and Aliyun release checks still fail because the published 1.15.5 and 1.18.2 Qdrant images are absent from both required release registries.\n\nFinding: Current-head evidence still does not include actual provisioning and service-connectivity validation for Qdrant 1.15.5 and 1.18.2.

weicao:
Current-head owner follow-up on `969224305858fdf71f0ee6bc5204240fc1b3a1cc`: the new `memberLeave.timeoutSeconds` finding is confirmed and is a fourth merge blocker.

`addons/qdrant/templates/cmpd.yaml` gives the lifecycle action 30 seconds, but the script has no single outer deadline aligned to that limit. Before the bounded 20-second drain loop, collection/control queries and shard-move requests can each consume 5 seconds; after the loop, peer removal can consume a 5-second GET, 5-second DELETE, and another 5-second GET. KubeBlocks can therefore terminate the action before the script reaches its own success/failure result.

Please make the action budget and script budget one contract: either raise the lifecycle timeout with a documented safety margin, or pass one outer deadline through every query/move/drain/remove step and fail before KubeBlocks' deadline. Add a deterministic slow-control fixture proving the script exits with its own classified result inside the lifecycle timeout. The existing bootstrap-authority, missing release images, and current-head provisioning/service-connectivity blockers also remain open.

This is current-head static review only; it is not runtime evidence.

weicao:
Addressed the ordinal-0 bootstrap review on current head `fc98b1fdfe34826ad3ba66eb28f6e9917c5d3160`.

The service probe is no longer sufficient authorization to bootstrap. A fresh ordinal-0 must now atomically persist a one-shot bootstrap claim on its data PVC, bound to the Pod UID injected through the Downward API. Once that claim exists, retries from the same Pod and recreated ordinal-0 Pods both take the join/wait path during an outage. Existing service or durable Raft state records an `existing-cluster` marker. Missing UID, non-empty unmarked storage, concurrent claims, and marker write failures fail closed.

Focused verification:
- Qdrant ShellSpec suite: `24 examples, 0 failures`
- `shellcheck --severity=error`: pass
- Bash syntax checks: pass
- `helm lint addons/qdrant`: `1 chart(s) linted, 0 failed`
- `helm template`: pass; rendered `CURRENT_POD_UID` uses `metadata.uid`
- `git diff --check`: pass

Design-contract review on this head:
- Pre-flight: all three changed files serve only this bootstrap-authorization fix; production command remains `/bin/bash`, and supported Qdrant images use Debian-based runtimes with the required core utilities.
- Class 1: no fallback from exhausted discovery to unclaimed bootstrap.
- Class 2: Pod UID is explicitly injected and non-empty; all marker writes are checked.
- Class 3: the atomic one-shot claim closes same-Pod and recreated-Pod retry discontinuities.
- Class 4/4a: marker states (`bootstrap-attempt:<uid>` and `existing-cluster`) are distinct and caller output is explicit.
- Classes 5-7: no cleanup, NotFound, or release-state path is introduced.
- Class 8: no ambiguous boolean precedence.
- Caller re-read: an existing-cluster marker write failure returns no start mode, so `qdrant_setup_main` exits through the unknown-mode error instead of starting Qdrant.

Scope boundary: this closes only the ordinal-0 split-bootstrap review. The existing memberLeave timeout, registry-image, and live provisioning/connectivity evidence items remain open; this is not a release-readiness claim.

weicao:
CI is now terminal on `fc98b1fdfe34826ad3ba66eb28f6e9917c5d3160`: all 8 checks passed, including full ShellSpec (`5m46s`). The scope boundary above is unchanged; the other review items remain open.

leon-ape (commented):
Finding: memberLeave.timeoutSeconds remains 30 seconds, while preflight queries, per-collection queries and move requests, the 20-second drain loop, and peer removal do not share an outer deadline. KubeBlocks can terminate the action before the script reaches its own result.

Finding: Current head exposes Qdrant 1.15.5 and 1.18.2 without live provision and service-connectivity evidence for those service versions.

weicao:
Addressed the `memberLeave.timeoutSeconds=30` review finding on current head `955bc9befdcb4e6281ca8f9cb427184bb30d85d7` (parent `fc98b1fdfe34826ad3ba66eb28f6e9917c5d3160`).

Contract anchors:
- `docs/design/addon-kbagent-action-timeout-cap-guide.md`: positive action timeouts are capped at 60s; fit actual work in about 50s and keep a safety buffer.
- `docs/addon-api/05b-lifecycle-action-fields.md`: long actions must declare a bounded timeout/retry contract.
- `docs/addon-api/05c-lifecycle-action-contracts.md`: `memberLeave` must remain replay-safe and diagnosable after partial completion.

Root fix:
- raise the lifecycle action timeout from 30s to 60s;
- start one immutable 50s script deadline before cluster preflight, leaving a 10s kbagent safety buffer;
- derive the drain phase from the same deadline and reserve 10s for peer-removal finalization;
- route every Qdrant API call through a deadline-aware curl wrapper that clips `--max-time` to remaining budget;
- fail closed with rc 124 and an explicit diagnostic before issuing a call when the budget is exhausted;
- keep the existing replay-safe move/peer-removal behavior and bounded 20s drain observation.

RED before fix: 14 examples, 4 failures (deadline absent, curl not budget-clipped, exhaustion not fail-closed, deadline reset after preflight).

Verification on this head:
- full Qdrant ShellSpec: 30 examples, 0 failures;
- focused member-leave ShellSpec: 15 examples, 0 failures;
- error-level ShellCheck: pass;
- Bash syntax: pass;
- `helm lint addons/qdrant`: 1 chart, 0 failures;
- Helm render: pass, including `timeoutSeconds: 60` and the shared-deadline helper;
- `git diff --check`: pass.

XP design-contract walk:
1. Silent fallback: none; exhausted budget returns rc 124.
2. Non-empty unenforced config: action budget is validated as integer 1..50; curl/drain/finalization budgets are enforced.
3. Commit-state discontinuity: one action deadline is created before preflight and is never reset; phase deadline is derived from it.
4. Sentinel collision: rc 124 has one explicit meaning and propagates fail-closed under `set -e`.
5. Cleanup ownership: N/A; this change does not add cleanup.
6. NotFound semantics: unchanged.
7. Release claims: none.
8. Ambiguous booleans: none added.

The separate live provisioning/service-connectivity evidence requested for Qdrant 1.15.5 and 1.18.2 remains open. This commit closes only the timeout/deadline code-review finding; it is not a release-ready claim.

weicao:
CI terminal on `955bc9befdcb4e6281ca8f9cb427184bb30d85d7`: 8/8 required checks passed. The timeout/deadline finding is code-closed on the current head. The separately requested live provision/connectivity evidence for Qdrant 1.15.5 and 1.18.2 remains open, so this is not a release-ready conclusion.

addons/qdrant/scripts/qdrant-member-leave.sh
```diff
@@ -1,201 +1,626 @@
...
+    echo "INFO: shard ${shard_id} in ${col_name} is already moving from peer ${leave_peer_id}"
+    return 0
+  fi
+
+  if ! qdrant_shard_on_leaving_peer "$col_cluster_info" "$shard_id" "$leave_peer_id"; then
+    echo "INFO: shard ${shard_id} in ${col_name} is already off peer ${leave_peer_id}"
+    return 0
+  fi
+
+  echo "INFO: move shard ${shard_id} in ${col_name} from peer ${leave_peer_id} to peer ${target_peer_id}"
```
leon-ape: Finding: `target_peer_id` is selected once for the entire leave operation without checking whether it already holds each shard. For replicated collections, `move_shard` can target an existing replica; Qdrant then removes the source without creating a replacement replica, reducing the shard replica count during scale-in.

weicao:
Addressed the current-head shard target review on `ae32d2c77eb0f6743b2a94a5ca5bace155beb63e` (parent `955bc9befdcb4e6281ca8f9cb427184bb30d85d7`).

Problem and API contract:
- The previous script chose one cluster-wide `target_peer_id` and reused it for every collection/shard.
- Qdrant's production `GET /collections/{collection}/cluster` response provides the serving `result.peer_id`, `local_shards`, `remote_shards`, and `shard_transfers`; those fields are now the ownership source for each shard.

Root fix:
- select a target independently for each collection/shard, after the existing replay checks;
- derive occupied peers from the local owner, every remote owner, and matching in-flight transfer destinations;
- exclude the leaving peer and all occupied peers;
- prefer the Raft leader only when it is eligible, otherwise choose another eligible survivor;
- if no unused surviving peer exists, fail closed before `move_shard` and before peer removal rather than reduce replica count.

RED before fix: 21 examples, 5 failures because shard-aware selection did not exist.

Verification on this head:
- focused memberLeave ShellSpec: 21 examples, 0 failures;
- full Qdrant ShellSpec: 35 examples, 0 failures;
- Bash syntax and error-level ShellCheck: pass;
- `helm lint addons/qdrant`: 1 chart, 0 failures;
- Helm render and `git diff --check`: pass.

Production-shaped fixtures cover: eligible leader, leader holding the shard remotely, leader holding it locally, in-flight transfer destination, every survivor already holding the shard, malformed local ownership without `peer_id`, request payload choosing the unused peer, and no request when preserving replica count is impossible.

XP design-contract walk:
1. Silent fallback: none; no safe target is an explicit failure.
2. Unenforced input: ownership comes from the production collection-cluster response and is enforced before submit.
3. Commit-state discontinuity: target selection uses the same per-shard snapshot checked for replay; no cluster-wide target is carried across shards.
4. Sentinel collision: empty target has one meaning and the caller fails closed.
5. Cleanup ownership: N/A; no cleanup added.
6. NotFound semantics: unchanged.
7. Release claims: none.
8. Ambiguous booleans: none added.

This closes only the source-level replica-preservation finding. It is not live scale-in evidence and does not close the separate Qdrant 1.15.5/1.18.2 provisioning and service-connectivity evidence requirement.

weicao:
CI terminal on `ae32d2c77eb0f6743b2a94a5ca5bace155beb63e`: 8/8 checks passed, including ShellSpec in 5m30s. The per-shard safe-target source finding is code-closed on the current head. Live scale-in evidence and the separate Qdrant 1.15.5/1.18.2 provision/connectivity evidence remain open, so this is not a release-ready conclusion.

addons/qdrant/scripts/qdrant-member-leave.sh
```diff
@@ -1,201 +1,662 @@
...
-  echo "member ${KB_LEAVE_MEMBER_POD_NAME} is not in the cluster (not found in peer URIs)"
-  exit 0
-fi
+  echo "$collection_cluster_info" | "$jq_bin" -e \
+    --argjson shard "$shard_id" \
+    --argjson peer "$peer_id" \
+    '((.result.remote_shards // [])
+      | any(.peer_id == $peer and .shard_id == $shard))
+     or
+     ((.result.local_shards // [])
```
leon-ape: Finding: Qdrant `local_shards` entries do not carry `peer_id`; their owner is `result.peer_id`. Falling back to the requested leaving peer makes every local shard in a surviving control response appear to remain on the leaving peer. If a move succeeds but its HTTP response is lost and the transfer completes before recheck, this replay path reports failure instead of recognizing that the shard is already off the peer.

leon-ape (commented):
Current head still exposes Qdrant 1.15.5 and 1.18.2 without actual provisioning and service-connectivity evidence for either service version.

weicao:
Current-head runtime evidence is now available for both exposed service versions on exact addon commit `2456c3c127f6770043becaa17918ad7022962487`.

Test identity:
- KubeBlocks `1.2.0-alpha.2`, live controller imageID `sha256:d612066a11f7924aaf220d4e5bbf1dccc5f2f6407e8813a75d1f0d9fcc439989`
- syncer source `2489b4b7e52d4bdbec128e4758c749ae619cc5bd`, live imageID `sha256:6580fd93a4d909a41eeebd74c93577e415a5c14bbb033dfdbadf38e17eb36972`
- formal run `pr3172-r2-0944z`, Job UID `d98d0483-8174-43d5-b03e-5cf821578217`, Pod UID `23640c9e-afd8-4bd8-ae8d-681f1fe60544`, restart 0, exit 0, `2026-07-24T09:47:11Z` to `2026-07-24T09:55:00Z`

Results:
- Qdrant `1.15.5`: fresh 3-node provision and Service-DNS connectivity succeeded; before and after scale-in the retained JSON response was non-empty with `.status == "ok"`. The 3->2 memberLeave OpsRequest succeeded; removed peer was absent afterward; collection was green with 40 points and 3 shards x 2 unique Active replicas. Ordinary cleanup ended NotFound/residue 0 without force delete or finalizer edits.
- Qdrant `1.18.2`: the same fresh provision, before/after Service-DNS connectivity, 3->2 memberLeave, green/40/readback, 3 shards x 2 Active replicas, and ordinary residue-0 cleanup gates succeeded.

The corrected accepted run is scoped product N=2, one sample per service version. An earlier runner result was excluded from N because its connectivity payloads were empty; the accepted run executes bounded curl from a guest Qdrant Pod and fails explicitly on curl/jq errors.

Evidence archive SHA-256: `3436eea406573f716960cae2242e6acfd19fefcb4698f2ae6ae19c0ae14c5e8e`. I independently re-downloaded it, matched the outer SHA, verified all 123 root-manifest payload hashes, and checked the two non-empty connectivity responses, OpsRequest terminal phases, removed-peer absence, replica ledger, 40-point readback, and cleanup receipts. This closes the requested provisioning + service-connectivity evidence for the two service versions; it is not a broader release-readiness claim.

leon-ape (approved):
LGTM

PR #3173, opened 2026-07-13
fix(postgresql): rebuild replica after rewind failure

### LINKED ISSUE #3171: PostgreSQL replica can remain stuck after pg_rewind failure
## Problem

The PostgreSQL addon enables Patroni `use_pg_rewind`, but the generated local Patroni configuration does not enable `remove_data_directory_on_rewind_failure`.

When `pg_rewind` fails, Patroni's default behavior is to keep the existing data directory and try to follow the new leader. A former primary with a divergent timeline can then remain in startup forever instead of being recreated as a replica.

## Reproduction

Candidate under test:

- addon base: `5003f0515b398197119216c5eb5d0fd5b613f13d`
- PostgreSQL: 18.4, two replicas
- test: `kubeblocks-tests` replication T08 Restart

Observed sequence:

1. pod0 was the old primary; pod1 became the timeline-2 primary.
2. pod0 ran `pg_rewind` against pod1.
3. `pg_rewind` exited 1 because local WAL segment `000000010000000000000004` was missing and the divergence record could not be read.
4. Patroni started pod0 as a secondary using the same data directory.
5. pod0 had no walreceiver and kept waiting for timeline-2 segment 4 while the primary was still on segment 3.
6. pod0 never became Ready and the Restart OpsRequest timed out after 1200 seconds.

The live local Patroni configuration contained `use_pg_rewind: true` and valid `primary_conninfo`, but neither `remove_data_directory_on_rewind_failure` nor `remove_data_directory_on_diverged_timelines`.

## Expected behavior

If rewind fails, Patroni should discard the unusable former-primary data directory and recreate the replica through the existing replica creation path.

## Test plan

- Deterministic unit RED: generate final Patroni YAML with `use_pg_rewind=true` and assert the local `postgresql` mapping enables `remove_data_directory_on_rewind_failure`.
- Focused runtime: rerun the preserved T08 failure scenario with the patched addon and verify replica reinitialization, role convergence, data consistency, and Restart completion.
- Fresh runtime: rerun replication from T01, then resume the full acceptance suite only after the focused case passes.

## Scope

Addon-only Patroni policy fix. No KubeBlocks controller, vcluster, restart-order, or unrelated Patroni policy changes.

weicao:
## Problem

Closes #3171.

The PostgreSQL addon enables Patroni `use_pg_rewind`, but its generated local Patroni configuration leaves `remove_data_directory_on_rewind_failure` at the default `false`.

During the PR #3112 runtime acceptance, replication T08 Restart reproduced this failure:

1. the former primary ran `pg_rewind` against the new primary;
2. rewind failed because a required local WAL segment was missing;
3. Patroni reused the divergent PGDATA and kept trying to follow;
4. the replica stayed in startup and the Restart OpsRequest timed out.

## Fix

Set `postgresql.remove_data_directory_on_rewind_failure=true` in the final local Patroni configuration generated by `generate_patroni_yaml.py`.

This is intentionally limited to the observed rewind-exit failure path. It does not enable the separate diverged-timeline fallback and does not change restart ordering, KubeBlocks controller behavior, vcluster, or templates under `bootstrap.dcs`.

## TDD

- RED on base `5003f0515`: focused ShellSpec 4 examples / 1 failure with the exact missing fallback key.
- GREEN on this head: focused 4/0 under bash 3.2 and bash 5.3.
- Related PostgreSQL specs excluding the pre-existing macOS PITR portability case pass under bash 3.2 and bash 5.3.
- The excluded PITR failure is identical on clean base and candidate (`ls -lI` / BSD basename), so it is not introduced here.
- `python3 -m py_compile`, Helm dependency build, lint, render, rendered scripts ConfigMap check, and `git diff --check` pass.

## Review

Agent peer review of exact `09a37377e2f25db886e7b884dba2072fd60320c9`: APPROVE, 0 blocking findings. The reviewer independently verified that the key lands in the final top-level local `postgresql` mapping and is absent from `bootstrap.dcs`.

## Runtime boundary

The PR remains draft. Runtime validation is still required on the exact candidate:

1. focused fresh T08 Restart reproducer;
2. verify rewind failure triggers replica reinitialization through the existing replica method;
3. verify roles, readiness, replication, data consistency, and OpsRequest convergence;
4. then run a fresh replication suite from T01 before resuming full acceptance.

### DIFF
--- a/addons/postgresql/scripts-ut-spec/generate_patroni_yaml_spec.sh
+++ b/addons/postgresql/scripts-ut-spec/generate_patroni_yaml_spec.sh
@@ -21,8 +21,7 @@ restore_data:
   command: bash /home/postgres/pgdata/kb_restore/kb_restore.sh --replica
 EOF
     cat > "${conf_dir}/patroni.yaml" <<'EOF'
-postgresql:
-  use_pg_rewind: true
+ttl: 30
 EOF
     cat > "${conf_dir}/kb_pitr.conf" <<'EOF'
 method: kb_restore_from_time
@@ -38,7 +37,8 @@ EOF
     export SPILO_CONFIGURATION='bootstrap:
   initdb:
   - auth-host: md5
-postgresql: {}'
+postgresql:
+  use_pg_rewind: true'
   }
 
   cleanup() {
@@ -54,6 +54,20 @@ import yaml
 with open(sys.argv[1], 'r') as stream:
     config = yaml.safe_load(stream)
 sys.exit(0 if 'recovery_conf' not in config['bootstrap']['kb_restore'] else 1)
+PY
+    printf "%s" "$?"
+  }
+
+  rewind_failure_reinitializes_replica() {
+    python3 - "${out}" <<'PY'
+import sys
+import yaml
+
+with open(sys.argv[1], 'r') as stream:
+    postgresql = yaml.safe_load(stream)['postgresql']
+
+assert postgresql['use_pg_rewind'] is True
+assert postgresql['remove_data_directory_on_rewind_failure'] is True
 PY
     printf "%s" "$?"
   }
@@ -81,6 +95,12 @@ PY
     The status should eq 0
     The contents of file "${out}" should not include "method: kb_restore"
     The contents of file "${out}" should not include "create_replica_methods:"
+  End
+
+  It "reinitializes a replica when pg_rewind fails"
+    When run python3 ../scripts/generate_patroni_yaml.py "${out}"
+    The status should eq 0
+    The result of function rewind_failure_reinitializes_replica should eq 0
   End
 
   It "keeps the recovery configuration for point-in-time restore"
--- a/addons/postgresql/scripts-ut-spec/generate_patroni_yaml_spec.sh
+++ b/addons/postgresql/scripts-ut-spec/generate_patroni_yaml_spec.sh
@@ -38,7 +38,10 @@ EOF
   initdb:
   - auth-host: md5
 postgresql:
-  use_pg_rewind: true'
+  use_pg_rewind: true
+  remove_data_directory_on_rewind_failure: false
+  parameters:
+    max_connections: 123'
   }
 
   cleanup() {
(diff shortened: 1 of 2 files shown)


weicao:
## Focused runtime evidence: real `pg_rewind` failure fallback

Exact product identity:

- PR head: `37a4b56709538ce8aaaeaed1f1475b4c26dd3e5e`
- integration candidate: `1beb0f0c3a6dcd582065b8971c57661f6daaa1d4`
- chart/app: `1.2.0-alpha.0` / `18.4.0`
- chart SHA256: `3210b9aba86f748c34010f79875d15fe1021ab24415abda382a1bf1f58084405`
- rendered manifest SHA256: `0a152b3f3187739ae7f6fb7f3fcb797071b6f845e1cb57fbb9a80d4d70646284`
- runner script SHA256: `c82295507ae2e24848268c5c7a297904524a64d11ea5b47421a70381ec162fc0`
- runner image ID: `sha256:54f3f826f70e3b84fbd09363300700aa060e3a73efafab9df92f05845b01d3ac`

The first four attempts were runner injection/setup failures before the target process path. Each had bounded `pg_rewind` invocation count `0`, so they are noncounting and are not product failures.

The fifth fresh attempt closed the direct process chain:

1. `pg_controldata` reached exact `shut down` before fault injection.
2. Patroni logged real-source `running pg_rewind`.
3. `pg_rewind` exited 1 after proving timeline divergence/common checkpoint and hitting the injected target-relation `Permission denied`.
4. Patroni logged `remove_data_directory_on_rewind_failure is set. removing...`.
5. Patroni rebuilt the failed old primary with `basebackup_fast_xlog`; it rejoined as replica.

Isolation/data checks showed the rebuilt old-primary sentinel absent, while the promoted primary retained its sentinel and primary role. Service SQL, Cluster/Component Running, InstanceSet 2/2, both Pods 5/5 restart 0, PVCs Bound, controller/syncer exact identities, and runner exit 0/0 all closed.

Evidence manifest: `task51-r5-terminal-clean.tar.gz`, SHA256 `2895c5afe5c4ffa1a7108138bd7e816443ce409a65b1f23afbdcc8edc113d6d9`; tar paths safe, 20/20 internal checksums PASS, credential scan PASS. The artifact is retained in the team evidence system under PostgreSQL task #51. Normal cleanup completed with guest/host scoped residue 0 and no force deletion or finalizer edit.

**Verdict boundary:** focused real-path CLEAN N=1 for this exact PR/integration identity. This is not full-suite, soak, broad compatibility, or release-ready evidence.

weicao:

PR #3113, opened 2026-07-08
feat(rabbitmq): add physical backup and restore support

weicao:
## Problem

RabbitMQ currently has no addon-side Backup/Restore product surface: no `BackupPolicyTemplate`, no `ActionSet`, no backup/restore scripts, and no user examples. Per the addon-api data protection contract, backup support requires the template plus executable backup/restore behavior; a runtime B/R test cannot be launched honestly before that surface exists.

RabbitMQ physical restore also has a hard identity boundary: RabbitMQ documents that disk restore for modern versions requires restoring to a node with the same node name; node renaming is not supported for quorum queues or streams. This PR therefore keeps the first product slice intentionally narrow.

## Solution

- Add `rabbitmq-backup-policy-template` with one `physical` full backup method.
- Select all RabbitMQ pods (`strategy: All`) and mount each pod's `data` volume.
- Add `rabbitmq-physical-br` ActionSet with:
  - backup: stop target app, archive the data directory through `datasafed`, restart app, fail closed via `DP_BACKUP_INFO_FILE.exit`.
  - restore prepareData: refuse non-empty data dirs, pull archive, restore ownership.
  - postReady: reconcile the restored KubeBlocks root account password/tags/permissions.
- Add BackupRepo/Backup/Restore examples and document the same-identity restore boundary.
- Add ShellSpec contract coverage for BPT/ActionSet wiring and script safety checks.

## Boundary

- Runtime validation has not run yet; this PR is addon-side implementation plus static/render/script checks.
- Claimed scope: full physical same-identity restore only: same namespace, cluster name, component name, replica count, and serviceVersion.
- Not claimed: PITR, logical export/import, cross-topology restore, cross-version restore, `scaleOut.fromBackup`, or `RebuildInstance`.

## Verification

- `bash -n addons/rabbitmq/dataprotection/backup.sh addons/rabbitmq/dataprotection/restore.sh addons/rabbitmq/dataprotection/post-restore.sh`
- `shellspec --load-path ./shellspec --default-path addons/rabbitmq/scripts-ut-spec --shell bash` -> 8 examples, 0 failures
- `helm dependency build addons/rabbitmq`
- `helm lint addons/rabbitmq`
- `helm template rabbitmq-test addons/rabbitmq --namespace demo`
- `git diff --check`

## Runtime validation needed after review

Run a same-identity physical B/R validation on an approved vcluster:
1. create 3-replica RabbitMQ, publish durable messages, create user/vhost/policy, verify management login;
[...]

Issue #81, opened 2023-12-13
[Improvement] set constraint on ob clusters


shanshanying:
Users cannot set arbitary values on zones, replicas, and ob clusters
where
- \#zones <= \#replicas
- \#obclusters <=2
- tenant resources < replicas resourcs

PR #3183, opened 2026-07-13
fix(rabbitmq): derive restore archive stem from target path

weicao:
## Problem

Volume-populator restore supplies `DP_TARGET_RELATIVE_PATH` as `<target-name>/<target-pod-name>`. `DP_BACKUP_BASE_PATH` is already scoped to that relative path. Treating the full relative path as the archive stem makes restore look for `<target-name>/<target-pod-name>.tar.zst` inside the per-target base path, while backup writes `<target-pod-name>.tar.zst`.

This addresses the current human review blocker on #3113 without changing the backup artifact layout.

## Fix

- Prefer `DP_TARGET_POD_NAME` when the backup/restore job provides it.
- Otherwise require exactly `<target-name>/<target-pod-name>` and derive only the final pod-name segment.
- Fail closed when the relative path has no target or pod segment, or has extra path segments.
- Keep `DATASAFED_BACKEND_BASE_PATH` equal to the injected `DP_BACKUP_BASE_PATH`.
- Update the contract test to use the actual `<target>/<pod>` shape and assert both base path and archive key.

## Validation

- RED before implementation: focused ShellSpec looked for `rabbitmq/rmq-br-5400-rabbitmq-2.tar.zst`.
- Additional RED: `rabbitmq/` was incorrectly accepted as archive stem `rabbitmq` before the fail-closed guard.
- GREEN after implementation: focused RabbitMQ backup/restore ShellSpec, 15 examples / 0 failures.
- `shellcheck addons/rabbitmq/dataprotection/restore.sh`
- `bash -n addons/rabbitmq/dataprotection/restore.sh`
- `git diff --check`

Runtime B/R evidence is intentionally not claimed by this static fix; the exact repinned candidate still needs same-identity backup/restore validation.

### DIFF
--- a/addons/rabbitmq/dataprotection/restore.sh
+++ b/addons/rabbitmq/dataprotection/restore.sh
@@ -3,9 +3,26 @@ set -Eeuo pipefail
 
 DATA_DIR="${DATA_DIR:-/var/lib/rabbitmq}"
 # Backup jobs inject DP_TARGET_POD_NAME. Volume-populator AsDataSource prepareData
-# may only inject DP_TARGET_RELATIVE_PATH (= pod identity / archive stem). Prefer
-# POD_NAME when present; otherwise accept RELATIVE_PATH so KB12 dataSource restore works.
-TARGET_POD_NAME="${DP_TARGET_POD_NAME:-${DP_TARGET_RELATIVE_PATH:?DP_TARGET_POD_NAME or DP_TARGET_RELATIVE_PATH is required}}"
+# may only inject DP_TARGET_RELATIVE_PATH (= target name / target pod name).
+# DP_BACKUP_BASE_PATH is already scoped to that relative path, so only its final
+# pod-name segment belongs in the archive key.
+if [ -n "${DP_TARGET_POD_NAME:-}" ]; then
+  TARGET_POD_NAME="${DP_TARGET_POD_NAME}"
+else
+  TARGET_RELATIVE_PATH="${DP_TARGET_RELATIVE_PATH:?DP_TARGET_POD_NAME or DP_TARGET_RELATIVE_PATH is required}"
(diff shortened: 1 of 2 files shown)


weicao:
Runtime blocker on the integrated exact candidate `f48116a58a7482a96445ae63f4c397ad5f72e53d` (single full run, same topology; three PVCs in that run).

Run: `combined-full-f48116a5-fce8f608-20260713T203014Z`
Result: `PASS=396 FAIL=1 SKIP=12`; sole FAIL was B/R T05.

Direct restore Job evidence for all three target PVCs:

- KB 1.2 volume-populator injected a single-segment value, e.g. `DP_TARGET_RELATIVE_PATH=rmq-br-1974-rabbitmq-2`.
- `DP_BACKUP_BASE_PATH` was already fully scoped to `.../rmq-br-1974-rabbitmq-2`.
- the restore container exited 1 before `datasafed pull` with `ERROR: DP_TARGET_RELATIVE_PATH must be <target-name>/<target-pod-name>`.
- each Job reached `BackoffLimitExceeded`; each Restore kept phase `AsDataSource` with condition reason `Failed`.

This disproves the PR's two-segment runtime assumption for this exact KB 1.2 path. The parent implementation at `0c027e5d` accepted the single-segment `DP_TARGET_RELATIVE_PATH` directly as the archive stem; that is the smallest correction for the observed contract. Please do not reuse `52ff725f` or integrated `f48116a5` as a runnable B/R candidate without a corrective exact SHA.

Full artifact SHA256: `bd83529ae8d00af3a6bc6544e687ea1abd96d539d52f11697cf490327f835a41` (task #81 thread attachment). A separate tests PR, apecloud/kubeblocks-tests#772, only fixes fail-fast detection and does not change the addon verdict.

PR #3193, opened 2026-07-13
fix(rabbitmq): restore single-segment volume target paths

weicao:
## Problem

The integrated candidate `f48116a5` includes #3183's assumption that KB 1.2 volume-populator injects `DP_TARGET_RELATIVE_PATH=<target-name>/<target-pod-name>`.

In task #81 full runtime, all three prepareData Jobs instead received a single pod-identity segment, for example:

```text
DP_TARGET_RELATIVE_PATH=rmq-br-1974-rabbitmq-2
DP_BACKUP_BASE_PATH=.../rmq-br-1974-rabbitmq-2
```

The restore action rejected that value before `datasafed pull`, all three Jobs reached `BackoffLimitExceeded`, and B/R T05 was the full run's sole failure (`PASS=396 FAIL=1 SKIP=12`).

## Fix

Restore the parent behavior from `0c027e5d`: prefer `DP_TARGET_POD_NAME`, otherwise use the injected single-segment `DP_TARGET_RELATIVE_PATH` directly as the archive stem. `DP_BACKUP_BASE_PATH` remains unchanged because DP already scopes it to the target pod.

This is a one-commit correction on top of the exact integrated candidate branch.

## Evidence

- TDD RED on `f48116a5`: focused contract spec 14 examples / 1 failure with the observed single-segment input; stderr was the exact runtime rejection.
- GREEN on head `80f98fe3`: focused 14/0; all RabbitMQ ShellSpecs 28/0.
- `bash -n`, ShellCheck, and `git diff --check` pass.
- Runtime evidence and sealed artifact are linked on #3183: <https://github.com/apecloud/kubeblocks-addons/pull/3183#issuecomment-4963853648>.

Evidence scope is one full run / one topology with three target PVCs. A focused B/R rerun on this exact head is required before broader acceptance.

### DIFF
--- a/addons/rabbitmq/dataprotection/restore.sh
+++ b/addons/rabbitmq/dataprotection/restore.sh
@@ -3,9 +3,26 @@ set -Eeuo pipefail
 
 DATA_DIR="${DATA_DIR:-/var/lib/rabbitmq}"
 # Backup jobs inject DP_TARGET_POD_NAME. Volume-populator AsDataSource prepareData
-# may only inject DP_TARGET_RELATIVE_PATH (= pod identity / archive stem). Prefer
-# POD_NAME when present; otherwise accept RELATIVE_PATH so KB12 dataSource restore works.
-TARGET_POD_NAME="${DP_TARGET_POD_NAME:-${DP_TARGET_RELATIVE_PATH:?DP_TARGET_POD_NAME or DP_TARGET_RELATIVE_PATH is required}}"
+# may only inject DP_TARGET_RELATIVE_PATH, either as <pod-name> or
+# <target-name>/<pod-name>. DP_BACKUP_BASE_PATH is already scoped to that path,
+# so the archive key always uses only the final pod-name segment.
+if [ -n "${DP_TARGET_POD_NAME:-}" ]; then
+  TARGET_POD_NAME="${DP_TARGET_POD_NAME}"
+else
+  TARGET_RELATIVE_PATH="${DP_TARGET_RELATIVE_PATH:?DP_TARGET_POD_NAME or DP_TARGET_RELATIVE_PATH is required}"
+  case "${TARGET_RELATIVE_PATH}" in
+    /*|*/|*/*/*)
+      echo "ERROR: DP_TARGET_RELATIVE_PATH must be one pod segment or <target-name>/<target-pod-name>" >&2
+      exit 1
+      ;;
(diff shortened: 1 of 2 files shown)


weicao:
Blocker: this fix restores the observed one-segment case, but regresses the other valid KubeBlocks DP restore shape.

`GetTargetRelativePath(targetName, targetPodName)` explicitly emits either `pod-0` when `targetName` is empty or `target/pod-0` when it is present; upstream unit tests pin both shapes: https://github.com/apecloud/kubeblocks/blob/51236eb8db6ff5733d56447a37292084ecbc9b2e/pkg/dataprotection/restore/utils_test.go#L153-L158

At this head, the two-segment input becomes `ARCHIVE_NAME=target/pod-0.tar.zst`. RabbitMQ backup writes `pod-0.tar.zst` inside the already target-scoped `DP_BACKUP_BASE_PATH`, so restore will look for a nested key that was never written. The previous candidate handled only two segments; this PR handles only one. The addon needs to accept both documented controller outputs and derive the archive stem from the final non-empty segment, while failing closed on malformed/extra-depth input. Please retain/add focused tests for both:

- `DP_TARGET_RELATIVE_PATH=pod-0` -> `pod-0.tar.zst`
- `DP_TARGET_RELATIVE_PATH=target/pod-0` -> `pod-0.tar.zst`

Keep `DP_TARGET_POD_NAME` precedence and the missing-input failure. No issue with the observed KB12 diagnosis or the rest of the restore path.

weicao:
Re-review on exact head `e1dc5c0b6c402f24b0f127850492e2ce2d65a384`: blocker resolved.

The production parser now preserves `DP_TARGET_POD_NAME` precedence, accepts both controller-pinned shapes (`pod-0` and `target/pod-0`), derives the final pod archive stem, and fails closed for leading/trailing-empty or extra-depth paths. Focused specs exercise both legal shapes through `restore.sh`, malformed inputs, and precedence.

Independent verification: focused ShellSpec 17/0 on Bash 3.2.57; `bash -n`, ShellCheck on `restore.sh`, and `git diff --check` pass. GitHub shell-check, ShellSpec, chart render, and generate-files checks are all SUCCESS on the exact head.

Focused code/contract: APPROVE. Runtime remains separate: current exact-head B/R is RUNNING/N=0 until terminal evidence; fresh full still required after focused GREEN.

PR #3200, opened 2026-07-14
fix(rustfs): reject scale-up that creates pools too small for inherited parity

weicao:
## Summary

- Validate every RustFS erasure pool against the parity inherited from pool 0 before startup.
- Derive each pool's actual erasure-set size using the beta.8 `RUSTFS_ERASURE_SET_DRIVE_COUNT` semantics.
- Derive pool-0 parity using the beta.8 `RUSTFS_STORAGE_CLASS_STANDARD` semantics.
- Prevent deterministic `TooFewDataShards` startup panics without rejecting valid configured layouts.

## Root cause

RustFS beta.8 freezes `common_parity_drives` from the first pool's actual erasure set and applies it to later pools. If a later pool has `drives_per_set <= common_parity_drives`, its data-shard count becomes zero or negative and Reed-Solomon initialization panics.

The addon guard originally used raw pool sizes and default parity only. That diverged when RustFS split a pool into smaller erasure sets or when operators configured set size / STANDARD parity through environment variables.

## Configuration contract

- Unset or explicit-zero `RUSTFS_ERASURE_SET_DRIVE_COUNT` uses automatic layout.
- A set value follows 64-bit Rust `usize` parsing; single-node layout bypass occurs only after successful parsing.
- Non-single layouts require a configured set size in 2..16 that divides every pool.
- Empty `RUSTFS_STORAGE_CLASS_STANDARD` uses default parity; non-empty values must be `EC:<usize>` and no larger than half the first set.
- Later pools use the first pool parity, matching beta.8.

## Test plan

- [x] Red-first configuration regressions on the pre-fix implementation.
- [x] Focused RustFS startup ShellSpec: 47 examples, 0 failures.
- [x] Positive boundaries: explicit set `0`, single-node set `1` / `u64::MAX`, STANDARD `EC:0`, and parity exactly equal to half the first set.
- [x] Four targeted mutants are killed by only the new positive cases: 47/1, 47/2, 47/1, 47/1.
- [x] Covers configured/automatic layout, single-node ordering, leading `+` / zeroes, malformed/empty values, and 64-bit overflow before shell arithmetic.
- [x] Direct dash behavioral checks.
- [x] `sh -n`, `bash -n`, ShellCheck error, Helm lint/render, and `git diff --check` pass.

## Evidence boundary

These are code and static-test results only; they do not claim a Kubernetes runtime PASS or release readiness.

### DIFF
--- a/addons/rustfs/scripts/startup.sh
+++ b/addons/rustfs/scripts/startup.sh
@@ -72,6 +72,28 @@ default_parity_count() {
   esac
 }
 
+# RustFS splits each pool into erasure sets whose size is the largest
+# factor of pool_size in [2,16].  Parity is derived from that set size,
+# not from the raw pool drive count.
+drives_per_set() {
+  local pool_size=$1
+  if [ "$pool_size" -le 16 ]; then
+    echo "$pool_size"
+    return 0
+  fi
+  local d=16
+  while [ $d -ge 2 ]; do
+    if [ $((pool_size % d)) -eq 0 ]; then
+      echo "$d"
+      return 0
+    fi
+    d=$((d - 1))
+  done
+  echo "FATAL: pool has $pool_size drives which is not divisible by any erasure set size in [2,16]." >&2
+  echo 0
+  return 1
+}
+
 validate_pool_sizes() {
   local replicas=$1
   local prev=0
@@ -81,13 +103,18 @@ validate_pool_sizes() {
   IFS=','
   for cur in $replicas; do
     local pool_size=$((cur - prev))
+    local set_size
+    set_size=$(drives_per_set "$pool_size") || {
+      IFS="$old_ifs"
+      return 1
+    }
     if [ $pool_index -eq 0 ]; then
-      first_pool_parity=$(default_parity_count "$pool_size")
+      first_pool_parity=$(default_parity_count "$set_size")
     fi
-    if [ "$pool_size" -le "$first_pool_parity" ]; then
-      echo "FATAL: erasure pool $pool_index has $pool_size drive(s) but inherited parity is $first_pool_parity (from pool 0)." >&2
-      echo "data_blocks would be $((pool_size - first_pool_parity)) which causes TooFewDataShards panic." >&2
-      echo "Scale to a replica count where every pool has at least $((first_pool_parity + 1)) drives." >&2
+    if [ "$set_size" -le "$first_pool_parity" ]; then
+      echo "FATAL: erasure pool $pool_index has drives_per_set=$set_size but inherited parity is $first_pool_parity (from pool 0)." >&2
+      echo "data_blocks would be $((set_size - first_pool_parity)) which causes TooFewDataShards panic." >&2
+      echo "Scale to a replica count where every pool's erasure set has at least $((first_pool_parity + 1)) drives." >&2
       IFS="$old_ifs"
       return 1
     fi
--- a/addons/rustfs/scripts/startup.sh
+++ b/addons/rustfs/scripts/startup.sh
@@ -61,6 +61,43 @@ read_replicas_history() {
   echo "$content"
 }
 
+default_parity_count() {
+  local drives=$1
+  case $drives in
+    1) echo 0 ;;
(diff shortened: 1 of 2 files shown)


addons/rustfs/scripts/startup.sh
```diff
@@ -61,6 +61,43 @@ read_replicas_history() {
...
+  local replicas=$1
+  local prev=0
+  local first_pool_parity=""
+  local pool_index=0
+  local old_ifs="$IFS"
+  IFS=','
+  for cur in $replicas; do
+    local pool_size=$((cur - prev))
+    if [ $pool_index -eq 0 ]; then
+      first_pool_parity=$(default_parity_count "$pool_size")
```
leon-ape: `default_parity_count` is applied to the pool total replica count, while RustFS beta.8 derives `common_parity_drives` from `pool_eps.drives_per_set` after splitting the pool into erasure sets. For history `25,28`, pool 0 is split into 5-drive sets (parity 2), so the 3-drive second pool is valid; this guard computes parity 4 from 25 and rejects the valid scale-up.

weicao: Good catch — you're right. The guard was computing parity from raw pool size instead of `drives_per_set` (the largest factor of pool size in [2,16], matching RustFS's `common_set_drive_count()`).

Fixed in 8361569: added a `drives_per_set()` function that mirrors the upstream algorithm and now the guard uses `default_parity_count(drives_per_set)` instead of `default_parity_count(pool_size)`. Your `25,28` example (pool 0 → 5-drive sets → parity 2, pool 1 = 3 drives → valid) now passes correctly. Added 8 new tests including that case.

addons/rustfs/scripts/startup.sh
```diff
@@ -61,6 +61,70 @@ read_replicas_history() {
...
+  IFS=','
+  for cur in $replicas; do
+    local pool_size=$((cur - prev))
+    local set_size
+    set_size=$(drives_per_set "$pool_size") || {
+      IFS="$old_ifs"
+      return 1
+    }
+    if [ $pool_index -eq 0 ]; then
+      first_pool_parity=$(default_parity_count "$set_size")
```
leon-ape: RustFS beta.8 reads `RUSTFS_ERASURE_SET_DRIVE_COUNT` when selecting erasure sets and `RUSTFS_STORAGE_CLASS_STANDARD` in `ec_drives_no_config`, but this guard ignores both. For example, with history `4,6` and `RUSTFS_STORAGE_CLASS_STANDARD=EC:1`, RustFS uses parity 1 and the 2-drive second pool is valid, while this code hardcodes parity 2 and exits before startup.

weicao: Fixed in 71edf548. The startup guard now mirrors beta.8 config semantics: it parses RUSTFS_ERASURE_SET_DRIVE_COUNT before layout selection (0/unset = automatic; non-single pools require a 2..16 divisor for each pool; single-node bypass follows successful usize parsing) and derives the first-pool parity from RUSTFS_STORAGE_CLASS_STANDARD (empty = default, otherwise EC:<usize>). Parsing accepts Rust-compatible leading + and zeroes, rejects u64 overflow before shell arithmetic, and keeps later pools on the first pool parity. The 4,6 + EC:1 case now passes. Focused startup ShellSpec is 42/42, with dash behavioral checks, sh/bash syntax, ShellCheck error, Helm lint/render, and diff-check passing.

leon-ape (dismissed):
LGTM

addons/rustfs/scripts/startup.sh
weicao: Follow-up test hardening in 7afe995d: added positive contract cases for explicit set size 0 (automatic), set size 1 and u64::MAX in single-node mode, STANDARD EC:0, and parity exactly equal to half the first set. Focused startup ShellSpec is now 47/47. Four targeted mutants are killed by only these new cases; the production startup.sh blob is unchanged from 71edf548.

leon-ape (approved):
LGTM

PR #3201, opened 2026-07-14
chore: align example service versions and remove unused Reconfiguration ops

shanshanying:


examples/clickhouse/README.md
```diff
@@ -15,13 +15,13 @@ There are two key components in the ClickHouse cluster:
 
 | Topology           | Horizontal scaling | Vertical scaling | Expand volume | Restart | Stop/Start | Configure | Expose | Switchover |
 | ------------------ | ------------------ | ---------------- | ------------- | ------- | ---------- | --------- | ------ | ---------- |
-| standalone/cluster | Yes                | Yes              | Yes           | Yes     | Yes        | Yes       | Yes    | N/A        |
+| standalone/cluster | Yes                | Yes              | Yes           | Yes     | Yes        | No        | Yes    | N/A        |
```
leon-ape, line 18: Target `release-1.1` still supports `OpsRequest` type `Reconfiguring`, and the ClickHouse chart renders `ParametersDefinition`/`ParamConfigRenderer` for `user.xml` and `00_default_overrides.xml`. This change marks Configure as unsupported and removes a capability that remains exposed by the target API and addon.

examples/postgresql/README.md
```diff
@@ -80,7 +81,7 @@ spec:
     - name: postgresql
       # ServiceVersion specifies the version of the Service expected to be
       # provisioned by this Component.
-      # Valid options are: [12.14.0,12.14.1,12.15.0,12.22.0,14.7.2,14.8.0,14.18.0,15.7.0,15.13.0,16.4.0,16.9.0,17.5.0]
+      # Valid options are: [12.14.0,12.14.1,12.15.0,12.22.0,14.7.2,14.8.0,14.18.0,14.23.0,15.7.0,15.13.0,15.18.0,16.4.0,16.9.0,16.14.0,17.5.0,17.10.0,18.1.0,18.4.0]
```
leon-ape, line 84: This updated list adds `14.23.0`, `15.18.0`, `16.14.0`, `17.10.0`, `18.1.0`, and `18.4.0`, but the `kubectl get cmpv postgresql` expected output below still omits all of them, so the same README reports contradictory supported versions.

examples/zookeeper/README.md
```diff
@@ -8,7 +8,7 @@ Apache Zookeeper is a centralized service for maintaining configuration informat
 
 | Horizontal<br/>scaling | Vertical <br/>scaling | Expand<br/>volume | Restart   | Stop/Start | Configure | Expose | Switchover |
 |------------------------|-----------------------|-------------------|-----------|------------|-----------|--------|------------|
-| Yes                    | Yes                   | Yes              | Yes       | Yes        | Yes       | Yes    | No      |
+| Yes                    | Yes                   | Yes              | Yes       | Yes        | No        | Yes    | No      |
```
leon-ape, line 11: The deleted examples are still referenced by `addons/clickhouse/README.md`, `addons/mogdb/README.md`, `addons/rabbitmq/README.md`, `addons/weaviate/README.md`, and `addons/zookeeper/README.md`; those documents now contain commands pointing to nonexistent files.

shanshanying: Fixed in 6718fe168. Removed the stale Reconfigure sections from the ClickHouse, MogDB, RabbitMQ, Weaviate, and ZooKeeper addon READMEs. A scoped link check confirms the remaining example YAML references in those READMEs exist.

examples/clickhouse/README.md
shanshanying: The release-1.1 API still includes Reconfiguring and the chart creates ParametersDefinition/ParamConfigRenderer, but the parameter controller only resolves ComponentDefinition configs marked externalManaged: true (pkg/parameters/parameter_utils.go, ResolveParameterTemplate). ClickHouse user.xml and 00_default_overrides.xml configs do not set that field, so those objects alone do not make this OpsRequest functional. We are keeping the invalid OpsRequest example removed; commit 6718fe168 also removes the stale generated README section while retaining Using Config Templates.

examples/postgresql/README.md
shanshanying: Fixed in 6718fe168. The expected kubectl output now includes every serviceVersion rendered by the PostgreSQL CMPV, and the generated addon README was updated as well.

leon-ape: `examples/postgresql/README.md` is fixed, but `addons/postgresql/README.md` still keeps the old Versions table and both `Valid options` snippets; only its expected CMPV output was updated. The addon README remains internally inconsistent.

examples/mongodb/README.md
```diff
@@ -19,12 +19,13 @@ MongoDB is a document database designed for ease of application development and
...
-| 5.0 | 5.0.28 |
-| 6.0 | 6.0.16 |
-| 7.0 | 7.0.12 |
+| Major Versions | Service Versions |
+|---------------|------------------|
+| 4.x | 4.0.28, 4.4.29 |
+| 5.x | 5.0.29 |
+| 6.x | 6.0.27 |
+| 7.x | 7.0.28 |
+| 8.x | 8.0.17 |
```
leon-ape, line 28: Running `hack/extend-readme-with-yaml.sh` on the current head changes 25 tracked `addons/*/README.md` files. The updated example version tables were not propagated to the packaged addon READMEs; for example, `addons/mongodb/README.md` still advertises 4.2.24/5.0.28/6.0.16/7.0.12 while this table and the rendered ComponentVersion use the new versions.

examples/tidb/README.md
```diff
@@ -17,7 +17,7 @@ TiDB is an open-source, cloud-native, distributed, MySQL-Compatible database for
 | 8.4.0  |
 | 7.5.2  |
 | 7.1.5  |
-| 6.5.10 |
+| 6.5.12 |
```
leon-ape, line 20: Current release-1.1 now publishes 7.1.6-enterprise for the TiDB, PD, and TiKV ComponentVersions, but this rewritten Versions table omits it. Merging the current head leaves both README copies inconsistent with the target chart.

leon-ape (approved):
LGTM

PR #3207, opened 2026-07-14
fix(mongodb): quote parameter ownership matchers

weicao:
## Problem

MongoDB `ParametersDefinition.spec.componentDef` interpolates rendered ComponentDefinition names as unescaped regular expressions. Chart-version dots therefore match arbitrary characters, so a matcher for `mongo-config-server-1.2.0-alpha.0` also accepts near-collisions such as `mongo-config-server-1x2.0-alpha.0`.

The two ParametersDefinition objects also carry `helm.sh/resource-policy: keep`. Their historical metadata names must remain stable across upgrades; this PR intentionally does not derive them from user prefixes.

## Change

- quote every rendered ComponentDefinition name with Helm `regexQuoteMeta` before constructing the anchored ownership regex
- preserve the historical `mongodb-config-pd-<chartVersion>` and `mongos-config-pd-<chartVersion>` object identities
- add a focused ShellSpec covering default, `resourceNamePrefix`, `cmpdVersionPrefix`, and combined renders
- assert exact single ownership, reject dotted near-collisions, and assert retained identities remain stable across all renders
- materialize Helm dependencies separately so repository-status output cannot enter YAML parsing

## TDD / verification

Base: `41801859b685dea4f8d603b7aa4b90a7fb52e2d1`
Head: `8e2601669b29472230ed73b4696205cf2ade2cb1`

- focused ShellSpec against exact base: RED, 1 example / 1 failure; first error is regex matching `mongo-config-server-1x2.0-alpha.0`
- focused ShellSpec against exact head on Bash 3.2 and Bash 5.3: GREEN, 1/0 each
- all MongoDB ShellSpec against exact head: 10/0
- Helm lint: 1 chart / 0 failures
- ShellCheck and `git diff --check`: pass

## Scope and boundary

Net scope is two files: the two quoted matchers and their regression spec. The reviewed prefix-derived naming helpers and metadata-name changes are removed completely because retained-object renames cannot guarantee upgrade-safe single ownership.

All evidence is offline/static. MongoDB runtime remains `N=0/NOT_RUN`; this PR does not claim runtime PASS or release readiness.

### DIFF
--- a/addons/mongodb/templates/_helpers.tpl
+++ b/addons/mongodb/templates/_helpers.tpl
@@ -188,23 +188,3 @@ Define config server component definition name prefix
 {{- define "cfgServer.componentDefNamePrefix" -}}
 {{- printf "mongo-config-server-" -}}
 {{- end -}}
-
-{{/*
-Define the ParametersDefinition name prefix. ParametersDefinition is
-cluster-scoped, so its name must follow both chart isolation dimensions.
-*/}}
-{{- define "mongodb.parametersDefinitionNamePrefix" -}}
-{{- if ne (len .Values.resourceNamePrefix) 0 -}}
-{{ .Values.resourceNamePrefix }}-
-{{- end -}}
-{{- end -}}
-
-{{/*
-Define the ParametersDefinition version suffix shared with ComponentDefinitions.
-*/}}
-{{- define "mongodb.parametersDefinitionVersion" -}}
-{{- if ne (len .Values.cmpdVersionPrefix) 0 -}}
-{{ .Values.cmpdVersionPrefix }}-
-{{- end -}}
-{{ .Chart.Version }}
-{{- end -}}
--- a/addons/mongodb/templates/_helpers.tpl
+++ b/addons/mongodb/templates/_helpers.tpl
@@ -188,3 +188,23 @@ Define config server component definition name prefix
 {{- define "cfgServer.componentDefNamePrefix" -}}
 {{- printf "mongo-config-server-" -}}
 {{- end -}}
+
+{{/*
+Define the ParametersDefinition name prefix. ParametersDefinition is
+cluster-scoped, so its name must follow both chart isolation dimensions.
+*/}}
+{{- define "mongodb.parametersDefinitionNamePrefix" -}}
+{{- if ne (len .Values.resourceNamePrefix) 0 -}}
+{{ .Values.resourceNamePrefix }}-
+{{- end -}}
+{{- end -}}
+
+{{/*
+Define the ParametersDefinition version suffix shared with ComponentDefinitions.
+*/}}
+{{- define "mongodb.parametersDefinitionVersion" -}}
+{{- if ne (len .Values.cmpdVersionPrefix) 0 -}}
+{{ .Values.cmpdVersionPrefix }}-
+{{- end -}}
+{{ .Chart.Version }}
+{{- end -}}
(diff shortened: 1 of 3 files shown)


addons/mongodb/templates/paramsdef.yaml
```diff
@@ -2,13 +2,13 @@
 apiVersion: parameters.kubeblocks.io/v1alpha1
 kind: ParametersDefinition
 metadata:
-  name: {{ printf "mongodb-config-pd-%s" .Chart.Version }}
+  name: {{ printf "%smongodb-config-pd-%s" (include "mongodb.parametersDefinitionNamePrefix" .) (include "mongodb.parametersDefinitionVersion" .) }}
```
leon-ape: With either prefix set, this renames a ParametersDefinition that carries `helm.sh/resource-policy: keep`. Upgrading an existing release retains the old object and creates this one; both match the same ComponentDefinition and `mongodb.conf`, so KubeBlocks rejects parameter resolution as duplicate file ownership.

leon-ape: Combining both unrestricted prefixes can exceed Kubernetes' 253-character metadata.name limit. Two 111-character prefixes keep the chart's separately prefixed resources within the limit but render this ParametersDefinition name at 255 characters, causing creation to fail.

addons/mongodb/templates/_helpers.tpl
```diff
@@ -187,4 +187,24 @@ Define config server component definition name prefix
...
+{{ .Values.resourceNamePrefix }}-
+{{- end -}}
+{{- end -}}
+
+{{/*
+Define the ParametersDefinition version suffix shared with ComponentDefinitions.
+*/}}
+{{- define "mongodb.parametersDefinitionVersion" -}}
+{{- if ne (len .Values.cmpdVersionPrefix) 0 -}}
+{{ .Values.cmpdVersionPrefix }}-
```
leon-ape: The two prefixes are not encoded uniquely. `(resourceNamePrefix=a, cmpdVersionPrefix=b-mongodb-config-pd-c)` and `(resourceNamePrefix=a-mongodb-config-pd-b, cmpdVersionPrefix=c)` produce the same cluster-scoped ParametersDefinition name, so otherwise isolated releases conflict.

addons/mongodb/templates/paramsdef.yaml
weicao: Fixed in 8e26016 by removing the ParametersDefinition rename entirely. Both objects are retained, so the only upgrade-safe scope here is to preserve the historical `mongodb-config-pd-<chartVersion>` / `mongos-config-pd-<chartVersion>` identities and keep this PR limited to regex quoting. The focused four-render spec now asserts those retained identities remain stable while every current ComponentDefinition is covered exactly once.

weicao: Fixed in 8e26016 by deleting the unrestricted prefix concatenation. ParametersDefinition metadata names are back to the bounded historical chart-version form; prefixes affect only the quoted ComponentDefinition matcher. The focused spec asserts stable retained names across default, resourceNamePrefix, cmpdVersionPrefix, and combined renders.

addons/mongodb/templates/_helpers.tpl
weicao: Fixed in 8e26016 by removing the concatenated prefix encoding rather than attempting another delimiter/hash migration for retained objects. The PR no longer derives ParametersDefinition names from either prefix, so the reported tuple collision is absent; only regex quoting remains as the product change.

leon-ape (approved):
LGTM

PR #3210, opened 2026-07-15
fix: add mysql57 performance_schema parameter schema

xuriwuyun:
## What changed
- add the missing `performance_schema` schema entry to the MySQL 5.7 parameter constraint

## Why
- `mysql5.7-config-effect-scope.yaml` already classifies `performance_schema` as a static parameter
- but `mysql5.7-config-constraint.cue` did not define the parameter itself
- that mismatch can break parameter API validation/rendering for MySQL 5.7 reconfiguration

## Impact
- the MySQL 5.7 parameter API can expose and validate `performance_schema` consistently
- the parameter remains restart-only rather than dynamic

## Root cause
- effect-scope and parameter schema drifted out of sync for MySQL 5.7

## Related issue
- Refs apecloud/apecloud#19172

## Validation
- checked the diff only
- ran `git diff --check` on the modified file


leon-ape (approved):
LGTM

wangyelei (approved):

PR #3216, opened 2026-07-15
mariadb: preserve secondary fence marker while pending

weicao:
## Problem

Fresh focused r18 on addon `ccd5c0e81ffaafa9b71c9f38c09b00a46246465c` and syncer `7502b58f45efd4fe8f2b06e5313230d9aec69510` showed the primary staying converged, but a healthy secondary's durable remote-root fence marker oscillating between `secondary` and absent.

The exact cross-script cause is in `replication-entrypoint.sh`: every `configure_replication_from_primary_service_once()` iteration begins with `mark_replication_pending()`, and that generic readiness helper unconditionally removed `.remote-root-fence-role`. The roleProbe later recreated `secondary` only after its successful grant/fence tick. A recurring runtime-secondary reconcile therefore exposed an absent-marker window even though the secondary remained `read_only=1` and otherwise healthy.

Evidence scope is one focused run (`N=1`), not a frequency claim:

- official seven-sample window: secondary marker `secondary` in 6 samples, absent in sample 7;
- independent read-only post-red window: `secondary` 18 times, absent twice;
- watchdog shows a new runtime-secondary reconcile beginning before the absent observation;
- primary was stable in 7/7 samples with `read_only=0`, all real ready markers, no pending/master.info/remote marker, and no new repair or authority commit;
- full/Ops/fault were not launched and every product scene remains frozen.

Evidence archive outer SHA-256: `f74b470331c40f075f366a59bd9fc0a35a7388ac045a50b9f04ae15a9d6ddd73`; inner manifest SHA-256: `7a86cfcdeed8fa9cef2c47fd53fb286f5c70b965592a58c9e446b366c400961a`; 93/93 files independently verified.

## Fix

Remove `.remote-root-fence-role` mutation from generic `mark_replication_pending()`.

This keeps ownership aligned with the durable contract introduced by the parent PR:

- `secondary` means a secondary remote-root fence is active;
- marker absence means committed-primary state;
- generic readiness/pending transitions retract ready markers and publish `.replication-pending`, but do not rewrite durable fence state;
- role-specific code removes the marker only after a primary transition has repaired grants successfully, and writes `secondary` only after the secondary fence is actually in place.

No read-only, root-account, listener, authority acquisition/commit, switchover, error rollback, or ready-publication operation is removed or reordered.

## TDD and gates

Before the production change, the exact-source test was deterministically RED:

- `74 examples, 2 failures`;
- static ownership check proved generic pending still referenced the marker;
[...]

PR #3217, opened 2026-07-15
fix(mariadb): make Galera preStop ordered and DNS-safe

weicao:
## Problems observed

This PR closes two evidence-backed failures in the same Galera native `preStop` path.

1. In the r21 frozen scene, Stop and Start OpsRequests both reported `Succeed`, but the data plane stayed failed. Every PVC recorded `mysqladmin_shutdown_failed`; pod-0 and pod-1 also exhausted the higher-ordinal wait. The hook had consumed the Pod grace budget with SQL and `mysqladmin shutdown` before kubelet could signal mariadbd PID 1.
2. After the first fix, the fresh r22 scene recovered to Primary/size3/all-Synced/rows9, but the hook contract still failed. pod-1 was deleted at 11:19:13Z and pod-2 at 11:19:20Z; pod-0 then waited the full 60 seconds, Kubernetes emitted `FailedPreStopHook` at 11:20:07Z, and pod-0's PVC persisted `reason=order_wait_timeout` with both deleted higher peers marked `budget-exhausted`. The hook treated their post-deletion DNS name absence as indefinite uncertainty.

The second result means r22 is **runtime FAIL/N=1 for preStop validation**, despite successful data-plane recovery.

## Root cause

`galera-prestop.sh` is a Kubernetes native container `preStop`, not a kbagent lifecycle action. Kubernetes starts the Pod termination grace countdown before the hook and sends TERM to container PID 1 only after the hook returns.

The old code had two separate contract violations:

- it ran client-side SQL/`mysqladmin shutdown` inside the hook and swallowed failures, delaying kubelet's real TERM;
- its bounded TCP probe only recognized `Connection refused` as stopped. Once a deleted higher Pod disappeared from headless-Service DNS, `Name or service not known` / `No address associated` could never close that peer, so the lower ordinal exhausted the 60-second wait.

Primary contracts:

- Kubernetes native hook ordering and shared grace budget: <https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/>
- MariaDB Galera orderly shutdown / last clean node safe-to-bootstrap contract: <https://mariadb.com/docs/galera-cluster/high-availability/resetting-the-quorum-cluster-bootstrap>

## Behavior fix

- Keep `preStop` responsible only for bounded high-to-low ordering and publishing `.galera-shutting-down` for the watcher guard.
- Let kubelet send TERM to the `exec`-owned mariadbd PID 1; do not run SQL or `mysqladmin` in the hook.
- Keep `Connection refused` as an immediate closed observation.
- Classify only authoritative name absence (`Name or service not known` / `No address associated`) as `absent`, and require the same peer to be absent in **two consecutive polls** before confirming it stopped.
- Clear that peer's absence streak on any live or uncertain observation (`open`, transient DNS failure, timeout, generic unreachable, or unknown) and continue fail-closed.
- Track peers independently, remember confirmed peers, preserve the existing wall-clock deadline, and return non-zero with persisted degraded evidence on unresolved peers.
- No K8s API, RBAC, sidecar, shared-volume coordination, or new engine shutdown mechanism is added.
- Bump the chart to `1.2.0-alpha.34` because the ComponentDefinition runtime contract changed.

The Chinese design document was published in the MariaDB team thread before this second code change. It covers the problem, alternatives, core flow, compatibility, and split Dev/Test plan.

## TDD and local source validation

First fix:

- Old parent focused RED: `22 examples / 6 failures`.
- Patched focused GREEN: `22 / 0`.

DNS/deleted-peer fix on exact `90434973494b6b8207b287e7ea5bbd2cc3292307`:

- Production RED with tests only: `28 examples / 6 failures`; failures covered authoritative absence classification, transient classification, two-poll confirmation, uncertain-state reset, and independent peer state.
- Current Bash 3.2 focused GREEN: `29 / 0`.
- Related Galera/switchover/version specs on Bash 5.3: `250 / 0 / 6 documented pending`.
- Full MariaDB ShellSpec on Bash 5.3: `777 / 0 / 9 documented pending`.
- `bash -n`, ShellCheck error gate, and `git diff --check`: PASS.
- Helm dependency build, lint, default render, and package: PASS.
- Offline package proof only: `mariadb-1.2.0-alpha.34.tgz` SHA256 `51af67d5c73bf32a8e2957f1faabc4b881d10a45a8a5d62f7b35d7e9a6e2dbac`; default render SHA256 `95890869b273f7d8eba1e29813028a426596f02b3c374971643ea4112246e447`.

## Current boundary

- Exact head: `c509f506861d21e88ce16f4498ac08ae3fce7e46`.
- CI: rerunning for this exact head; not yet terminal at the time of this update.
- Runtime: **no post-`c509f5068` runtime result**. The r22 scene remains frozen read-only and is not evidence for the new source.
- Merge order: this PR is stacked on PR #3216 and must follow it.
- A fresh Test-owned immutable package and Galera Stop/Start run must show no `FailedPreStopHook`, no new persisted degraded record, and the exact Primary/size3/Synced/rows oracle before any runtime-fixed claim.

### DIFF
--- a/addons/mariadb/Chart.yaml
+++ b/addons/mariadb/Chart.yaml
@@ -4,7 +4,7 @@ description: MariaDB is a high performance open source relational database manag
 
 type: application
 
-version: 1.2.0-alpha.32
+version: 1.2.0-alpha.33
 
 appVersion: "11.4.10"
 
--- a/addons/mariadb/Chart.yaml
+++ b/addons/mariadb/Chart.yaml
@@ -4,7 +4,7 @@ description: MariaDB is a high performance open source relational database manag
 
 type: application
 
-version: 1.2.0-alpha.33
+version: 1.2.0-alpha.34
 
 appVersion: "11.4.10"
 
--- a/addons/mariadb/scripts-ut-spec/galera_memberleave_removed_spec.sh
+++ b/addons/mariadb/scripts-ut-spec/galera_memberleave_removed_spec.sh
@@ -4,9 +4,10 @@
 # kbagent execution face (no mariadb client), all SQL guarded with
 # `2>/dev/null || true`, so its claimed FLUSH TABLES / wsrep_desync=ON never
 # executed. Galera evicts a departed node natively (keepalive timeout) and
-# the shutdown-time graceful leave (final seqno + safe_to_bootstrap, clean
-# InnoDB flush) is handled by the mariadb container preStop, which runs
-# in-container against 127.0.0.1. The broken action + its script were removed.
+# the native container preStop orders peer termination and publishes the
+# watcher guard. Kubelet then signals mariadbd PID 1, which performs the
+# shutdown-time graceful leave (final seqno + safe_to_bootstrap, clean InnoDB
+# flush). The broken action + its script were removed.
 
 Describe "galera memberLeave removal (H10)"
   CMPD_GALERA="${SHELLSPEC_CWD:?}/addons/mariadb/templates/cmpd-galera.yaml"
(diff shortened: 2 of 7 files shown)


weicao:
Superseded by #3276. This PR's current head c509f506861d21e88ce16f4498ac08ae3fce7e46 is included in the consolidated candidate 7978812cf8ca066cfc13ac6baf0ddd78505e0965; further review and testing should use #3276.

weicao:

PR #3231, opened 2026-07-17
fix: align addon KubeBlocks version constraints

shanshanying:
## What changed

- Require KubeBlocks `>=1.2.0` for 23 addon charts that render APIs introduced after the `release-1.1` schema, including direct `ParametersDefinition` bindings and `ComponentDefinition.spec.configs[].reconfigure`.
- Require KubeBlocks `>1.0.2` for MogDB, Orchestrator, and PolarDB-X because they render exclusive replica roles.
- Require KubeBlocks `>=1.0.2` for Elasticsearch because it renders `podUpgradePolicy`.
- Include MariaDB and Valkey, as identified in review feedback.

## Why

Several addon charts declared a KubeBlocks version floor that could not accept their rendered custom resources. This allowed incompatible addons to be selected and then rejected by the target KubeBlocks API server.

The constraints are derived from full rendered-manifest comparison against structural CRD schemas, rather than from a checker hardcoded to individual API fields.

## Impact

Addon compatibility metadata now prevents installation on unsupported KubeBlocks versions:

- 23 addons using post-1.1 configuration APIs require the 1.2 line.
- Addons using only the exclusive-role API remain compatible with the post-1.0.2 line.
- Elasticsearch remains compatible with KubeBlocks 1.0.2 and later.

This `main` PR is intentionally labeled `nopick`; release-branch constraints are handled in a separate PR based on the resources actually rendered by `release-1.1`.

## Validation

- Refreshed and compared against KubeBlocks `v1.0.1`, `v1.0.2`, `v1.0.3-beta.0`, `release-1.1`, and `main` CRD schemas.
- Rendered and linted all 42 installable addon charts.
- Confirmed the 23 `>=1.2.0` addons fail `release-1.1` structural validation and pass `main`.
- Confirmed Elasticsearch first passes at `v1.0.2`.
- Confirmed MogDB, Orchestrator, and PolarDB-X fail `v1.0.2` and pass `v1.0.3-beta.0`.
- Ran `git diff --check`.
- Confirmed the final diff changes only one compatibility annotation in each of 27 `Chart.yaml` files.

### DIFF
--- a/addons/apecloud-mysql/Chart.yaml
+++ b/addons/apecloud-mysql/Chart.yaml
@@ -31,6 +31,6 @@ dependencies:
     alias: extra
 
 annotations:
-  addon.kubeblocks.io/kubeblocks-version: ">=1.0.0"
+  addon.kubeblocks.io/kubeblocks-version: ">=1.2.0"
   addon.kubeblocks.io/model: "RDBMS"
   addon.kubeblocks.io/provider: "community"
--- a/addons/clickhouse/Chart.yaml
+++ b/addons/clickhouse/Chart.yaml
@@ -27,6 +27,6 @@ dependencies:
     alias: extra
 
 annotations:
-  addon.kubeblocks.io/kubeblocks-version: ">=1.0.0"
+  addon.kubeblocks.io/kubeblocks-version: ">=1.2.0"
   addon.kubeblocks.io/model: "RDBMS"
   addon.kubeblocks.io/provider: "community"
--- a/addons/elasticsearch/Chart.yaml
+++ b/addons/elasticsearch/Chart.yaml
@@ -30,7 +30,7 @@ maintainers:
     url: https://github.com/vipshop
 
 annotations:
-  addon.kubeblocks.io/kubeblocks-version: ">=1.0.0"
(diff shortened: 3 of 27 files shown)


leon-ape (commented):
addons/mariadb and addons/valkey still declare KubeBlocks >=1.0.0 while rendering roles[].isExclusive; valkey also renders podUpgradePolicy. These charts remain installable on KubeBlocks versions that reject those fields, so the compatibility constraint alignment is incomplete.

shanshanying:
@leon-ape Thanks, this was correct for the previous head `b1c201b`.

It is fixed in the current head `ff37737`: both MariaDB and Valkey now declare KubeBlocks `>=1.2.0`.

I used `>=1.2.0` rather than only `>1.0.2` because the full rendered-schema audit also found that both charts use post-1.1 configuration APIs (`ComponentDefinition.spec.configs[].reconfigure` and direct `ParametersDefinition` bindings). Those resources are rejected by the KubeBlocks `release-1.1` schema and accepted by `main`.

The release-1.1-specific constraints are handled separately in #3238. All checks on this PR currently pass.

leon-ape (approved):
LGTM

PR #3233, opened 2026-07-17
fix(mongodb): verify toggle balancer state

weicao:
## Problem

Omitting `enableBalancer` from the MongoDB toggle-balancer operation currently disables the balancer even though the OpsDefinition schema documents `default: true`.

The operation script uses `$(enableBalancer)`, which Bash treats as command substitution. KubeBlocks injects environment variables only for explicitly supplied custom-operation parameters, so omission runs a nonexistent command, produces an empty value, and selects `sh.stopBalancer()`.

The same script also prints a successful informational message after `sh.startBalancer()` or `sh.stopBalancer()` without checking the command status. That trailing `echo` changes an action failure into an overall rc 0. Finally, the operation never verifies the resulting balancer state.

## Reproduction

A focused test extracts and executes the command from the rendered OpsDefinition:

- before the fix, omission emits `enableBalancer: command not found` and invokes `sh.stopBalancer()`;
- a mocked `sh.startBalancer()` rc 17 becomes overall rc 0;
- a successful start with `sh.getBalancerState()` returning `false` is accepted.

This is rendered-command test evidence, not a completed cluster runtime run.

## Solution

- Resolve omission with `${enableBalancer:-true}` and reject values other than `true` or `false` before mutation.
- Capture and propagate the selected balancer command's status before printing success.
- Read `sh.getBalancerState()` once after command success and require an exact requested/observed match.

The postcheck is intentionally single-shot: the balancer enabled flag is synchronous configuration state, not an asynchronous convergence condition. Retrying here would conceal a failed action/postcondition.

## Verification

- MongoDB ShellSpec: 15 examples, 0 failures
- Helm lint: 1 chart, 0 failures
- Helm render and extracted command `bash -n`
- `shellcheck`
- `git diff --check`

## Boundary

- CI has not run yet.
- Cluster runtime validation remains pending after CI.
- No controller, CRD schema, chart value, or unrelated MongoDB behavior changes are included.

### DIFF
--- a/addons/mongodb/scripts-ut-spec/toggle_balancer_spec.sh
+++ b/addons/mongodb/scripts-ut-spec/toggle_balancer_spec.sh
@@ -1,0 +1,145 @@
+# shellcheck shell=bash
+
+Describe "MongoDB toggle-balancer operation contract"
+
+  render_toggle_balancer_script() {
+    local output_file="$1"
+    local chart_dir
+
+    chart_dir=$(cd .. && pwd)
+    # shellcheck disable=SC2016
+    helm template kb-addon-mongodb "$chart_dir" --dependency-update | ruby -ryaml -e '
+      documents = YAML.load_stream($stdin.read).compact
+      matches = documents.select do |document|
+        document["kind"] == "OpsDefinition" &&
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

PR #3238, opened 2026-07-20
fix: align release-1.1 addon KubeBlocks version constraints

shanshanying:
## What changed

- Require KubeBlocks `>1.0.2` for 15 release-1.1 addon charts that render exclusive replica roles.
- Require KubeBlocks `>=1.0.2` for Elasticsearch because it renders `podUpgradePolicy`.
- Keep MySQL unchanged because it already declares the correct `>1.0.2` floor.

## Why

These release-1.1 charts declared a KubeBlocks version floor that could not accept their rendered custom resources. This allowed incompatible addons to be selected on KubeBlocks 1.0.0 through 1.0.2.

The constraints are derived from full rendered-manifest comparison against structural CRD schemas, rather than from a checker hardcoded to individual API fields.

## Release-1.0 backport

The remote `release-1.0` snapshot renders the same incompatible fields for every chart common to this PR, so this PR is labeled `pick-1.0`.

FalkorDB exists on `release-1.1` but not on `release-1.0`. If the automatic cherry-pick reports a modify/delete conflict, omit `addons/falkordb/Chart.yaml`; the other 15 metadata changes apply to `release-1.0`.

## Validation

- Rendered and linted all 39 installable `release-1.1` addon charts.
- Compared the rendered resources against KubeBlocks `v1.0.1`, `v1.0.2`, `v1.0.3-beta.0`, and `release-1.1` structural CRD schemas.
- Confirmed Elasticsearch fails `v1.0.1` and passes `v1.0.2`.
- Confirmed the 16 exclusive-role addons fail `v1.0.2` and pass `v1.0.3-beta.0`; MySQL already had the correct floor.
- Rendered and linted all 39 installable addons from the remote `release-1.0` snapshot and confirmed the same schema requirements for its common charts.
- Ran `git diff --check`.
- Confirmed the final diff changes only one compatibility annotation in each of 16 `Chart.yaml` files.

### DIFF
--- a/addons/apecloud-mysql/Chart.yaml
+++ b/addons/apecloud-mysql/Chart.yaml
@@ -31,6 +31,6 @@ dependencies:
     alias: extra
 
 annotations:
-  addon.kubeblocks.io/kubeblocks-version: ">=1.0.0"
+  addon.kubeblocks.io/kubeblocks-version: ">1.0.2"
   addon.kubeblocks.io/model: "RDBMS"
   addon.kubeblocks.io/provider: "community"
--- a/addons/clickhouse/Chart.yaml
+++ b/addons/clickhouse/Chart.yaml
@@ -27,6 +27,6 @@ dependencies:
     alias: extra
 
 annotations:
-  addon.kubeblocks.io/kubeblocks-version: ">=1.0.0"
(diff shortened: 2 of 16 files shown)


leon-ape (approved):
LGTM

PR #3251, opened 2026-07-22
fix(mongodb): map 4.0 dump image tag

### LINKED ISSUE #3250: MongoDB 4.0 replica dump silently uses the 8.0 image tag
## Problem

The MongoDB chart declares 4.0.28 as an active release, but the replica BackupPolicyTemplate dump method has no `4.0` entry in `IMAGE_TAG.versionMapping`. KubeBlocks therefore omits the method override and silently retains the ActionSet default `IMAGE_TAG=8.0.17`. Backup and restore jobs for a 4.0.28 replica-set dump can use the wrong MongoDB tool image.

The sharded mappings already resolve 4.0 to 4.0.28.

## Expected

- Replica dump maps serviceVersion 4.0.x to image tag 4.0.28.
- A render-level regression test derives all active releases from values.yaml and verifies MongoDB/PBM mappings for replica and sharded backup methods.

## Evidence

TDD RED: `replica dump 4.0.28 IMAGE_TAG=nil, expected "4.0.28"`.

Scope is addon source/render only; runtime validation remains separate.

weicao:
## What this PR does

- maps MongoDB replica dump serviceVersion `4.0` to the active `4.0.28` image tag
- adds a render-level contract test that derives non-deprecated releases from `values.yaml`
- verifies replica and sharded MongoDB/PBM backup image mappings for every active release
- keeps Helm dependency status output out of the YAML stream consumed by the contract test

Without the mapping, KubeBlocks omits the BackupMethod env override and silently retains the ActionSet default `IMAGE_TAG=8.0.17` for a 4.0.28 replica dump.

## Scope

- exact base: `62e3e560536092d7330a7b47159444123d952556`
- exact head: `3bf327a3220b5e940c170d5c3cba1f25b001dbf5`
- two commits, two files

## Verification

- mapping contract RED on parent: 1 example / 1 failure with `replica dump 4.0.28 IMAGE_TAG=nil`
- clean-worktree dependency-status regression RED on `d779420`: 1 example / 1 failure with `Psych::SyntaxError`
- clean-worktree dependency-status regression GREEN on exact head: 1 example / 0 failures
- focused mapping spec on Bash 3.2 and Bash 5.3: 1 example / 0 failures each
- all MongoDB ShellSpec on exact head: 17 examples / 0 failures
- `helm lint addons/mongodb`: 1 chart / 0 failures
- `git diff --check`: pass

The source/render correction changes no runtime product behavior beyond the missing 4.0 dump mapping. Runtime N remains 0 and is not claimed.

Closes #3250

### DIFF
--- a/addons/mongodb/scripts-ut-spec/backup_version_mapping_spec.sh
+++ b/addons/mongodb/scripts-ut-spec/backup_version_mapping_spec.sh
@@ -1,0 +1,83 @@
+# shellcheck shell=bash
+
+Describe "MongoDB backup version mapping contract"
+
+  render_and_validate_backup_version_mappings() {
+    local chart_dir
+
+    chart_dir=${MONGODB_CHART_DIR:-$(cd .. && pwd)}
+    helm template kb-addon-mongodb "$chart_dir" --dependency-update | ruby -ryaml -e '
+      def mapping_for(method, env_name)
+        env = Array(method["env"]).find { |item| item["name"] == env_name }
+        env&.dig("valueFrom", "versionMapping") || []
+      end
+
+      def resolve(mapping, service_version)
+        exact = mapping.find do |entry|
+          Array(entry["serviceVersions"]).include?(service_version)
+        end
+        return exact["mappedValue"] if exact
+
+        prefix = mapping.find do |entry|
+          Array(entry["serviceVersions"]).any? do |version|
+            service_version.start_with?(version)
+          end
+        end
+        prefix&.fetch("mappedValue", nil)
+      end
+
+      values = YAML.load_file(ARGV.fetch(0))
+      active = values.fetch("versions").flat_map do |group|
+        group.fetch("minors").reject { |minor| minor[4] }.map do |minor|
+          {
+            "serviceVersion" => minor[1],
+            "mongodbTag" => minor[2],
+            "pbmTag" => minor[3]
+          }
+        end
+      end
+
+      documents = YAML.load_stream($stdin.read).compact
+      policies = documents.select { |doc| doc["kind"] == "BackupPolicyTemplate" }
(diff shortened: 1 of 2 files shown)


addons/mongodb/scripts-ut-spec/backup_version_mapping_spec.sh
```diff
@@ -0,0 +1,83 @@
+# shellcheck shell=bash
+
+Describe "MongoDB backup version mapping contract"
+
+  render_and_validate_backup_version_mappings() {
+    local chart_dir
+
+    chart_dir=${MONGODB_CHART_DIR:-$(cd .. && pwd)}
+    helm template kb-addon-mongodb "$chart_dir" --dependency-update | ruby -ryaml -e '
```
leon-ape: In a clean worktree, this test fails on its first run because `helm template --dependency-update` writes repository/dependency status to stdout before the rendered manifests, and that mixed stream is passed to `YAML.load_stream`. Exact head reproduces `Psych::SyntaxError` instead of the asserted mapping result.

weicao: Fixed in 3bf327a. The spec now materializes dependencies with a separate fail-fast `helm dependency build` (stdout suppressed), then sends only `helm template` manifest stdout to Ruby. Reproduced the parent d779420 clean-worktree failure with representative dependency-status stdout: focused ShellSpec 1/1 failed with `Psych::SyntaxError`. Exact head clean first run is 1/0; focused Bash 3.2 and 5.3 are 1/0 each; all MongoDB ShellSpec is 17/0; Helm lint is 1/0.

leon-ape (approved):
LGTM

PR #3252, opened 2026-07-23
fix(tidb): preserve PITR log task on monitor failure

weicao:
## Summary

- retry transient `br log status` failures and attach to an existing PITR task before attempting `br log start`
- preserve the BR log task on abnormal monitor exit while retaining the explicit TERM/INT stop path
- fail closed before publishing malformed, inverted, or invalid-size continuous-backup status

## Evidence

A live exact-stack run observed one TSO status failure terminate the monitor. The old EXIT trap then stopped the BR log task; restart created a new task whose time range began after the requested restore time, and restore failed out of range. This change is limited to the addon monitor lifecycle and status-writing contract.

## Verification

- TDD receipt: initial focused spec 7/7 RED on base plus source guard/main split
- Bash 3.2 focused: 23/0; all TiDB shell specs: 33/0
- Bash 5.3 focused: 23/0; all TiDB shell specs: 33/0
- mutation gates: empty-default 2F, normalize fail-open 2F, range-delete 1F, size-delete 3F on both Bash versions
- `bash -n`, ShellCheck error-level, `git diff --check`
- Helm lint/template and rendered ActionSet command `bash -n`
- independent exact-byte focused review: BLOCK0

## Boundary

This PR has local source/spec/static and target-image compatibility evidence. Fresh packaged runtime validation remains separate; this PR does not claim PITR runtime PASS or address the independent KubeBlocks restore-status invariant.

### DIFF
--- a/addons/tidb/dataprotection/backup-pitr.sh
+++ b/addons/tidb/dataprotection/backup-pitr.sh
@@ -1,44 +1,169 @@
 #!/bin/bash
 
-function save_backup_status() {
-    # shellcheck disable=SC2086
-    res=$(/br log status --task-name=pitr --pd "$PD_ADDRESS" $EXTRA_ARGS)
-    start_time_str=$(echo "$res" | awk -F': ' '/^\s*start:/ {print $2}')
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

PR #3255, opened 2026-07-23
chore(mongodb): pin parameter definition identity mapping

weicao:
## Problem

PR #3207 correctly preserved the two kept ParametersDefinition identities, but its post-merge spec only compared the name set between renders. That leaves three detector gaps: a consistently wrong fixed name can become the baseline, swapping the two retained names does not bind either name to its selector, and the near-collision check changes only the first literal dot.

## Change

- pin the two historical literal object names for chart `1.2.0-alpha.0`
- pin each retained name to its exact selector in default, resource-prefix, cmpd-prefix, and combined renders
- check a one-character near-collision at every literal dot (48 candidates across four renders)
- add directed fixed-name, swapped-name, and every-dot-unquoted mutants; each example requires the validator to return failure
- build Helm dependencies once per spec, before any render output is parsed

Only `addons/mongodb/scripts-ut-spec/parameters_definition_spec.sh` changes. No production template, runtime runner, or documentation changes are included.

## Verification

- exact base/merge: `4231e045d7cf424407cf57095300f1d09d33ab31`
- exact head: `54f720cc71db47829da18156201eeea079e05264`
- targeted ShellSpec, Bash 3.2: 4 examples / 0 failures
- targeted ShellSpec, Bash 5.3: 4 examples / 0 failures
- fixed-name mutant: validator failure required, literal-name diagnostic asserted
- swapped-name mutant: validator failure required, name-to-selector diagnostic asserted
- every-dot-unquoted mutant: validator failure required, near-collision diagnostic asserted
- ShellCheck: pass
- `git diff --check`: pass
- spec SHA256: `5dd408eb5f8ec627729a92a3041f8a8fdfdbc7d19cc43c12cc2735e859e10274`

## Boundary

This is an offline/static post-merge regression-spec correction. It does not add MongoDB runtime evidence and does not change the runtime/release-readiness ledger.

### DIFF
--- a/addons/mongodb/scripts-ut-spec/parameters_definition_spec.sh
+++ b/addons/mongodb/scripts-ut-spec/parameters_definition_spec.sh
@@ -2,13 +2,19 @@
 
 Describe "MongoDB ParametersDefinition ownership contract"
 
+  prepare_chart() {
+    local chart_dir
+    chart_dir=${MONGODB_CHART_DIR:-$(cd .. && pwd)}
+    helm dependency build "$chart_dir" >/dev/null
+  }
+
   render_and_validate_parameters_definitions() {
-    local chart_dir render_dir status
+    local chart_dir mutant render_dir status
 
     chart_dir=${MONGODB_CHART_DIR:-$(cd .. && pwd)}
+    mutant=${1:-}
     render_dir=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


leon-ape (approved):
LGTM

PR #3257, opened 2026-07-23
fix(mongodb): declare snapshot target volume

### LINKED ISSUE #3256: MongoDB snapshot BPT misses custom ComponentDefinition prefix and snapshot volume declaration
## Problem

MongoDB's replica BackupPolicyTemplate declares a CSI volume-snapshot method
for target volume `data`, but the matching ComponentDefinition does not mark
that volume with `needSnapshot: true`.

The supported `cmpdVersionPrefix` chart value exposes a second contract gap:
the replica ComponentDefinition name follows the custom prefix, while the
BackupPolicyTemplate continues to match `mongodb`. Under KubeBlocks'
prefix-or-regex matching rules, that render resolves zero ComponentDefinitions.

## Evidence

A structural four-render test against current `main` fails in both Bash 3.2 and
Bash 5.3:

```text
cmpd-prefix.yaml: mongodb-backup-policy-template must resolve exactly one ComponentDefinition, got 0
```

After correcting only the effective prefix, the same test exposes the
independent volume declaration gap:

```text
volume-snapshot target volume "data" is not declared needSnapshot=true
```

This is a source/render contract defect. No historical runtime failure is
claimed.

## Expected behavior

- The replica BPT should resolve exactly one ComponentDefinition for default,
  resource-prefix, cmpd-prefix, and combined-prefix renders.
- Every `snapshotVolumes: true` target volume should be declared
  `needSnapshot: true` in the resolved ComponentDefinition.

weicao:
## What problem does this solve?

MongoDB's replica BackupPolicyTemplate has the chart's only CSI
`snapshotVolumes: true` method and targets volume `data`, but the matching
ComponentDefinition did not declare `data.needSnapshot: true`.

The supported `cmpdVersionPrefix` value also changed the replica
ComponentDefinition name without changing the BPT match pattern. That render
resolved zero ComponentDefinitions under KubeBlocks' prefix-or-regex matching
rules.

This is a source/render contract fix. It does not claim a captured runtime
failure.

## What is changed?

- Make the replica ComponentDefinition prefix helper follow the effective
  `cmpdVersionPrefix`, preserving `mongodb-` as the default.
- Render the replica BPT `compDefs` from that helper.
- Declare replica volume `data` with `needSnapshot: true`.
- Add a structural ShellSpec invariant across default, resource-prefix,
  cmpd-prefix, and combined-prefix renders.

For each render independently, the test requires exactly one snapshot method
and target, each BPT to resolve exactly one ComponentDefinition, and exact
equality between all resolved CSI snapshot targets and all rendered
`needSnapshot: true` declarations. This proves both that every target is
declared and that shard/config-server/mongos do not gain unrelated snapshot
declarations. It additionally requires both sets to equal an independent
per-render replica `/data` identity instead of deriving product scope from the
candidate BPT.

## Evidence

Current `main` with the final test:

```text
cmpd-prefix.yaml: mongodb-backup-policy-template must resolve exactly one ComponentDefinition, got 0
```

Prefix-only intermediate:

```text
volume-snapshot target volume "data" is not declared needSnapshot=true
```

Review-induced mutants:

- method distribution `0/2/1/1` across the four renders: RED per render in
  Bash 3.2 and Bash 5.3
- unrelated shard `data.needSnapshot: true`: RED exact
  target/declaration mismatch in both shells
- coordinated migration of the BPT selector and snapshot declaration from
  replica to shard: RED independent expected-identity check in both shells

Candidate:

- exact head: `051f229d7e639bf342b6b2604003aba46d88af9e`
- parent: `2b9256b411591f713b5f1937b76113cbb12ff7bb`
- focused ShellSpec: Bash 3.2 `1/0`, Bash 5.3 `1/0`
- all MongoDB ShellSpec on Bash 3.2: `19/0`
- relevant Bash 5.3 specs: `5/0`
- Helm lint: `1/0`
- ShellCheck and diff-check: clean

The all-MongoDB Bash 5.3 suite still reports a pre-existing endpoint-regex
warning with zero failed examples. It is tracked separately and is outside
this PR.

## Boundary

Runtime validation, packaging, release readiness, and merge are separate.

Closes #3256

### DIFF
--- a/addons/mongodb/scripts-ut-spec/backup_snapshot_volume_spec.sh
+++ b/addons/mongodb/scripts-ut-spec/backup_snapshot_volume_spec.sh
@@ -1,0 +1,106 @@
+# shellcheck shell=bash
+
+Describe "MongoDB backup snapshot volume contract"
+
+  render_and_validate_snapshot_volumes() {
+    local chart_dir render_dir status
+
+    chart_dir=${MONGODB_CHART_DIR:-$(cd .. && pwd)}
+    render_dir=$(mktemp -d)
+
+    helm dependency build "$chart_dir" >/dev/null || return
+    # shellcheck disable=SC2016
+    helm template kb-addon-mongodb "$chart_dir" \
+      --set cmpdVersionPrefix=static-cmpd > "$render_dir/cmpd-prefix.yaml" &&
+      helm template kb-addon-mongodb "$chart_dir" \
+        --set resourceNamePrefix=static-resource \
+        --set cmpdVersionPrefix=static-cmpd > "$render_dir/both-prefixes.yaml" &&
+      helm template kb-addon-mongodb "$chart_dir" > "$render_dir/default.yaml" &&
+      helm template kb-addon-mongodb "$chart_dir" \
+        --set resourceNamePrefix=static-resource > "$render_dir/resource-prefix.yaml" &&
+      ruby -ryaml -e '
+      def compdef_match?(name, pattern)
+        return true if name.start_with?(pattern)
+        return false unless pattern.match?(/[\\.+*?()|\[\]{}^$]/)
+
+        Regexp.new(pattern).match?(name)
+      rescue RegexpError
+        false
+      end
+
+      total_snapshot_methods = 0
+      total_snapshot_volumes = 0
+      expected_snapshot_targets = {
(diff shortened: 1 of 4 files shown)


leon-ape (approved):
LGTM

PR #3260, opened 2026-07-23
fix(mongodb): serialize sharding provisioning

weicao:
## Problem

The MongoDB sharding topology declares `config-server`, `mongos`, and `shard`, but it does not declare a provision order. The source dependency is serial: mongos needs the config server, and shards are added through mongos.

This is a source-contract gap. No live runtime failure is claimed by this PR.

## Solution

Add the exact serial provision stages:

```yaml
provision:
  - config-server
  - mongos
  - shard
```

The existing terminate and update orders are unchanged.

## Tests

- focused ShellSpec: Bash 3.2 and Bash 5.3, 3/0 each
- base plus the new spec: 3/1 each, with only the missing provision order failing
- full MongoDB ShellSpec: Bash 3.2 21/0; Bash 5.3 21/0 plus one pre-existing dataprotection regex warning
- valid duplicate `sharding` topology rejected in both shells
- missing, wrong, parallel, incomplete, terminate-drift, and update-drift mutants rejected in both shells
- valid Helm output followed by exit 42 preserves status 42
- Helm lint, ShellCheck error level, dual-shell syntax, and diff check pass

## Review focus

Please verify that the serial order matches the declared source dependency and that the focused oracle fails closed for zero or duplicate `sharding` topologies.

## Boundary

Source/render/TDD only. This PR does not claim API admission, controller execution, live `sh.addShard`, backup/restore, package, deployment, release, or runtime acceptance.

### DIFF
--- a/addons/mongodb/scripts-ut-spec/clusterdefinition_provision_order_spec.sh
+++ b/addons/mongodb/scripts-ut-spec/clusterdefinition_provision_order_spec.sh
@@ -23,8 +23,9 @@ Describe "MongoDB sharding topology provision order"
       abort "expected one MongoDB ClusterDefinition, got #{definitions.length}" unless definitions.length == 1
 
       topologies = definitions.first.dig("spec", "topologies") || []
-      sharding = topologies.find { |topology| topology["name"] == "sharding" }
-      abort "MongoDB sharding topology is missing" unless sharding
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

gnolong:
config-server depends on mongos's k8s-service.

PR #3264, opened 2026-07-24
fix(tidb): make TiKV member leave state driven

weicao:
## Problem

`tikv_member_leave.sh` deletes the target store, then waits until the store address disappears before it invokes `store remove-tombstone`. PD retains the target address while the store is `Offline` or `Tombstone`, so that ordering can wait indefinitely and hit the lifecycle-action timeout.

Contract anchors:
- `docs/addon-api/05c-lifecycle-action-contracts.md`: memberLeave must distinguish convergence states, remain retry-safe, and avoid fixed/unbounded waiting.
- `docs/addon-api/05a-lifecycle-roles-and-probe.md`: memberLeave must be repeatable or fail safely.

## Change

- Probe the target store state before mutating.
- Treat `Absent` as an idempotent success.
- Delete only an `Up` store.
- Return a classified retry-safe failure while the store is `Offline`.
- Remove a `Tombstone` record, then positively verify the target is absent.
- Fail closed on invalid input, malformed probe output, unexpected state, and rejected PD mutations.
- Add five ShellSpec cases covering absent, Up-to-Offline defer, Tombstone removal, malformed probe output, and delete rejection.

## Verification

- ShellSpec, Bash 3.2: 38 examples, 0 failures
- ShellSpec, Bash 5.3: 38 examples, 0 failures
- ShellCheck: pass
- `helm lint addons/tidb`: pass
- `helm template tidb addons/tidb`: pass

## Boundary

This PR closes the product-code and code-level-test gap only. It does not claim a runtime TiKV scale-in PASS, formal addon acceptance, cleanup success, or release readiness.

### DIFF
--- a/addons/tidb/scripts-ut-spec/tikv_member_leave_spec.sh
+++ b/addons/tidb/scripts-ut-spec/tikv_member_leave_spec.sh
@@ -1,0 +1,116 @@
+# shellcheck shell=bash
+# shellcheck disable=SC2034,SC2317,SC2329
+
+Describe "tikv member leave"
+    TIDB_SCRIPTS_DIR="../scripts"
+    __SOURCED__=true
+    Include ../scripts/tikv_member_leave.sh
+
+    setup() {
+        export KB_LEAVE_MEMBER_POD_FQDN="tidb-tikv-1.tidb-tikv-headless.default.svc"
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

PR #3266, opened 2026-07-25
fix(mongodb): require ping success before balancer toggle

weicao:
## Problem

The MongoDB toggle-balancer action treated any ping output containing `ok` as readiness, without requiring the ping command to succeed or the server result to be `ok: 1`.

This is a code-contract fix. We did not capture a live-cluster failure for this path, so this PR does not claim a reproduced runtime incident.

## Change

- Make `mongosh` exit successfully only when `db.adminCommand({ ping: 1 }).ok === 1`.
- Retry when the command fails, returns `ok: 0`, or emits unrelated output containing `ok`.
- Add focused regression cases that verify a second ping occurs before the balancer action in all three false-positive cases.

## Evidence

Using the same committed regression spec against the parent and this head:

- Parent: 10 examples, 3 failures on Bash 3.2 and Bash 5.3.
- This head: 10 examples, 0 failures on Bash 3.2 and Bash 5.3.
- The existing seven toggle-balancer cases remain green.
- The rendered action command and source/spec changes passed an independent exact-source review.

## Review focus

Please focus on:

- whether the `mongosh` exit-status expression correctly represents readiness;
- whether retry and timeout behavior remain unchanged;
- whether the three regression cases distinguish false-positive readiness from eventual success.

## Boundary

- Offline rendered-command and dual-Bash tests: passed.
- Independent source/spec review: approved with no blockers.
- CI: pending.
- Kubernetes/runtime validation: not run in this development change.
- Package, deployment, merge, and release: not claimed.

### DIFF
--- a/addons/mongodb/scripts-ut-spec/toggle_balancer_spec.sh
+++ b/addons/mongodb/scripts-ut-spec/toggle_balancer_spec.sh
@@ -59,6 +59,7 @@ Describe "MongoDB toggle-balancer operation contract"
     local observed_state="$2"
     local action_rc="$3"
     local state_rc="${4:-0}"
+    local readiness_mode="${5:-ready}"
     local temp_dir
     local script_file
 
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
LGTM

PR #3269, opened 2026-07-26
fix(falkordb): serialize rebalance config handoff

### LINKED ISSUE #3268: FalkorDB rebalance can start before its generated config is published
## Problem

The FalkorDB rebalance OpsDefinition runs the config producer and the `ape-dts` consumer as regular containers in the same Job. Kubernetes does not guarantee startup ordering between regular containers, so the consumer can start before `/tmp/reshard.ini` exists.

A fresh runtime run observed one rebalance action fail when `falkordb-rebalance` started before `busybox-init` published the shared config. Three bounded retries on the retained cluster later succeeded, so this issue is limited to the demonstrated startup race and does not claim deterministic failure.

The producer must remain a regular container because KubeBlocks injects the selected Pod metadata and credentials into `podSpec.containers` for this action.

## Expected behavior

- The producer publishes only a complete config.
- The consumer waits for that publication with a bounded timeout.
- Missing config fails with an explicit diagnostic instead of racing or waiting forever.
- The final process remains `ape-dts` so its exit status is preserved.

weicao:
## What this fixes

The rebalance config producer and `ape-dts` consumer are regular containers, so Kubernetes may start the consumer before `/tmp/reshard.ini` exists. A fresh runtime run observed that ordering once; three bounded retries on the retained cluster succeeded, so this change is scoped to the demonstrated startup race.

The producer cannot be converted to an init container because the KubeBlocks action workload injects the selected Pod metadata and credentials into regular `podSpec.containers`.

## Changes

- write the generated config to `/tmp/reshard.ini.tmp` and atomically publish it with `mv`
- make the consumer wait up to 120 seconds for a non-empty final config
- fail with an explicit diagnostic on timeout
- `exec` the original `/ape-dts` process so its exit status remains authoritative
- add a rendered structural contract test covering container placement/order, shared storage, exact image/commands, atomic publication, bounded timeout, explicit failure, and final exec ordering

## Validation

- focused ShellSpec: 1 example, 0 failures
- FalkorDB ShellSpec: 44 examples, 0 failures, 4 pre-existing Bash 3.2 skips
- `helm lint addons/falkordb`
- full FalkorDB chart render
- ShellCheck for the new spec and rendered consumer script
- independent focused re-review: NO BLOCKER

This PR is intentionally stacked on PR #3265. After #3265 merges, it can be retargeted to `main` without changing this commit.

Fixes #3268

### DIFF
--- a/addons/falkordb/scripts-ut-spec/rebalance_config_handoff_contract_spec.sh
+++ b/addons/falkordb/scripts-ut-spec/rebalance_config_handoff_contract_spec.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+# shellcheck disable=SC2016
+
+Describe "FalkorDB rebalance config handoff contract"
+  chart_path() {
+    printf '%s/addons/falkordb\n' "$(cd "${SHELLSPEC_PROJECT_ROOT:-.}" && pwd)"
+  }
+
+  validate_rendered_handoff() {
+    helm template test "$(chart_path)" \
+      --show-only templates/opsdefinition-rebalance.yaml \
+      | ruby -ryaml -e '
+          ops = YAML.load_stream($stdin.read).compact.find do |doc|
+            doc["kind"] == "OpsDefinition" &&
+              doc.dig("metadata", "name") == "falkordb-cluster-rebalance"
+          end
+          abort "rendered OpsDefinition not found" unless ops
+
+          workload = ops.dig("spec", "actions", 0, "workload")
+          abort "rebalance workload type changed" unless \
+            workload.fetch("type") == "Job"
+          abort "rebalance extractor binding changed" unless \
+            workload.fetch("podInfoExtractorName") == "falkordbUrl"
+
+          extractors = ops.dig("spec", "podInfoExtractors")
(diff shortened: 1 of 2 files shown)


weicao:
Fresh focused runtime evidence for exact head `4a7c7a81ef3b89bab285aaf49276500ae50128d9`:

- run: `pr3269-rebalance-r1-1832`
- natural terminal: runner exit 0, restart 0
- result: 42 passed, 0 failed, 0 skipped
- paths exercised: 3 -> 4 shards, custom rebalance, and 4 -> 3 shards
- each operation reached `Succeed`; slot maps were clean; key checks were base 80/80 and out 40/40; graph checks were >= 20
- evidence archive SHA-256: `32f6703a32247fa43556362ec96579d6c6a4616e25e25879ac6e2764d18196e9`
- ordinary cleanup reached NotFound with no force/finalizer edits, and the original OpsDefinition spec was restored

Evidence boundary: this is one scoped focused runtime PASS for the replacement head. It is not a whole-suite or release-readiness claim. A fresh full run has started separately and is not yet terminal.

PR #3273, opened 2026-07-28
test(mongodb): validate README relative links

weicao:
## Problem

The MongoDB addon README had three local links that resolve under `addons/docs/...`, where the referenced files do not exist:

- installation prerequisites;
- addon installation;
- BackupRepo setup.

The repository also had no MongoDB-native check for local Markdown link ownership, URL decoding, repository containment, symlink escapes, or malformed destinations. A naive checker introduced while closing that gap mixed Ruby character offsets with byte parsing: UTF-8 destinations were truncated, while converting all results to binary made `%FF.md` fall through to filesystem `missing=` instead of invalid-encoding rejection.

## Solution

- Correct the three README links to `../../examples/docs/...`.
- Add one MongoDB ShellSpec oracle covering both addon and example READMEs.
- Validate local link paths after one percent decode, reject malformed/ambiguous encodings, lexical and symlink escapes, missing files, and shared-document identity drift.
- Parse pointy and bare Markdown destinations with bounded byte coordinates, nested-parenthesis/backslash parity, malformed-segment recovery, and explicit cursor invariants.
- Preserve source encoding on parser results so UTF-8 semantics remain intact after binary scanning.

No addon runtime scripts, charts, or generated package content change.

## Evidence

Observed on the pre-change tree:

- the three original README targets resolve to missing `addons/docs/...` paths;
- UTF-8 destination `[Unicode](你好.md)` was truncated to malformed `"你好."`;
- a UTF-8 prefix shifted the target to malformed `"de](local."`;
- binary scanning without result-encoding restoration treated decoded `0xff` as a filesystem path instead of malformed encoding.

Final exact-head tests:

- focused relative-link spec: 185 examples, 0 failures on Bash 3.2.57 and Bash 5.3.9;
- full MongoDB ShellSpec directory: 263 examples, 0 failures on both shells;
- negative controls independently kill binary-without-encoding-restoration and bytesize-without-binary-scanning;
- retained parser mutants fail with exact ownership: previous-byte 1, next-opener 1, capped parity 1, escaped-close 2, inverted bare parity 35;
- Bash syntax, ShellSpec syntax, ShellCheck, diff-check, and strict Git fsck pass.

## Review focus

Please focus on:

1. whether the three corrected README targets are the canonical shared docs;
2. the local/external URL ownership and one-decode ordering;
3. pointy/bare malformed recovery and cursor bounds;
4. the binary-coordinate/source-encoding boundary in `markdown_link_targets`.

## Boundary

This is repository-native source/document validation only. Package, Kubernetes, backup/restore runtime, release, and accepted runtime counts are not claimed.

### DIFF
--- a/addons/mongodb/README.md
+++ b/addons/mongodb/README.md
@@ -33,8 +33,8 @@ MongoDB is a document database designed for ease of application development and 
 - Kubernetes cluster >= v1.21
 - `kubectl` installed, refer to [K8s Install Tools](https://kubernetes.io/docs/tasks/tools/)
 - Helm, refer to [Installing Helm](https://helm.sh/docs/intro/install/)
-- KubeBlocks installed and running, refer to [Install Kubeblocks](../docs/prerequisites.md)
-- MongoDB Addon Enabled, refer to [Install Addons](../docs/install-addon.md)
+- KubeBlocks installed and running, refer to [Install Kubeblocks](../../examples/docs/prerequisites.md)
+- MongoDB Addon Enabled, refer to [Install Addons](../../examples/docs/install-addon.md)
 - Create K8s Namespace `demo`, to keep resources created in this tutorial isolated:
 
   ```bash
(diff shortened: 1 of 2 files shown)


leon-ape (commented):
`readme_relative_link_spec.sh` does not implement Markdown link semantics: reference-style links are not scanned, links inside fenced code are treated as real links, and valid inline links with optional titles are reported as malformed. The check can therefore pass broken local links and fail valid README content.

PR #3229, opened 2026-07-16
fix(mariadb): make switchover completion non-preemptive

weicao:
## Problem and observed evidence

This stacked PR now closes two addon-side parts of the MariaDB switchover contract.

### 1. Explicit candidates can converge after the action budget

Two independent runs showed the same shape:

- MariaDB 11.4.5: DCS request and old-primary fence closed, but the named candidate converged about 162 seconds later. The action exhausted its budget and returned `rc=1` first.
- MariaDB 11.4.8: the candidate converged about 141 seconds later, again after the action had returned `rc=1`.

Evidence strength is 11.4.5 N=1 plus 11.4.8 N=1. It is not a frequency claim.

### 2. A second request must not preempt an unfinished DCS switchover

In the fresh r7 failure, syncer prepared and promoted the candidate at 07:20:58, but the addon saw `secondary` for 28 polls and Stage 5 returned `rc=1`. The first authoritative lease refresh appeared at 07:25:45 and the database write commit at 07:25:48.

The old addon always passed `syncerctl switchover --force`. That allowed a later request to delete an unfinished marker without proving that the previous lease, database, or grant changes had rolled back.

This evidence proves the circular completion timing at the old exact. It does **not** prove the new addon exact or the companion syncer patch at runtime.

### Candidate-mode completion

- Explicit candidate: after DCS acceptance and a positive old-primary fence, return structured `convergence=delegated`. The exact KubeBlocks controller keeps the OpsRequest Processing until that named candidate reports the expected role.
- Auto-selected candidate: retain Stage 5/6 candidate-role and remote-root readiness proof, then return `convergence=observed`. An empty CandidateName has no controller-side named-candidate role gate.
- Unknown candidate mode fails closed.

The internal action deadline remains 50 seconds under the 60-second ComponentDefinition timeout.

### Non-preemptive in-flight handling

- Remove `--force` from both real `syncerctl switchover` invocations.
- Map syncer's stable `SWITCHOVER_IN_PROGRESS` response to `outcome=retryable_conflict mutation=0`.
- Return immediately on that proven conflict, before duplicate final-state probing or uncertain-DCS local fencing.
- If DCS was accepted but Stage 5/6 cannot prove terminal truth, report `outcome=indeterminate_dcs_in_progress mutation=remote-partial`; keep the old primary fenced.
- Keep the existing uncertain-DCS fence for all other non-zero/timeout responses, where persistence cannot be proven either way.

## TDD and independent review

[...]

PR #3215, opened 2026-07-15
mariadb: align primary roleProbe marker contract

weicao:
## Problem

Fresh diagnostic r12 on addon `3e51be9bebd950b26405399966243fba64d5596b` and syncer `7502b58f45efd4fe8f2b06e5313230d9aec69510` narrowed the post-commit repair loop to one directly conflicting predicate:

```text
2026-07-15T04:50:15Z primary-write-commit rc=0 authority=fresh-lease-cas
2026-07-15T04:50:18Z runtime-primary-listener-reconcile-repair-begin primary_ready=present remote_root_fence=present master_info=absent listener_wildcard=present
```

The entrypoint clears `.remote-root-fence-role` before authoritative commit and treats marker absence as part of the committed-primary invariant. The roleProbe then calls `apply_remote_root_fence "primary"` and writes `primary` back to that marker, causing the runtime repair path to retract readiness and fence the database.

Canonical evidence:

- outer SHA256: `f953215c36658cf9564294f8842b3dc4df915aaec60fca65856efcbccc4758a2`
- inner `SHA256SUMS` SHA256: `9bd27dc268c3a5338cfefdbb08f030f40e3e047b337b8057549ac064f57a56fd`
- 22/22 payload files verified; scene remains frozen, repair/cleanup 0

This is exact-source N=1 evidence. It proves the first observed conflicting predicate for this run, not fleet frequency.

## Change

Give `.remote-root-fence-role` one durable meaning:

- `secondary` present: secondary remote-root fence is active
- absent: committed primary remote-root state

An already-accepted primary with no marker no longer rewrites grants or recreates the marker. A secondary-to-primary transition still applies the primary grant set, then removes the marker. If that transition fails, it preserves the secondary marker so the entrypoint observes and repairs the drift instead of hiding it. Secondary behavior is otherwise unchanged.

## Predecessor TDD

Production RED on the old source: `50 examples, 3 failures, 1 historical pending`:

1. accepted primary with absent marker failed because roleProbe tried to recreate primary state;
2. successful secondary-to-primary transition left a `primary` marker;
3. failed transition erased the existing `secondary` fence marker.

GREEN:

- focused roleProbe: `50 examples, 0 failures, 1 historical pending`
- related roleProbe + merged recovery + semisync fence: `153 examples, 0 failures, 1 historical pending`
- full MariaDB ShellSpec: `770 examples, 0 failures, 9 historical pendings`
- POSIX `sh -n`
- ShellCheck error-level
[...]

PR #3214, opened 2026-07-15
mariadb: log primary reconcile drift predicates

weicao:
## Problem

Fresh focused r11 on addon combined `73d05dcfd2a907b7bbcb292d2f0d8d3de7d7abf1` and syncer `7502b58f45efd4fe8f2b06e5313230d9aec69510` proved an addon-side marker oscillation. After an authoritative primary write commit succeeds, the runtime primary listener reconciler enters `primary-role-state-drift` and starts repair before the next HA promotion. Repair retracts the authority markers, but the current aggregate log does not identify which of its four predicates first failed.

Evidence:

- r11 first-red outer SHA256: `5ca8e2cc2a71c74d177ef3b8c4e0dd2633bab90e75a3320964494130ce6f579a`
- r11 first-red manifest SHA256: `cf3d4b219a8ad4416b9133b3a615f787a86d30389677bf88fa68c12bd83081fc` (14/14)
- supplemental addon-log outer SHA256: `83aabd76e9612f9291bdefae358f75e2bc4622a5b35955f69ed00299eee69096`
- supplemental manifest SHA256: `234e06383a2362f02bbd2a5fff82cab5a0c27177ec82dd10d7d5aae345a5ee60` (3/3)
- observed sequence: authority commit/reconcile complete at `03:22:20Z`; addon `runtime-primary-listener-reconcile-repair-begin` at `03:22:24Z`; only afterwards does the repair fail closed.

## Change

Capture and log the four primary-publish predicates before repair can mutate state:

- `primary_ready`
- `remote_root_fence`
- `master_info`
- `listener_wildcard`

The fast-path decision uses that same captured snapshot. This is diagnostic observability only; it does not change the repair policy or claim to fix the oscillation.

## TDD and source gates

- RED: focused ShellSpec `30 examples, 1 failure` before production instrumentation
- GREEN: focused ShellSpec `30 examples, 0 failures`
- full MariaDB ShellSpec: `768 examples, 0 failures, 9 historical pendings`
- `bash -n addons/mariadb/scripts/replication-entrypoint.sh`
- `shellcheck -S error addons/mariadb/scripts/replication-entrypoint.sh`
- `helm lint addons/mariadb`
- default `helm template` plus rendered predicate-log verification
- `git diff --check`

## Runtime boundary

The next focused round is diagnostic N=1 only. Its acceptance target is an unambiguous pre-mutation value for all four predicates across the commit-to-repair window. It is not release evidence, and the full suite remains NO-LAUNCH.

PR #3276, opened 2026-07-28
fix(mariadb): harden HA safety, Galera recovery, and configuration

weicao:
## Summary

This PR is the single integration point for the current MariaDB addon candidate. It consolidates 15 previously stacked draft PRs into one reviewable and testable head.

The combined change hardens four production areas:

1. replication failover must not expose two writable primaries or publish readiness without current HA authority;
2. Galera must shut down, restart, bootstrap, and recover stalled joiners without corrupting cluster safety;
3. reconfigure and chart inputs must select valid MariaDB behavior instead of MySQL-only, read-only, or ambiguous paths;
4. credentials and dead privileged scripts must not persist into snapshots or remain wired into the chart.

## Problems solved

| Area | Previous problem and risk | What this PR changes |
| --- | --- | --- |
| Replication switchover fail-closed behavior | A timeout/non-zero return from the DCS request did not prove that the request had not been persisted. The old rollback reopened the former primary, so an asynchronously promoted candidate could leave two writable primaries. | Once DCS persistence is uncertain, keep the old primary fenced. Distinguish a proven in-flight conflict, uncertain persistence, and accepted-but-not-yet-observed convergence instead of rolling back to writable. |
| Primary write authority | Promotion, user-root grants, global writes, readiness markers, role reconciliation, and HA lease changes were not one linearized commit. A secondary reconcile could re-fence the promoted candidate, or a stale role observation could reopen writes after authority had changed. | Keep the full addon acceptance attempt under one local lock, gate ordinary writers before publication, use fenced promotion for non-bootstrap callers, and make the authoritative syncer commit the last visible write/readiness operation. preStop shares the same lock and performs no mutation when it cannot acquire it. |
| Switchover retry and completion | Every switchover used `--force`, so a second request could preempt an unfinished request. Explicit candidates can legitimately become authoritative after the addon's bounded action budget, causing false action failure even though controller-side convergence can still complete. | Remove forced preemption, map `SWITCHOVER_IN_PROGRESS` to a mutation-free retryable conflict, preserve the fail-closed fence for indeterminate outcomes, delegate explicit-candidate convergence after a positive old-primary fence, and retain direct observation for auto-selected candidates. |
| Durable role/fence markers | roleProbe recreated a `primary` fence marker after the entrypoint had committed primary state, while generic pending logic deleted a healthy secondary's durable fence marker. Both created marker oscillation and repeated repair/fencing. | Give the marker one meaning: `secondary` means the secondary fence is active; absence means committed primary state. Generic readiness transitions no longer rewrite durable fence ownership. |
| Primary drift diagnosis | When the runtime reconciler entered repair after a successful authority commit, one aggregate log did not show which prerequisite had drifted. | Log the captured pre-mutation values of `primary_ready`, `remote_root_fence`, `master_info`, and `listener_wildcard`, and use the same snapshot for the fast-path decision. This part is observability only. |
| Galera hard-crash bootstrap | `seqno=-1` was treated as sufficient for pod-0 to bootstrap. A brand-new PVC and an unclean hard crash both use `seqno=-1`; bootstrapping the latter can select stale data and overwrite newer transactions through SST. | Require `safe_to_bootstrap=1` for automatic bootstrap. `seqno=-1 + safe_to_bootstrap=0` now defers through the fail-closed recovery path. Peer-alive and persisted role/synced evidence are hardened so query errors or stale files do not authorize bootstrap. |
| Galera shutdown ordering | The native Kubernetes preStop hook ran SQL and `mysqladmin shutdown` before kubelet could signal PID 1, consuming the shared grace period. It also waited forever on a deleted higher-ordinal peer because authoritative DNS absence was treated as uncertainty. | Keep preStop limited to bounded high-to-low ordering and shutdown-marker publication, then return so kubelet terminates PID 1. Confirm authoritative peer absence in two consecutive polls, keep transient failures fail-closed, and persist degraded timeout evidence. |
| Galera stalled joiner recovery | A mariadbd process could remain socket-ready in `Primary / Joining` with no SST marker/helper and no progress, while the old watcher had no exact predicate for this orphaned state. | Add an evidence-keyed watcher: require a complete eight-field Galera receipt, stable PID+starttime identity and progress token for the configured threshold (default 90s), revalidate at action time, then make at most one TERM→bounded KILL attempt for that exact process. SST, graceful shutdown, incomplete receipts, identity changes, or token progress cancel the mutation. |
| Galera memberLeave | The action ran in kbagent, which has no MariaDB client, swallowed its SQL failure, and therefore looked successful while doing nothing. | Remove the action and script. Galera membership eviction remains automatic; graceful leave is owned by the in-container preStop path. |
| Galera SST credential storage | An unused plaintext root credential was written below the snapshot-enabled data directory and therefore copied into volume snapshots/backups. | Stop writing `wsrep_sst_auth`, remove its defaults file from startup, and scrub the stale file on upgrade. Historical snapshots still require independent credential rotation. |
| Reconfigure against TLS-disabled servers | A newer MariaDB client could require TLS while the local server had TLS disabled, producing `ERROR 2026` before `SET GLOBAL`. | For loopback internal-admin reconfigure only, add `--skip-ssl` when TLS is disabled/empty/unset; preserve the existing client behavior when `TLS_ENABLED=true`. |
| MariaDB parameter semantics | `log_error_verbosity` is MySQL-only and can crash MariaDB on restart; `secure_timestamp` is read-only but was classified dynamic, so runtime `SET GLOBAL` failed. | Remove `log_error_verbosity` and classify `secure_timestamp` as static so changes use the restart path. |
| Cluster chart topology | `mode` only changed replica count while `componentDef: mariadb` ambiguously matched standalone, replication, and Galera definitions. The schema also allowed only standalone/one replica. | Use `clusterDef: mariadb` plus `topology: standalone|replication|galera`; expand the schema and replica guidance so the selected mode resolves the intended component definition. |
| Dead reloader path | An unused root-authenticated reloader ConfigMap/script was still created and mounted even though no production path invoked it. | Remove the dead ConfigMap, mount, script, and obsolete tests. The live inline reconfigure action remains unchanged. |
| SQL literal safety | Replication helpers escaped single quotes but not backslashes under MariaDB's default SQL mode, so a credential ending in `\` could escape the closing quote. | Escape backslashes before single quotes in the three replication SQL helpers. |

## Consolidated source

The old 15 drafts were closed as superseded only after this PR was confirmed open and mergeable. Their current heads are all ancestors of this exact candidate:

- #3229 — non-preemptive switchover and candidate-mode completion
- #3228 — explicit loopback TLS mode for reconfigure
- #3217 — ordered, kubelet-driven, DNS-safe Galera preStop
- #3216 — preserve the secondary fence marker while pending
- #3215 — align the primary roleProbe marker contract
- #3214 — log primary reconcile drift predicates
- #3208 — linearize primary write acceptance with HA authority
- #3144 — fail-closed hard-crash bootstrap and backslash-safe SQL quoting
- #3108 — bounded Galera ordered shutdown
- #3070 — correct MariaDB parameter effect scope
- #3068 — deterministic standalone/replication/Galera chart topology
- #3066 — fail closed after uncertain DCS switchover
- #3057 — stop persisting the Galera SST root credential
- #3055 — remove the silent-no-op Galera memberLeave action
- #3034 — remove the dead reloader path

The branch-tip change at `7978812cf8ca066cfc13ac6baf0ddd78505e0965` adds the exact stalled-Joining recovery described above.

## Explicit exclusion

- #3117 `4280ec407a05f7cb2e3424e5f3ee70584bac51ca` is **not included**. It is a broad MySQL/cross-addon compatibility PR; its only MariaDB-path delta changes the chart's minimum KubeBlocks annotation from `>=1.0.0` to `>=1.2.0`. It was not one of the 15 MariaDB product-fix PRs and is not part of the tested candidate. It remains open and independent.

## Frozen candidate

| Field | Exact value |
| --- | --- |
| Target branch | `main` |
| Target head at freeze | `8e745a45e2540b1fc36d2d73cfbedda63c3a591f` |
| Development branch | `helen/mariadb-r26-galera-orphan-joining` |
| Candidate commit | `7978812cf8ca066cfc13ac6baf0ddd78505e0965` |
| Candidate tree | `22226af11fe44b0bfe8fd3119ce000b89b3a6035` |
| Candidate parent | `36da5dad10dc4262cf7a55fcfe7b88d97799c806` |
| Addon path | `addons/mariadb` |
| Chart / app | `1.2.0-alpha.34` / `11.4.10` |
| KubeBlocks + DataProtection candidate | `5a57a85922ef7fb8f704cc1d51f656a62f771485` |
| init-syncer candidate | PR #313, `456c2bc074365e4d96dcee9562fcba850c62143c` (`0.7.7 + PR #313`) |
| Last Dev-correlated tests source pin | `b94f79e08850ca0d70d5024af7393a0409774ca3` |
| Test-owned generated runner source | Must be freshly pinned by Test when packaging this PR; the last R16 v5 generated package was not published as a canonical Git commit |
| vcluster exact | Must be freshly read and recorded by Test at the launch gate |

The candidate is 21 commits behind and 52 commits ahead of the target head at this freeze. Rebasing, merging `main`, or adding any commit changes the candidate SHA and requires a new package identity and fresh acceptance.

## Validation status and boundaries

Fresh offline gates on the exact candidate:

- clean worktree
- `git diff --check`
- `helm lint addons/mariadb`
- conflict-free synthetic merge with the frozen target via `git merge-tree --write-tree`
- all 15 included PR current heads verified as candidate ancestors

Latest full product-body signal on the same addon commit:

- Job: `mariadb-r16-v5-full-r3-7978812c`
- result body: 310 PASS / 0 FAIL / 15 SKIP
- four controller task #204 invalid-value Reconfigure cases were skipped by the agreed known-issue rule

This is **not** a release-ready claim:

- ordinary cleanup for the latest full body has not closed, so the formal ledger remains N8 / PASS7 / FAIL1;
- controller task #204 remains an excluded control-plane known issue, not fixed by this PR;
- existing snapshots created before the SST credential fix may still contain the old credential;
- Test must package this PR as a fresh candidate, record exact tests/vcluster/runtime image identities, run the complete MariaDB suite, stop at the first non-environment non-SKIP failure, and close ordinary cleanup before counting the round;
- any head SHA change invalidates acceptance for this exact PR head.

Keep the PR in draft until the exact head passes the required repeated full-suite acceptance and a human approves integration.


weicao:
Closing this consolidation PR per owner direction. The original 15 MariaDB PRs are being reopened as the review and validation units; no candidate code commit is changed by this metadata rollback.

PR #3282, opened 2026-07-29
feat(mongodb): adapt shard lifecycle to syncer APIs

gnolong:
## Summary

- invoke Syncer's asynchronous `add-shard` API from the MongoDB shard `postProvision` lifecycle
- invoke Syncer's asynchronous `remove-shard` API from the sharding `shardRemove` lifecycle
- replace credential-bearing mongosh orchestration with a thin script that only polls `syncerctl`
- use explicit 10-minute add / 1-hour remove bounds and fail fast for permanent conditions such as jumbo chunks
- set lifecycle `timeoutSeconds: -1` because draining can exceed KubeBlocks' default action timeout

KubeBlocks 1.1.0-beta.8 action retries use coarse/exponential backoff and terminated a real remove operation before Syncer reached `Succeeded`; keeping the bounded polling loop inside the lifecycle action avoids that mismatch while preserving Syncer as the sole shard-operation owner.

## Dependency

Requires the Syncer shard operation implementation from `apecloud/syncer` branch `feature/mongodb-shard-operations` (commit `0d4de47` or a descendant) in the configured Syncer image.

## Validation

- `shellspec ... shard_manage_spec.sh`: 3 examples, 0 failures
- `helm lint addons/mongodb`
- `helm template mongodb addons/mongodb --namespace kb-system`
- `bash -n addons/mongodb/scripts/mongodb-shard-manage.sh`
- `git diff --check`
- live MongoDB 8.0.17 E2E: PASS in 466.37s with KubeBlocks/DataProtection 1.1.0-beta.8
  - sharding scale `1 -> 2 -> 1`
  - database primary and records placed on the actual scale-in victim
  - victim Component and Pod deleted; one shard Component remained
  - data and database primary migrated to the survivor
  - balancer disabled before removal and verified disabled afterward
  - Syncer `0d4de47`, companion Syncer-test `8f1a3e7`

Companion test: https://github.com/apecloud/syncer-test/pull/3

### DIFF
--- a/addons/mongodb/scripts-ut-spec/shard_manage_spec.sh
+++ b/addons/mongodb/scripts-ut-spec/shard_manage_spec.sh
@@ -1,0 +1,79 @@
+# shellcheck shell=bash
+
+Describe "MongoDB shard lifecycle polling"
+  setup() {
+    TEST_DIR=$(mktemp -d)
+    export TEST_DIR
+    export SYNCERCTL_BIN="$TEST_DIR/syncerctl"
+    export SYNCER_SHARD_POLL_INTERVAL_SECONDS=1
+    export SYNCER_ADD_SHARD_TIMEOUT_SECONDS=10
+    export SYNCER_REMOVE_SHARD_TIMEOUT_SECONDS=10
+  }
+  BeforeEach 'setup'
+
+  cleanup() {
+    rm -rf "$TEST_DIR"
+    unset TEST_DIR SYNCERCTL_BIN SYNCER_SHARD_POLL_INTERVAL_SECONDS
+    unset SYNCER_ADD_SHARD_TIMEOUT_SECONDS SYNCER_REMOVE_SHARD_TIMEOUT_SECONDS
+  }
+  AfterEach 'cleanup'
+
+  It "polls Running until add-shard succeeds"
+    cat > "$SYNCERCTL_BIN" <<'MOCK'
+#!/bin/bash
+count_file="$TEST_DIR/count"
+count=0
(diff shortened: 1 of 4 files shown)


weicao (commented):
I found four release-blocking issues:

- **Default Syncer image lacks the invoked shard commands.** The chart still renders `apecloud/syncer:0.7.7`, whose `syncerctl` exposes neither `add-shard` nor `remove-shard`, while this change invokes both. Default shard post-provision and shard removal therefore fail before the new lifecycle behavior can run, leaving creation and scale workflows stuck.

- **Released KubeBlocks does not enforce the declared polling budgets.** The new flow relies on `maxRetries` and `retryInterval`, but the supported released Core baselines serialize and ignore `ActionRequest.RetryPolicy`. A Running or persistently failing Syncer operation is governed by controller reconciliation rather than these bounds and can churn or block creation and scale workflows indefinitely.

- **Shard removal no longer restores an initially disabled balancer.** The declared Syncer implementation unconditionally runs `balancerStart` and never restores the prior state, whereas the deleted script did. Automatic migrations and MongoDB 7+ AutoMerger can remain enabled after removal and interfere with administrator-selected maintenance or manual-backup windows.

- **The destructive cross-repository workflow lacks exact-head runtime evidence.** The available checks do not demonstrate a live 1→2→1 shard cycle with data movement, database-primary movement, primary changes, Syncer restart/resume, or the materially different supported MongoDB versions. Scale-in can therefore merge without evidence that it preserves data and resumes safely on the real runtime path.

PR #3286, opened 2026-07-31
chore: remove cmpd serviceVersion release 1.0

JashBook:


leon-ape (commented):
Removing `serviceVersion: 8.0.17` from `milvus-minio-1.0.2` leaves that definition absent from every resolver candidate set: the Milvus `ComponentVersion` has no MinIO compatibility rule or release, and the default standalone topology supplies no MinIO service version. On release-1.0, creating the default Milvus topology therefore fails normalization with no matching component definition, blocking new standalone clusters.

The patch updates the same 99 ComponentDefinition identities in place while removing a field covered by release-1.0's immutable hash, but only 34 rendered objects carry `apps.kubeblocks.io/skip-immutable-check: true`. Upgrading any of the other 65 objects makes its ComponentDefinition `Unavailable` and stalls Components that consume it, breaking existing installations across the affected charts.

JashBook:
> Removing `serviceVersion: 8.0.17` from `milvus-minio-1.0.2` leaves that definition absent from every resolver candidate set: the Milvus `ComponentVersion` has no MinIO compatibility rule or release, and the default standalone topology supplies no MinIO service version. On release-1.0, creating the default Milvus topology therefore fails normalization with no matching component definition, blocking new standalone clusters.
> 
> The patch updates the same 99 ComponentDefinition identities in place while removing a field covered by release-1.0's immutable hash, but only 34 rendered objects carry `apps.kubeblocks.io/skip-immutable-check: true`. Upgrading any of the other 65 objects makes its ComponentDefinition `Unavailable` and stalls Components that consume it, breaking existing installations across the affected charts.

Thanks for the detailed review. Both issues are valid — here's how we're handling them:
Milvus MinIO — no version source after serviceVersion removal (both release-1.0 and release-1.1)
You're right. Removing serviceVersion from milvus-minio leaves it with no version source at all: the Milvus ComponentVersion has no MinIO compatibility entry, and the default standalone topology doesn't supply one either. On either branch, this would cause new standalone Milvus clusters to fail normalization with an empty candidate set for the MinIO component.
I'll revert the serviceVersion removal for milvus-minio on both branches. Removing it safely requires a separate change to add proper MinIO version coverage in the Milvus ComponentVersion first, which is out of scope for this PR.
Immutable hash / in-place upgrade breakage (65 ComponentDefinitions)
All affected addons will get a chart patch version bump on release-1.0. Since ComponentDefinition names embed the chart version (e.g. mysql-8.0-<chartVersion>), an upgrade creates new-named definitions instead of updating existing ones in place. The old definitions are never mutated, so the immutable hash check is never triggered and existing Components remain bound to their original definitions without disruption. This follows the repo's standard addon release pattern.
On release-1.1 (and main), neither branch has been formally released, so there are no deployed installations with these definitions to break. The in-place change carries no backward-compatibility impact.

weicao:
Exact-head review closure: `cab8bcc4a905d0b6dff4bce0fa335e8daaf2a73c`, rebased onto `release-1.0@548186abe12b7eefa05b36c0a7bc19004e1ca7d1` (mergeable, no base conflict).

- `milvus-minio-1.0.2` still renders `spec.serviceVersion: 8.0.17`; its version-source blocker remains closed.
- Restored `serviceVersion` on all 65 same-identity ComponentDefinitions that do not bypass KubeBlocks' immutable check. This follows the release-1.0 upgrade contract in `kubeblocks-addon-docs/docs/addon-cmpd-immutable-fields-version-bump-guide.md` (same-name immutable spec changes are not a release path).
- The final diff removes `serviceVersion` only from 34 rendered definitions. A base-vs-head render audit proved, for every one: same metadata.name, the only spec delta is deleting `serviceVersion`, and `apps.kubeblocks.io/skip-immutable-check: "true"` is present.
- Local offline gates: dependency build + Helm lint + Helm render PASS for clickhouse, elasticsearch, mongodb, and redis; `git diff --check` PASS.

weicao:

leon-ape (approved):
LGTM

PR #3287, opened 2026-07-31
chore: remove cmpd serviceVersion release 1.1

JashBook:


leon-ape (commented):
Removing `serviceVersion` leaves `milvus-minio-1.1.0-alpha.0` with no version source: no compatible `ComponentVersion` release covers it, and the default standalone topology does not supply a version. KubeBlocks therefore builds an empty candidate set during normalization, so a default standalone Milvus cluster fails before its MinIO component can be created.

These deletions change the immutable spec of 65 existing same-name `ComponentDefinition` objects that do not opt out of KubeBlocks' immutable check. On an addon upgrade, KubeBlocks detects the changed hash, marks each definition `Unavailable`, and consuming `Component` reconciliation rejects it, so otherwise routine chart upgrades can make the affected components unavailable.

JashBook:
> Removing `serviceVersion` leaves `milvus-minio-1.1.0-alpha.0` with no version source: no compatible `ComponentVersion` release covers it, and the default standalone topology does not supply a version. KubeBlocks therefore builds an empty candidate set during normalization, so a default standalone Milvus cluster fails before its MinIO component can be created.
> 
> These deletions change the immutable spec of 65 existing same-name `ComponentDefinition` objects that do not opt out of KubeBlocks' immutable check. On an addon upgrade, KubeBlocks detects the changed hash, marks each definition `Unavailable`, and consuming `Component` reconciliation rejects it, so otherwise routine chart upgrades can make the affected components unavailable.

Thanks for the detailed review. Both issues are valid — here's how we're handling them:
Milvus MinIO — no version source after serviceVersion removal (both release-1.0 and release-1.1)
You're right. Removing serviceVersion from milvus-minio leaves it with no version source at all: the Milvus ComponentVersion has no MinIO compatibility entry, and the default standalone topology doesn't supply one either. On either branch, this would cause new standalone Milvus clusters to fail normalization with an empty candidate set for the MinIO component.
I'll revert the serviceVersion removal for milvus-minio on both branches. Removing it safely requires a separate change to add proper MinIO version coverage in the Milvus ComponentVersion first, which is out of scope for this PR.
Immutable hash / in-place upgrade breakage (65 ComponentDefinitions)
All affected addons will get a chart patch version bump on release-1.0. Since ComponentDefinition names embed the chart version (e.g. mysql-8.0-<chartVersion>), an upgrade creates new-named definitions instead of updating existing ones in place. The old definitions are never mutated, so the immutable hash check is never triggered and existing Components remain bound to their original definitions without disruption. This follows the repo's standard addon release pattern.
On release-1.1 (and main), neither branch has been formally released, so there are no deployed installations with these definitions to break. The in-place change carries no backward-compatibility impact.

weicao:
Exact-head review closure: `e2c5b25a3644bd6f5c7a89d362179ab573c9a382`, rebased onto `release-1.1@91c339ead21398fa720dcbb24be091d9ab155ef4` (mergeable, no base conflict).

- `milvus-minio-1.1.0-alpha.0` still renders `spec.serviceVersion: 8.0.17`; its version-source blocker remains closed.
- Restored `serviceVersion` on all 65 same-identity ComponentDefinitions that do not bypass the immutable check. The final diff removes it only from 33 rendered definitions; a base-vs-head render audit proved each has the same metadata.name, only the `serviceVersion` spec delta, and `apps.kubeblocks.io/skip-immutable-check: "true"`.
- Added retained `rocketmq-exporter-1.0.3` to the `rocketmq-exporter-0.0.3` compatibility rule. The rendered rule now covers both `rocketmq-exporter-1.0.3` and `rocketmq-exporter-1.1.0-alpha.0`, matching `kubeblocks-addon-docs/docs/addon-api-1.1/04-component-version.md` lines 16 and 44.
- Added focused ShellSpec coverage: Bash 5 exact local run 1 example / 0 failures; ShellCheck PASS.
- Local offline dependency build + Helm lint + Helm render PASS for clickhouse, elasticsearch, falkordb, mongodb, redis, and rocketmq; `git diff --check` PASS.

weicao:

leon-ape (approved):
LGTM

PR #3288, opened 2026-07-31
chore: remove cmpd serviceVersion

JashBook:
fix https://github.com/apecloud/kubeblocks/issues/10730

### DIFF
--- a/addons/apecloud-mysql/templates/cmpd-wescale-controller.yaml
+++ b/addons/apecloud-mysql/templates/cmpd-wescale-controller.yaml
@@ -10,7 +10,6 @@ spec:
   provider: kubeblocks.io
   description: wescale controller
   serviceKind: wescale-ctrl
-  serviceVersion: "0.2.7"
   serviceRefDeclarations:
     - name: etcd
       serviceRefDeclarationSpecs:
--- a/addons/apecloud-mysql/templates/cmpd-wescale-controller.yaml
+++ b/addons/apecloud-mysql/templates/cmpd-wescale-controller.yaml
@@ -10,6 +10,7 @@ spec:
   provider: kubeblocks.io
   description: wescale controller
   serviceKind: wescale-ctrl
+  serviceVersion: "0.2.7"
   serviceRefDeclarations:
     - name: etcd
       serviceRefDeclarationSpecs:
--- a/addons/apecloud-mysql/templates/cmpd-wescale.yaml
+++ b/addons/apecloud-mysql/templates/cmpd-wescale.yaml
@@ -10,6 +10,7 @@ spec:
   provider: kubeblocks.io
   description: wescale
   serviceKind: wescale
+  serviceVersion: "0.2.7"
   serviceRefDeclarations:
     - name: etcd
       serviceRefDeclarationSpecs:
--- a/addons/apecloud-mysql/templates/cmpd-wescale.yaml
+++ b/addons/apecloud-mysql/templates/cmpd-wescale.yaml
@@ -10,7 +10,6 @@ spec:
   provider: kubeblocks.io
   description: wescale
   serviceKind: wescale
-  serviceVersion: "0.2.7"
   serviceRefDeclarations:
     - name: etcd
       serviceRefDeclarationSpecs:
--- a/addons/clickhouse/templates/cmpd-ch.yaml
+++ b/addons/clickhouse/templates/cmpd-ch.yaml
@@ -11,7 +11,6 @@ spec:
   provider: ApeCloud
   description: {{ .Chart.Description }}
   serviceKind: {{ .Chart.Name }}
-  serviceVersion: {{ .Chart.AppVersion }}
   podManagementPolicy: Parallel
   minReadySeconds: 10
   runtime:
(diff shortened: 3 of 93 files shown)


leon-ape (commented):
Removing `serviceVersion` changes the immutable spec of 62 rendered `ComponentDefinition` objects whose annotations do not bypass the immutable check, while their names remain unchanged. Upgrading an installation that already contains any of these definitions makes the ComponentDefinition controller reject the new hash and mark it `Unavailable`, blocking affected Components and cluster operations.

The linked 1.0→1.1 failure uses the retained `rocketmq-exporter-1.0.3` definition, but this branch only removes `serviceVersion` from the new 1.2 definition and leaves `ComponentVersion` compatibility restricted to the current definition identity. The retained definition therefore still has no compatible 0.0.3 release, so upgraded clusters continue to fail normalization and subsequent operations with `no matched component definition`.

khalildai:
Fresh exact-head CI evidence for `903561fb30f0c17fe2581c664426bdc9076d3121` shows one deterministic ShellSpec regression:

- run [30625369685](https://github.com/apecloud/kubeblocks-addons/actions/runs/30625369685): **2,166 examples, 1 failure, 9 pending**
- failing contract: `addons/mysql/scripts-ut-spec/mgr_version_support_spec.sh:79`, “uses 8.0.45 as the MGR 8.0 default”
- the fixture/parser now raises `KeyError: key not found: "serviceVersion"`, so `assert_mgr_default_is_supported` returns 1 and produces no `mgr_default=8.0.45`

Chart and shell checks on the same head are green, so this is a test/fixture contract drift caused by the `serviceVersion` removal, not runner infrastructure. Please update the fixture or the intended lookup contract and let the next exact-head ShellSpec sample validate it. No rerun/cancel or merge action was performed.

JashBook:
> Removing `serviceVersion` changes the immutable spec of 62 rendered `ComponentDefinition` objects whose annotations do not bypass the immutable check, while their names remain unchanged. Upgrading an installation that already contains any of these definitions makes the ComponentDefinition controller reject the new hash and mark it `Unavailable`, blocking affected Components and cluster operations.
> 
> The linked 1.0→1.1 failure uses the retained `rocketmq-exporter-1.0.3` definition, but this branch only removes `serviceVersion` from the new 1.2 definition and leaves `ComponentVersion` compatibility restricted to the current definition identity. The retained definition therefore still has no compatible 0.0.3 release, so upgraded clusters continue to fail normalization and subsequent operations with `no matched component definition`.

Thanks for the detailed review. Good catch on the upgrade impact — we're handling this differently per branch:
release-1.0: bump chart version for all affected addons
All 19 affected addons in release-1.0 embed .Chart.Version in their ComponentDefinition names (e.g. rocketmq-exporter-1.0.3), and all CMPDs carry helm.sh/resource-policy: keep via the default kblib.keepResource: true.
Bumping the chart version (x.y.z → x.y.z+1) means an addon upgrade will:
Create a new set of CMPDs under new names (without serviceVersion) — fresh objects, so no immutable-spec rejection, no Unavailable state.
Leave the old CMPDs untouched (resource-policy: keep). Existing clusters keep referencing the old definitions and continue working with zero impact.
Newly created clusters pick up the new definitions via the updated ClusterDefinition compDef references.
This follows the exact same pattern as every prior 1.0.x minor release, so it's not a new behavior.
release-1.1 and main: remove serviceVersion directly
Neither branch has been formally released yet, so there are no pre-existing CMPD instances in production environments to worry about. Direct removal is safe.
On the 1.0 → 1.1 cross-version upgrade issue you mentioned (retained 1.0 CMPDs like rocketmq-exporter-1.0.3 having no matching ComponentVersion release after 1.1 is installed) — that's a separate, pre-existing problem caused by CMPD identity discontinuity across major versions (1.0 and 1.1 already have different CMPD names). It's not introduced by this PR, and we'll track it independently.

weicao:
Exact-head review closure: `b92fdd4cda9104cb24e651c3f0b3950bb3ea8600`, rebased onto `main@dc0ab59ac1a48c98dfc7aeb9821346eb36e02786` (mergeable, no base conflict).

- `milvus-minio-1.2.0-alpha.0` still renders `spec.serviceVersion: 8.0.17`; its version-source blocker remains closed.
- Restored `serviceVersion` on all 62 same-identity ComponentDefinitions that do not bypass the immutable check. The final diff removes it only from 39 rendered definitions; a base-vs-head render audit proved each has the same metadata.name, only the `serviceVersion` spec delta, and `apps.kubeblocks.io/skip-immutable-check: "true"`.
- Added retained `rocketmq-exporter-1.0.3` to the `rocketmq-exporter-0.0.3` compatibility rule. The rendered rule now covers both `rocketmq-exporter-1.0.3` and `rocketmq-exporter-1.2.0-alpha.0`, satisfying the ComponentVersion mapping contract in `kubeblocks-addon-docs/docs/addon-api/04-component-version.md` lines 56-65.
- Added focused ShellSpec coverage: Bash 5 exact local run 1 example / 0 failures; ShellCheck PASS.
- Local offline dependency build + Helm lint + Helm render PASS for clickhouse, elasticsearch, falkordb, kafka, mongodb, redis, valkey, and rocketmq; `git diff --check` PASS.

weicao:

leon-ape (approved):
LGTM

PR #3296, opened 2026-08-06
fix: invalid config constraint for redis repl-diskless-load

wangyelei:


leon-ape (changes requested):
The Redis 5 standalone and cluster schemas now accept `repl-diskless-load=swapdb`, but the supported Redis 5.0.12 engine does not implement `repl-diskless-load` at all. The declared reconfigure action therefore fails when a Redis 5 user selects this newly advertised value, and Redis 5 also exits with a fatal configuration error if that directive is loaded from the rendered config.

wangyelei:
> The Redis 5 standalone and cluster schemas now accept `repl-diskless-load=swapdb`, but the supported Redis 5.0.12 engine does not implement `repl-diskless-load` at all. The declared reconfigure action therefore fails when a Redis 5 user selects this newly advertised value, and Redis 5 also exits with a fatal configuration error if that directive is loaded from the rendered config.Redis 5 的独立模式和集群模式现在都支持 `repl-diskless-load=swapdb` ，但支持的 Redis 5.0.12 引擎根本没有实现 `repl-diskless-load` 。因此，当 Redis 5 用户选择这个新发布的配置值时，声明的重新配置操作会失败；如果该指令是从渲染后的配置中加载的，Redis 5 也会因致命的配置错误而退出。

done

leon-ape (approved):
LGTM

PR #3304, opened 2026-08-11
chore(clickhouse): support multiple ZooKeeper endpoints

wangyelei:


addons-cluster/clickhouse/values.schema.json
```diff
@@ -142,6 +143,96 @@
         }
       }
     },
+    "zookeeper": {
```
leon-ape, line 184: **P1:** `withZookeeper` has no conditional requirement for a primary reference, while the new ComponentDefinition declaration is optional. The default `mode=withZookeeper` render therefore produces a two-shard/two-replica ClickHouse cluster with neither ZooKeeper nor embedded Keeper, leaving replicated metadata without a coordinator.

addons-cluster/clickhouse/templates/_helpers.tpl
```diff
@@ -167,4 +168,42 @@ Define clickhouse shardingComponentSpec with ComponentDefinition.
...
+{{- if gt (len $serviceRefs) 0 }}
+serviceRefs:
+{{- range $serviceRefs }}
+  - name: {{ .name }}
+    namespace: {{ .ref.namespace | quote }}
+    clusterServiceSelector:
+      cluster: {{ .ref.cluster }}
+      service:
+        component: {{ .ref.component | default "zookeeper" }}
+        service: default
```
leon-ape: **P1:** The public values schema accepts custom `service` and `port` selectors for every ZooKeeper reference, but this helper always renders `default`/`client`. KubeBlocks therefore resolves a different endpoint than the user selected, so bootstrap and ClickHouse coordination can bind the wrong service or fail.

addons-cluster/clickhouse/values.yaml
```diff
@@ -18,8 +21,8 @@ shards: 2
 replicas: 2
 
 ## ClickHouse per shard component configurations
-cpu: 1
-memory: 2
+cpu: 0.5
```
leon-ape: **P2:** This ZooKeeper change also halves the ClickHouse defaults from 1 CPU/2Gi to 0.5 CPU/1Gi for every topology. New clusters and upgrades relying on chart defaults are silently underprovisioned even when they do not use external ZooKeeper.

leon-ape (commented):
**P2:** The title omits the required `(clickhouse)` scope for this single-addon change, so it does not pass the repository title gate and cannot be approved.

leon-ape (dismissed):
LGTM

xuriwuyun (approved):

PR #3292, opened 2026-08-04
fix(postgresql): use exec role probe

weicao:
## Problem

PostgreSQL R11 against KubeBlocks main froze with both database Pods `5/5 Running` while the Cluster and Component stayed `Creating`. Every Pod repeatedly reported:

```
roleProbe code=-1: only exec action is supported: notImplemented
RoleProbeNotDone / primary is not present
```

The tested addon exact was main `5fa53f531defe335cd044723ebb854e6f4de4dbe`; its ComponentDefinitions still declared `roleProbe.http` for dbctl. The frozen failure-events artifact is SHA256 `9755c786fb6cec11d6f25e407b95bcbf4736b00af8404901e7581599c4e05db7`.

## Fix

- Call the same localhost dbctl `/v1.0/getrole` endpoint through `roleProbe.exec`.
- Bind the action to the existing `postgresql` lifecycle container so all lifecycle exec actions keep one execution image/container contract.
- Use `curl --fail --silent --show-error`, preserving the plain role token on stdout and converting non-2xx responses into probe failures.
- Advance the chart to `1.2.0-alpha.2` because ComponentDefinition lifecycle actions are immutable.
- Add a structured rendered-manifest contract test for all six PostgreSQL ComponentDefinitions.

## Verification

- Focused KubeBlocks API contract: 12 examples, 0 failures
- All PostgreSQL ShellSpec under Bash 5.3.9: 148 examples, 0 failures
- ShellCheck `--severity=error`: PASS
- `helm lint addons/postgresql`: PASS
- `helm template`: PASS
- `git diff --check`: PASS

## Exact-head runtime evidence

Fixed Test independently ran exact addon `be6048824da17383ca47517508f47d3e7e8b43d9` with tests `956d3e7fad07f5c42d36cf9c080db83c098ee1d8` as run `tr-20260809130327141-9d4cae71`.

- Natural terminal: `SUCCEEDED / PRODUCT_PASS`, version-smoke `PASS=12 / FAIL=0 / SKIP=0`
- Job UID: `f94becca-fea6-48c1-a267-445fe9edd55b`
- Cleanup: PASS; product identity and namespace absent; reported VCluster UID `9c8c428a...1208` deleted; unsafe operations `0`
- Immutable evidence member: `evidence/runner-results.tar.xz::001-events-start-vs-17.5.0-role-service-write-read.txt`
- Evidence member SHA256: `1f508347a93a6fc8c0d853739e8c104dd513b8e98092b25396547f31baee574a`
[...]

PR #3309, opened 2026-08-12
fix(postgresql): use exec role probe on current main

weicao:
Supersedes #3292, which conflicts with current `main` after PostgreSQL 13 support was added.

This successor preserves the current PG13 contracts, changes all seven PostgreSQL ComponentDefinitions from the legacy HTTP role probe to the KubeBlocks-supported exec action, and verifies the exact rendered command for 7/7 definitions.

Because `lifecycleActions.roleProbe` is immutable, the fix advances the chart and ComponentDefinition identities from `1.2.0-alpha.2` to `1.2.0-alpha.3`. ComponentVersion keeps the existing `postgresql-{major}-` compatibility prefixes, so published alpha.2 definitions are not updated in place and remain compatible. The chart does not bypass immutability with `apps.kubeblocks.io/skip-immutable-check`.

Source checks:
- focused KubeBlocks API contract: 19 examples, 0 failures
- all PostgreSQL ShellSpec under Bash 5.3.9: 157 examples, 0 failures
- structured render: `CMPD=7 alpha3=7 compatibility-prefixes=7`
- Helm lint/render, ShellCheck error gate, Bash syntax, and `git diff --check`: PASS

No Test package, runner, or Kubernetes execution is included.

### DIFF
--- a/addons/postgresql/templates/cmpd.yaml
+++ b/addons/postgresql/templates/cmpd.yaml
@@ -229,9 +229,16 @@ spec:
     roleProbe:
       periodSeconds: 1
       timeoutSeconds: 1
-      http:
-        port: "5001"
-        path: "/v1.0/getrole"
+      # KubeBlocks schedules role probes through the exec action path. dbctl
+      # returns the plain role token; curl --fail preserves non-2xx as failure.
+      exec:
+        container: postgresql
+        command:
+          - curl
+          - --fail
+          - --silent
+          - --show-error
+          - http://127.0.0.1:5001/v1.0/getrole
     switchover:
       # script worst case ~45s (leader query 5s + switchover API 20s +
       # bounded verification ~20s); kbagent clamps every action call at 60s
--- a/addons/postgresql/Chart.yaml
+++ b/addons/postgresql/Chart.yaml
@@ -4,7 +4,7 @@ description: A PostgreSQL (with Patroni HA) cluster definition Helm chart for Ku
 
 type: application
 
-version: 1.2.0-alpha.2
+version: 1.2.0-alpha.3
 
 # The helm chart contains multiple kernel versions of PostgreSQL (with Patroni HA),
 # appVersion should be consistent with the highest PostgreSQL (with Patroni HA) kernel version.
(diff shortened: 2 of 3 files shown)


addons/postgresql/templates/cmpd.yaml
```diff
@@ -229,9 +229,16 @@ spec:
     roleProbe:
       periodSeconds: 1
       timeoutSeconds: 1
-      http:
-        port: "5001"
-        path: "/v1.0/getrole"
+      # KubeBlocks schedules role probes through the exec action path. dbctl
+      # returns the plain role token; curl --fail preserves non-2xx as failure.
+      exec:
```
leon-ape, line 234: [P1] This changes seven already-published `postgresql-{12..18}-1.2.0-alpha.2` ComponentDefinitions in place. None of them renders `apps.kubeblocks.io/skip-immutable-check`, and KubeBlocks includes `lifecycleActions` in the ComponentDefinition immutable hash, so upgrading an existing alpha.2 installation marks every updated definition `Unavailable` with `immutable fields can't be updated`. Those installations therefore cannot consume this role-probe repair under the current identities.

weicao: Fixed in exact head `7cc3528b4`: chart and all seven ComponentDefinition identities now advance from `1.2.0-alpha.2` to `1.2.0-alpha.3`; the chart does not render `apps.kubeblocks.io/skip-immutable-check`. ComponentVersion keeps major-prefix compatibility (`postgresql-{12..18}-`), so existing alpha.2 definitions remain valid while new installs resolve alpha.3. Added RED/GREEN guards: old head produced focused 19 examples / 3 failures (chart stayed alpha.2, alpha.3 count 0, in-place alpha.2 count 7); new head is 19/0. Full PostgreSQL ShellSpec Bash 5.3.9 is 157/0; Helm dependency/lint/render, ShellCheck error gate, bash syntax, diff-check, and structured render assertion (`CMPD=7 alpha3=7 compatibility-prefixes=7`) pass.

leon-ape (approved):
LGTM

PR #3310, opened 2026-08-13
docs(rustfs): document version and scaling boundaries

weicao:
## Summary
- document the exact RustFS runtime and tools versions
- state the companion Cluster default-resolution boundary
- make the unsupported scale-in boundary explicit

## Validation
- `helm lint addons/rustfs`
- `helm lint addons-cluster/rustfs`
- `git diff --check`

Source-only documentation change; no Test packaging or runtime claims.


leon-ape (approved):
LGTM

PR #3300, opened 2026-08-09
fix(rustfs): pin beta.10 runtime release

weicao:
## Problem

RustFS R16 stopped before product START with `TARGET_DECLARATION`: final main declared `rustfs:1.0.0-beta.8`, while the accepted product target is `rustfs:1.0.0-beta.10`.

## Change

- pin addon `appVersion` and default ComponentVersion release/image/serviceVersion to `1.0.0-beta.10`
- pin cluster chart `appVersion` to `1.0.0-beta.10`
- bump addon chart `0.1.1 -> 0.1.2` and cluster chart `0.1.0 -> 0.1.1`
- structurally validate beta.10 runtime and roleProbe image mappings
- reject missing `images.roleProbe` and mismatched beta.9 roleProbe mappings
- fail cleanup when a fixture chart tree still exists despite an rm success status
- verify fake-rm/no-delete detection and trusted teardown
- refresh the generated root README version index

Contract anchor: `kubeblocks-addon-docs/docs/addon-api-1.1/04-component-version.md` requires predictable mapping across chart version, ComponentDefinition identity, serviceVersion, runtime image, and lifecycle action images, verified from both source and rendered objects.

## Exact Candidate

- target: `main@f8224b48188d8ff5d642b5f57220bfad4287282e`
- candidate: `19d29affe352f57521145fd7282d19c2ecdff978`
- candidate tree: `73143b9d730a8615eb4ffd790b76acb92cab2f52`
- included PR: this PR; prior merged #3247
- excluded RustFS PR: #3263, independent backup/restore scope; shared values file but YAML key overlap is zero
- Test runtime pins remain Test-owned; previous f822 intake, R16/R17, and earlier PR-head evidence do not migrate

## Validation

- pre-fix focused ShellSpec: 2 examples, 2 failures
- structured focused test: Bash 5.3 5/0; Bash 3.2 5/0
- image mutants: deleted roleProbe mapping rejected; beta.9 roleProbe mapping rejected
- cleanup mutant: rc0/no-delete rejected while one root exists; trusted teardown removes it
- full RustFS ShellSpec: Bash 5.3 59/0; Bash 3.2 12/0 (startup Bash>=4 cases conditionally skipped)
- fixture-root residue before/after focused and full runs: 0
- addon + cluster Helm lint: 2 charts, 0 failures
- addon + cluster Helm render: PASS; runtime/serviceVersion/roleProbe image all beta.10
- ShellCheck `--severity=error` on new spec: PASS
[...]

PR #3247, opened 2026-07-21
fix(rustfs): bound role probe cadence

weicao:
## Problem

In two independent RustFS beta.10 focused runs (r05 and r06), all four Pods were Ready, `/health/live` succeeded, roleProbe/Pod label was `readwrite`, and restart count stayed zero. Both runs stopped at the SB01 Cluster Running gate and did not enter S3 data-path validation.

The first role event arrived after the last InstanceSet status reconcile; `instanceStatus[].role` remained empty and Component/Cluster stayed Creating for 1500 seconds. This N=2 evidence proves that the role-publication control-plane chain did not converge in that exact scene. It does not yet prove cadence is the only cause, and it is not a RustFS data-path or release-wide claim.

## Falsifiable addon-side candidate

- add Helm render regression tests for the RustFS roleProbe cadence and unchanged exec command;
- set `periodSeconds: 1` and `timeoutSeconds: 3` on `lifecycleActions.roleProbe`;
- keep the existing loopback health check and `readwrite` / `notready` stdout semantics unchanged.

Contract anchors:

- `kubeblocks-addon-docs/docs/addon-api/05a-lifecycle-roles-and-probe.md`
- `kubeblocks-addon-docs/docs/addon-api/05c-lifecycle-action-contracts.md`

The fresh runtime gate must separately observe role event -> Pod role label -> InstanceSet role status -> Component Running -> Cluster Running. If event/label updates but InstanceSet role remains empty, the candidate is falsified and the first blocker moves to the control-plane branch.

## Current candidate identity

- target branch: `main`
- target commit: `dc0ab59ac1a48c98dfc7aeb9821346eb36e02786`
- development branch: `jessica/rustfs-roleprobe-cadence`
- candidate commit: `f170215ca27db0cf183f03db7fc9a2778068626b`
- candidate tree: `dfbc26a8f8ee297b8ffe3fe9ea9120faa1ded22d`
- addon path: `addons/rustfs`
- included addon PR: #3247
- inherited target update: merged PR #3200 is included through current `main`
- excluded addon PR: #3263; separate backup/restore scope with no changed-file overlap with this PR

The next Test package receipt must separately pin exact KubeBlocks, addon chart/package SHA, tests commit, vcluster identity/version, syncer identity/version, and any patch images. Those Test-owned runtime pins are intentionally not inferred from this Dev PR.

[...]

PR #3311, opened 2026-08-13
chore(rustfs): keep docs aligned with rendered boundaries

weicao:
## Summary

- keep README boundaries aligned with rendered RustFS objects
- serve as the RustFS development/main source PR for the exact target and child set below

## Exact source candidate

- target branch/head: `main@474b3d76af28d13737aa2fb51f28a6b19a116f22`
- development branch/head/tree: `jessica/rustfs-source-boundaries-20260813@7487efc3a8e61e745e1a41e36f095a515fabd54d` / `b091e5cdd5f2f7ad7f84474586e43d49a7c8515c`
- included merged RustFS PRs: #3247, #3300, #3310
- included open child PR: #3314 at `90644143ada02d10622e4c86356e62b9c40b5187`, tree `8ba7d44d88860aa2b05c7cc16ae5f11e75164899`
- excluded RustFS drafts: #3263 backup/restore scope; #3270 and #3271 older backup-capability chain; #3289 older versioned-definition chain
- open non-draft PRs targeting this development branch: #3314 only (last-known; refresh succeeded for #3314 exact identity)

## Version pins

- addon repository/path: `apecloud/kubeblocks-addons`, `addons/rustfs`
- addon chart: `0.1.2`
- cluster chart: `0.1.1`
- RustFS image/service version: `1.0.0-beta.10`
- kubeblocks-tools init image: `1.0.0`
- KubeBlocks controller: `UNKNOWN/null` (Test-owned)
- syncer: `UNKNOWN/null` (Test-owned)
- kubeblocks-tests: `UNKNOWN/null` (Test-owned)
- vcluster: `UNKNOWN/null` (Test-owned)
- package identity: `UNKNOWN/null` until Jean/Test publishes an exact receipt

Any included SHA change creates a new source candidate. CI, review, package, and acceptance results from an older SHA do not migrate.

## Validation

- PR #3311 exact development head: hosted 8/8 PASS and APPROVED at last exact read
- included child #3314 exact `90644143...`: focused Bash 3.2 and dash 18/0; full Bash 5.3 78/0; ShellCheck/lint/diff-check PASS
- child hosted checks: exact-head 7/7 PASS; ShellSpec run `31958490463`, job `95192738146`
- child independent exact-head review: pending; old `72448ff1` review/CI do not migrate

The child adds fail-closed API/local-write handling, concurrent compare-and-retry, exact payload/retry/manifest checks, malformed-history rejection, and large signed-integer fidelity without awk precision loss.

## Acceptance boundary

This PR records source/code-test state only. Source closure requires the included child set to stay pinned and pass exact-head review and checks. Release acceptance additionally requires Jean/Test to publish exact KubeBlocks/syncer/tests/vcluster/package identities and run the requested full RustFS suite against that frozen package. This PR does not claim Test package, Kubernetes runtime, P/F/S, cleanup, residue, or formal-N.

Successor to merged docs PR #3310.

### DIFF
--- a/addons/rustfs/scripts-ut-spec/version_pin_spec.sh
+++ b/addons/rustfs/scripts-ut-spec/version_pin_spec.sh
@@ -97,6 +97,39 @@ Describe "RustFS beta.10 version pin"
     validate_rendered_release
   }
 
+  validate_readme_boundaries() {
+    addon_render=$(helm template test "$(repo_root)/addons/rustfs") || return $?
+    cluster_render=$(helm template test "$(repo_root)/addons-cluster/rustfs") || return $?
+    printf '%s\n---\n%s\n' "${addon_render}" "${cluster_render}" | ruby -ryaml -e '
+      readme = File.read(ARGV.fetch(0))
+      documents = YAML.load_stream($stdin.read).select { |item| item.is_a?(Hash) }
+      component_definition = documents.find { |item| item["kind"] == "ComponentDefinition" }
+      component_version = documents.find { |item| item["kind"] == "ComponentVersion" }
+      cluster = documents.find { |item| item["kind"] == "Cluster" }
+      abort "rendered source objects are missing" unless component_definition && component_version && cluster
+
+      release = component_version.dig("spec", "releases").fetch(0)
+      images = release.fetch("images")
+      component = cluster.dig("spec", "componentSpecs").fetch(0)
+      replica_limit = component_definition.dig("spec", "replicasLimit")
+      member_leave = component_definition.dig("spec", "lifecycleActions", "memberLeave", "exec", "command").join("\n")
+
+      required = [
+        release.fetch("serviceVersion"),
+        images.fetch("rustfs"),
+        images.fetch("init"),
+        "creates #{component.fetch("replicas")} replicas",
+        "does not set `serviceVersion`",
+        "Scale-in is rejected",
+        "`minReplicas: #{replica_limit.fetch("minReplicas")}`"
+      ]
+      missing = required.reject { |text| readme.include?(text) }
+      abort "README boundary facts are missing: #{missing.join(" | ")}" unless missing.empty?
+      abort "memberLeave no longer rejects scale-in" unless member_leave.include?("does not support scale-in") && member_leave.match?(/exit\s+1/)
+      puts "README matches rendered version, replica, and scale-in boundaries"
+    ' "$(repo_root)/addons/rustfs/README.md"
+  }
+
   validate_cleanup_rejects_false_success() {
     chart_dir=${tmp_dir:?}
     fake_rm="${chart_dir}/fake-rm"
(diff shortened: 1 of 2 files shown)


addons/rustfs/scripts-ut-spec/version_pin_spec.sh
```diff
@@ -97,6 +97,39 @@ Describe "RustFS beta.10 version pin"
     validate_rendered_release
   }
 
+  validate_readme_boundaries() {
+    addon_render=$(helm template test "$(repo_root)/addons/rustfs") || return $?
+    cluster_render=$(helm template test "$(repo_root)/addons-cluster/rustfs") || return $?
```
leon-ape: P1: This renders the repository-root companion chart, but the ShellSpec workflow builds dependencies only for `addons/*` and `prepare_chart` only prepares a temporary addon chart. A clean checkout therefore lacks `addons-cluster/rustfs/charts/kblib`, so the new case deterministically fails and keeps the full ShellSpec gate red.

addons/rustfs/scripts-ut-spec/version_pin_spec.sh
```diff
@@ -97,6 +97,39 @@ Describe "RustFS beta.10 version pin"
...
+      member_leave = component_definition.dig("spec", "lifecycleActions", "memberLeave", "exec", "command").join("\n")
+
+      required = [
+        release.fetch("serviceVersion"),
+        images.fetch("rustfs"),
+        images.fetch("init"),
+        "creates #{component.fetch("replicas")} replicas",
+        "does not set `serviceVersion`",
+        "Scale-in is rejected",
+        "`minReplicas: #{replica_limit.fetch("minReplicas")}`"
```
leon-ape, line 129: P2: The oracle checks only `minReplicas` text and the literal claim that `serviceVersion` is omitted; it never compares rendered `maxReplicas` or verifies that the Cluster actually omits the field. Those chart contracts can change while this test still passes, leaving the README's `1 through 32` and default-resolution boundaries stale.

leon-ape (commented):
Exact-head review found two blocking test-contract defects.

addons/rustfs/scripts-ut-spec/version_pin_spec.sh
leon-ape: Fixed in `a53f22ff0`: the spec now copies `addons-cluster/rustfs` plus `addons-cluster/kblib` into its temporary fixture, builds that companion dependency explicitly, and renders only the prepared temporary charts. The focused RustFS spec passes 6/6 from a clean checkout.

addons/rustfs/scripts-ut-spec/version_pin_spec.sh
leon-ape: Fixed in `a53f22ff0`: the oracle now derives and requires the README text from both rendered `minReplicas` and `maxReplicas`, and it aborts if the rendered Cluster component contains a `serviceVersion` key. This makes drift in either documented boundary fail the test.

leon-ape (approved):
LGTM

PR #3313, opened 2026-08-13
fix(mssql): preserve restart readiness and XE path

weicao:
## Summary
- clear the stale initialized marker before restart bootstrap so readiness cannot remain green after configure failure
- bootstrap the slow-query XE event session without overwriting existing runtime settings
- escape apostrophes in the configured XE event-file path before T-SQL interpolation

## Contract
Checked against `kubeblocks-addon-docs/docs/addon-api/README.md` and `12a-minimum-acceptance.md`: this PR adds source/static evidence only; runtime remains a separate Test-owned layer.

## Validation
- focused restart lifecycle: 3 passed, 0 failed
- focused slow-query XE bootstrap: 4 passed, 0 failed
- full checked-in MSSQL source suite: 31 scripts passed
- Bash syntax, ShellCheck error-level, `git diff --check`: passed
- `helm lint addons/mssql`: passed
- `helm template task452 addons/mssql`: passed

Ported from private PR apecloud/apecloud-addons#1962 onto current public main while preserving public backup/reconfigure/D06 work.

### DIFF
--- a/addons/mssql/scripts/entrypoint.sh
+++ b/addons/mssql/scripts/entrypoint.sh
@@ -21,6 +21,12 @@ REMOTE_STANDBY_FLAG="${ROOT_DIR}/.remotestandby"
 BACKUP_DIR="${ROOT_DIR}/backup"
 AUDIT_SERVER_NAME="kbAuditLog"
 AUDIT_LOG_DIRECTORY="${AUDIT_LOG_DIRECTORY:-${ROOT_DIR}/audit}"
+SLOW_LOG_SESSION_NAME="kbSlowQueryLog"
+SLOW_LOG_DIRECTORY="${SLOW_LOG_DIRECTORY:-${ROOT_DIR}/slowlog}"
+SLOW_LOG_FILE_NAME="kb_slow_query.xel"
+SLOW_LOG_BOOTSTRAP_THRESHOLD_US=1000000
+SLOW_LOG_BOOTSTRAP_MAX_FILE_SIZE_MB=100
+SLOW_LOG_BOOTSTRAP_MAX_ROLLOVER_FILES=5
 
 # Log rotation settings
 MAX_LOG_SIZE=$((10 * 1024 * 1024))  # 10MB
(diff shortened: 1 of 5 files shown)


addons/mssql/scripts/entrypoint.sh
```diff
@@ -1784,6 +1846,7 @@ fi
 configure_function=""
 if [ -f "$init_flag" ]; then
   log "configure initialized"
+  rm -f "$init_flag"
```
leon-ape, line 1853: P1: Removing `.initialized` before `configure_initialized` completes destroys the only persistent distinction between a restart and first initialization. If this restart path fails (including the newly fail-closed XE bootstrap), the next container start sees no marker and selects `configure_primary`, `configure_secondary`, or `configure_remote_secondary`; those paths rerun non-idempotent first-boot operations such as `CREATE LOGIN` and replica/AG joins, so a transient restart configuration failure can become a persistent crash loop instead of retrying the restart path.

leon-ape (commented):
Exact-head review found one release-blocking restart-state regression.

addons/mssql/scripts/entrypoint.sh
weicao: Fixed in 565ae66749ac91838bdf31216fec823b79d911a7. Restart now writes a separate persistent `.restart-configure` identity before removing `.initialized`; readiness remains false after failure, while the next container still selects `configure_initialized`. Successful configure recreates `.initialized` and removes the retry marker. The promotion-to-primary path clears both markers. Focused lifecycle/readiness/XE tests pass 12/0; full checked-in MSSQL source suite passes 31/31; Bash syntax, ShellCheck error-level, diff check, Helm lint/template pass.

leon-ape (approved):
LGTM

leon-ape (changes requested):
The two new regression tests are placed under `addons/mssql/tests` as `*_test.sh`, but the repository Makefile and both ShellSpec workflows discover only `**/scripts-ut-spec/*_spec.sh`. Neither push nor approval-triggered CI executes these tests, so the PR can remain green while the restart-marker or slow-query XE behavior regresses.

weicao:
Fixed the CI-discovery gap in `79244d7adc2cb516ee8740b448e39a42fe6ef06c`.

- Added `addons/mssql/scripts-ut-spec/restart_xe_regression_spec.sh`, matching the repository's `**/scripts-ut-spec/*_spec.sh` discovery convention.
- The discoverable ShellSpec wrapper executes both existing regression scripts and asserts their terminal success summaries.
- Default-path discovery lists the new spec.
- MSSQL ShellSpec: 18 examples, 0 failures.
- Checked-in MSSQL source scripts: 31/31 pass.
- Focused restart/readiness/XE: 12/0.
- Helm lint/render and `git diff --check`: pass.

A repository-wide local Bash 5 run reached 2426 examples with one unrelated existing vanilla-PostgreSQL fixture failure; the scoped MSSQL suite is green.

leon-ape (approved):
LGTM

weicao:
Post-merge release audit: both automated cherry-picks failed because `release-1.0` and `release-1.1` contain zero MSSQL paths, so applying the main-only MSSQL patch would reintroduce a deleted addon rather than backport a supported release. Removed stale `pick-1.0` / `pick-1.1` labels; retained `chart-release`. Main remains exact `474b3d76` with the reviewed tree unchanged.

PR #3331, opened 2026-08-15
fix(postgresql): stop pgbouncer after setup failure

weicao:
## What changed

- Add a real `main()` entry point to `pgbouncer-setup.sh`.
- Return immediately when the common library or any generated configuration
  setup step fails, including directory creation, template copy, file writes,
  permission changes, account-file writes, and ownership setup.
- Add focused ShellSpecs proving Pgbouncer is not started after template-copy
  or ownership setup fails.

## Why

`build_pgbouncer_conf` returned the failing `chown` status, but the top-level
script ignored it and still called `start_pgbouncer`. This could hide the causal
ownership error behind a later startup failure and attempt to start with invalid
directory ownership.

The isolated pre-fix result was `build_rc=42 start_called=1`. The first
dedicated pre-fix ShellSpec failed because the setup sequence returned success
and emitted `start-called` after the mocked `chown` failure. Review then exposed
the same masking problem for intermediate commands: with mocked `cp=42` and a
successful later `chown`, the added RED case returned success and still emitted
`start-called`. A second review RED then fault-injected the first database
config `printf`; the grouped write returned the final successful `printf`
status and again allowed startup with a partial config.

## Candidate identity

- Target branch: `main`
- Fresh target commit: `474b3d76af28d13737aa2fb51f28a6b19a116f22`
- Fresh target tree: `9c47cae0a702af202ec0069236653ca70f863d1c`
- Fresh PostgreSQL subtree: `0e56df6327674c446de618ad3409839e05b293a1`
- Fresh base Pgbouncer blob: `f6f16ed301d79f152114a91c3f4078169b062470`
- Candidate commit: `dc0be5e2f8b2c7e74f9a09842c42639cefe3e5c7`
- Candidate tree: `e2f283c7de953d00ad5101c0a2c7fb08e05ae4b2`
- Scope: 2 files

Open-PR audit: draft #3117 is excluded because its PostgreSQL change is only a
repository-wide KubeBlocks compatibility annotation bump and does not intersect
this Pgbouncer source/test fix. Closed #3292 remains excluded; merged #3309 is
already included in target history.

## Validation

- Focused pre-fix ShellSpec: 1 failure, exit 101
- Review RED for intermediate `cp` failure: 2 examples, 1 failure, exit 101
- Review RED for an early grouped config write: 3 examples, 1 failure, exit 101
- Focused post-fix ShellSpec: 3 passed, 0 failed
- Full PostgreSQL ShellSpec on Bash 5.3.9: 160 passed, 0 failed
- Changed-file ShellCheck: exit 0
- Bash syntax checks: exit 0
- Helm lint: 1 passed, 0 failed
- Helm render: 41 documents, 987611 bytes
- `git diff --check`: exit 0
- Worktree: tracked clean

## Runtime boundary

This PR contains source/static evidence only. It does not claim Test-owned
runtime, package, KubeBlocks, DataProtection, syncer, vcluster, cleanup,
residue, or release-N evidence.

### DIFF
--- a/addons/postgresql/scripts/pgbouncer-setup.sh
+++ b/addons/postgresql/scripts/pgbouncer-setup.sh
@@ -89,6 +89,12 @@ start_pgbouncer() {
   su pgbouncer -c "/opt/bitnami/scripts/pgbouncer/run.sh"
 }
 
+main() {
+  load_common_library || return $?
+  build_pgbouncer_conf || return $?
+  start_pgbouncer
+}
+
 # This is magic for shellspec ut framework.
 # Sometime, functions are defined in a single shell script.
 # You will want to test it. but you do not want to run the script.
@@ -96,7 +102,4 @@ start_pgbouncer() {
 # end here. The script path is assigned to the __SOURCED__ variable.
 ${__SOURCED__:+false} : || return 0
 
-# main
-load_common_library
-build_pgbouncer_conf
-start_pgbouncer
+main
--- a/addons/postgresql/scripts/pgbouncer-setup.sh
+++ b/addons/postgresql/scripts/pgbouncer-setup.sh
@@ -20,15 +20,16 @@ build_pgbouncer_conf() {
     exit 1
   fi
 
-  mkdir -p $pgbouncer_conf_dir $pgbouncer_log_dir $pgbouncer_tmp_dir
-  cp $pgbouncer_template_conf_file $pgbouncer_conf_dir
-  echo "\"$POSTGRESQL_USERNAME\" \"$POSTGRESQL_PASSWORD\"" > $pgbouncer_user_list_file
-  # shellcheck disable=SC2129
-  echo -e "\\n[databases]" >> $pgbouncer_conf_file
-  echo "postgres=host=$CURRENT_POD_IP port=5432 dbname=postgres" >> $pgbouncer_conf_file
-  echo "*=host=$CURRENT_POD_IP port=5432" >> $pgbouncer_conf_file
-  chmod 644 $pgbouncer_conf_file
-  chmod 600 $pgbouncer_user_list_file
+  mkdir -p "$pgbouncer_conf_dir" "$pgbouncer_log_dir" "$pgbouncer_tmp_dir" || return $?
+  cp "$pgbouncer_template_conf_file" "$pgbouncer_conf_dir" || return $?
+  printf '"%s" "%s"\n' "$POSTGRESQL_USERNAME" "$POSTGRESQL_PASSWORD" > "$pgbouncer_user_list_file" || return $?
+  {
+    printf '\n[databases]\n'
+    printf 'postgres=host=%s port=5432 dbname=postgres\n' "$CURRENT_POD_IP"
+    printf '*=host=%s port=5432\n' "$CURRENT_POD_IP"
+  } >> "$pgbouncer_conf_file" || return $?
+  chmod 644 "$pgbouncer_conf_file" || return $?
+  chmod 600 "$pgbouncer_user_list_file" || return $?
 
   # Try to add user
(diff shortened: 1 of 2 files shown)


addons/postgresql/scripts/pgbouncer-setup.sh
```diff
@@ -89,14 +89,17 @@ start_pgbouncer() {
   su pgbouncer -c "/opt/bitnami/scripts/pgbouncer/run.sh"
 }
 
+main() {
+  load_common_library || return $?
+  build_pgbouncer_conf || return $?
```
leon-ape, line 94: P2: `main()` only observes `build_pgbouncer_conf`'s final status, while that function continues past failed `mkdir`, `cp`, redirection, and `chmod` commands and can return a later successful `chown`. An absent template therefore still produces a partial configuration and invokes `start_pgbouncer`, masking the causal setup error and leaving the sidecar in a misleading startup loop.

leon-ape (commented):
Exact-head review found one incomplete fail-fast path.

addons/postgresql/scripts/pgbouncer-setup.sh
weicao: Fixed in bfbe2c30e6d01de6256db34a03eeda1b1f6bce65. Added a RED ShellSpec with cp=42 and later chown=0 (2 examples, 1 failure before the fix), then made mkdir/cp/config writes/chmod/account-file writes propagate errors before start. Focused 2/0, full PostgreSQL ShellSpec 159/0, ShellCheck/Bash/diff rc0, Helm lint 1/0, render 41 docs.

addons/postgresql/scripts/pgbouncer-setup.sh
```diff
@@ -20,15 +20,16 @@ build_pgbouncer_conf() {
...
-  chmod 644 $pgbouncer_conf_file
-  chmod 600 $pgbouncer_user_list_file
+  mkdir -p "$pgbouncer_conf_dir" "$pgbouncer_log_dir" "$pgbouncer_tmp_dir" || return $?
+  cp "$pgbouncer_template_conf_file" "$pgbouncer_conf_dir" || return $?
+  printf '"%s" "%s"\n' "$POSTGRESQL_USERNAME" "$POSTGRESQL_PASSWORD" > "$pgbouncer_user_list_file" || return $?
+  {
+    printf '\n[databases]\n'
+    printf 'postgres=host=%s port=5432 dbname=postgres\n' "$CURRENT_POD_IP"
+    printf '*=host=%s port=5432\n' "$CURRENT_POD_IP"
+  } >> "$pgbouncer_conf_file" || return $?
```
leon-ape: [P2] The group reports only the final `printf` status, so a failure in either earlier config write is discarded when the last write succeeds. On this exact tree, fault-injecting the first group `printf` to return 42 while the remaining writes and `chown` succeeded made `main` return 0 and invoke `start_pgbouncer` with a partial two-line configuration, so setup can still be reported successful and PgBouncer started against an incomplete file.

leon-ape (commented):
Exact-head successor review found one remaining masked setup failure.

addons/postgresql/scripts/pgbouncer-setup.sh
weicao: Fixed in dc0be5e2f8b2c7e74f9a09842c42639cefe3e5c7. The new RED fault-injected the first database-config printf while later writes/chown succeeded: focused 3 examples, 1 failure, exit 101, with start-called. Each append now has its own failure propagation. Focused 3/0, full PostgreSQL ShellSpec 160/0, ShellCheck/Bash/diff rc0, Helm lint 1/0, render 41 docs.

leon-ape (approved):
LGTM

PR #3449, opened 2026-09-05
fix(redis): honor pull policy and official-registry image repos

weicao:
## Summary
- Redis/sentinel/twemproxy/cluster runtime containers still use `image.pullPolicy` (`redis.imagePullPolicy`).
- `metrics.image.pullPolicy` and `dbctlImage.pullPolicy` are honored again. Setting those to `Never` while `image.pullPolicy=Always` leaves the 8 metrics + 8 init-dbctl containers on `Never`. Platform Always is bound from ct-platform mappings (`metrics.image.pullPolicy`, `dbctlImage.pullPolicy`), not by ignoring Chart fields.
- `apecloud/` repository rewrite applies only when `image.registry` is `apecloud-registry.cn-zhangjiakou.cr.aliyuncs.com`. `docker.io` keeps public repos. A custom registry such as `registry.example.test` with `image.repository=redis` stays `registry.example.test/redis`, not `.../apecloud/redis`. Same rule for busybox and `ceRepository`.
- `python3 addons/redis/tests/test_pull_policy.py` 8/8. `helm lint` passes. Pair with ct-platform #391 (42 official `apecloud/*` refs). No live deployment.

## Test plan
- [x] `python3 addons/redis/tests/test_pull_policy.py`
- [x] `helm lint addons/redis`
- [x] official ACR rewrite still emits `apecloud/*`; custom registry keeps declared repository paths; metrics/dbctl keep their own pullPolicy

### DIFF
--- a/addons/redis/templates/_helpers.tpl
+++ b/addons/redis/templates/_helpers.tpl
@@ -129,6 +129,38 @@ Define redis cluster component script template name
 redis-cluster-scripts-template-{{ .Chart.Version }}
 {{- end -}}
 
+{{- define "redis.officialRepository" -}}
+{{- $registry := .registry | default "docker.io" -}}
+{{- $repository := .repository -}}
+{{- if eq $registry "docker.io" -}}
+{{- $repository -}}
+{{- else if hasPrefix "apecloud/" $repository -}}
+{{- $repository -}}
+{{- else if eq $repository "redis" -}}
+apecloud/redis
+{{- else if eq $repository "redis/redis-stack-server" -}}
+apecloud/redis-stack-server
+{{- else if eq $repository "malexer/twemproxy" -}}
+apecloud/twemproxy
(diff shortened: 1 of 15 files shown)


leon-ape (approved):
LGTM

leon-ape (approved):
Reviewed928969801652deb5da1cd65b8cfca06b0ae0d530. The two rendering regressions found during review are fixed: per-image pullPolicy overrides remain effective, and a custom registry does not receive an unsolicited apecloud repository prefix.

Eight real Helm regression cases passed. The complete actual source Chart, combined with the platform Redis mappings, rendered1,004,809bytes with42 image references, all required images covered and pull policies set toAlways for the test configuration. This uses the full Chart rather than only the reduced platform fixture. Redis tests-source and platform/shared-cleanup combination reviews are also complete; live execution remains a separate deployment step.

PR #3450, opened 2026-09-06
fix(elasticsearch): honor image pull policy for plugin and exporter

weicao:
Elasticsearch ignores `image.pullPolicy` for `prepare-plugins`, and the exporter has no explicit policy. Setting `image.pullPolicy=Always` therefore fails the daemon's rendered runtime coverage check.

Use the existing image policy value for both containers. The default remains IfNotPresent. Scope: release-1.1 base eaef6cabda93d99ee80f4b946cda863e5a66e393; no change to image references or engine versions.

Validation: actual Helm rendering with default, Always and Never passes for every runtime container. The original Chart fails production runtime image coverage; the patched Chart passes coverage, frozen mapping reread and wrong-registry rejection. Registry digests were fixtures; no live Kubernetes or registry operations were performed.

ComponentDefinition runtime context: `kubeblocks-addon-docs/docs/addon-api/02-component-definition.md`. Always is the daemon's execution requirement, not a new KubeBlocks API requirement.

### DIFF
--- a/addons/elasticsearch/templates/_helpers.tpl
+++ b/addons/elasticsearch/templates/_helpers.tpl
@@ -281,7 +281,7 @@ lifecycleActions:
 runtime:
   initContainers:
     - name: prepare-plugins
-      imagePullPolicy: IfNotPresent
+      imagePullPolicy: {{ .Values.image.pullPolicy }}
       command:
         - sh
         - -c
--- a/addons/elasticsearch/templates/_helpers.tpl
+++ b/addons/elasticsearch/templates/_helpers.tpl
@@ -470,6 +470,7 @@ runtime:
           readOnly: true
 {{- end }}
     - name: exporter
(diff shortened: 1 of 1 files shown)


leon-ape (approved):
LGTM

leon-ape (approved):
Reviewed exact head d506c8ea366fb871ed891f8a7972cb8a9b51b9f3 against release-1.1 eaef6cabda93d99ee80f4b946cda863e5a66e393. The single helper change correctly applies the existing image.pullPolicy to prepare-plugins and exporter.

Independent actual Helm renders passed default, Always and Never for all 66 containers. Production image derivation/coverage passed for 23 image references, frozen reread was unchanged and wrong-registry input was rejected. The original Chart separately reproduced the non-Always production coverage rejection. Registry digests are fixtures; this is not live addon acceptance. Hosted Chart, ShellCheck and ShellSpec checks passed on this head.

PR #3452, opened 2026-09-07
fix(milvus): honor MinIO image pull policy

weicao:
Setting `images.pullPolicy=Always` or `Never` currently renders `IfNotPresent`
for both the MinIO container and its volume-permissions init container. Correct
the two reversed Helm `default` arguments so explicit policies take effect;
default, empty, and null values retain `IfNotPresent`.

Adds a real Helm rendering regression with YAML assertions for both containers:
six value cases in install and upgrade mode. Helm 3.19.0 and 4.2.0 pass all 12
render cases. Before the fix, Always/Never produce eight failed container
assertions. Complete before/after comparison confirms only the two intended
policies change; eight default/empty/null render pairs remain byte-identical.

Chart and shell checks pass. ShellSpec run [34073576496](https://github.com/apecloud/kubeblocks-addons/actions/runs/34073576496)
fails the unchanged PostgreSQL `pgdump_restore_spec.sh:96` case. The exact base
`e7b8fd03` already failed that same case in [33831535711](https://github.com/apecloud/kubeblocks-addons/actions/runs/33831535711);
both runs report 2463 examples, 1 failure, 9 pendings. The PostgreSQL script and
test blobs are identical between this PR and its base.

Design: [DOC-95@v1](https://infracreate.com/project-manager/designs/b3904720-a21e-4f1a-94fb-8572a5ae85b2).
Validation: `python3 addons/milvus/tests/test_minio_pull_policy.py`.
No running cluster was changed. The release-1.1 backport is #3453;
`nopick` avoids duplicating this explicit backport.

### DIFF
--- a/addons/milvus/templates/cmpd-minio.yaml
+++ b/addons/milvus/templates/cmpd-minio.yaml
@@ -39,7 +39,7 @@ spec:
     initContainers:
       - name: volume-permissions
         image: {{ .Values.images.ostools.registry | default ( .Values.images.registry | default "docker.io" ) }}/{{ .Values.images.ostools.repository }}:{{ .Values.images.ostools.tag }}
-        imagePullPolicy: {{default .Values.images.pullPolicy "IfNotPresent"}}
+        imagePullPolicy: {{ default "IfNotPresent" .Values.images.pullPolicy }}
         command:
           - /bin/bash
           - -ec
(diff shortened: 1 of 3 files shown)


leon-ape (approved):
Reviewed this exact commit. Both template fixes are correct; independently verified full source/tree/diff, reproduced the original failures and passed all 12 real Helm cases/24 container assertions on each corrected branch. Full before/after manifests differ only in the two intended policies. The failing PostgreSQL ShellSpec case also fails on the unchanged base; this approval does not waive required CI or claim main is green.

leon-ape (approved):
LGTM

PR #3453, opened 2026-09-07
fix(milvus): honor MinIO image pull policy [release-1.1]

weicao:
Backport of #3452 to release-1.1.

Setting `images.pullPolicy=Always` or `Never` currently renders `IfNotPresent`
for both MinIO and volume-permissions. Correct the two reversed Helm `default`
arguments. Default, empty, and null values still use `IfNotPresent`.

The main commit is cherry-picked onto release-1.1 without changing this branch's
existing account configuration, image addresses, commands, or chart version.
Real Helm 3.19.0 and 4.2.0 regression passes 12 render cases and 24 container
assertions per version. Original source fails eight Always/Never assertions.
Complete before/after object comparison permits only the two intended policy
changes, and all eight default/empty/null render pairs are byte-identical.

Design: [DOC-95@v1](https://infracreate.com/project-manager/designs/b3904720-a21e-4f1a-94fb-8572a5ae85b2).
Validation: `python3 addons/milvus/tests/test_minio_pull_policy.py`.
No running cluster was changed. `nopick` is used because both target branches
have explicit PRs.

### DIFF
--- a/addons/milvus/templates/cmpd-minio.yaml
+++ b/addons/milvus/templates/cmpd-minio.yaml
@@ -34,7 +34,7 @@ spec:
     initContainers:
       - name: volume-permissions
         image: {{ .Values.images.ostools.registry | default ( .Values.images.registry | default "docker.io" ) }}/{{ .Values.images.ostools.repository }}:{{ .Values.images.ostools.tag }}
-        imagePullPolicy: {{default .Values.images.pullPolicy "IfNotPresent"}}
+        imagePullPolicy: {{ default "IfNotPresent" .Values.images.pullPolicy }}
         command:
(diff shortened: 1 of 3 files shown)


leon-ape (approved):
Reviewed this exact backport. Both template fixes preserve defaults and honor Always/Never. Independently verified complete before/after source trees and exact diff; reproduced 8 original policy failures, passed 12 corrected real Helm cases/24 container assertions, and confirmed full manifests change only the two intended policies. Executed PR checks are successful. Milvus+etcd integration and live Job acceptance remain separate.

leon-ape (approved):
LGTM

PR #3454, opened 2026-09-07
fix(etcd): backport inject image pull policy [release-1.1]

weicao:
Backport the one-line etcd fix from #3126 to release-1.1. The `inject-bash`
init container currently ignores `images.pullPolicy=Always` or `Never` because
Helm `default` receives a nonempty constant as its input. Reverse the arguments
to honor explicit policies while retaining the `IfNotPresent` fallback.
The primary etcd container remains independently controlled by `image.pullPolicy`.
Main already includes this fix in `bd611ed3` and needs no duplicate template change.

Adds a real Helm/YAML regression for install and upgrade: defaults, explicit
IfNotPresent, Always, Never, empty, null, primary-only, and differing primary/init
policies. Helm 3.19.0 and 4.2.0 pass all 16 render cases on the corrected etcd
chart and on current main. The original release-1.1 chart fails six init-policy
assertions. Complete before/after manifests change only inject-bash's policy.

Also validated real Milvus + etcd renders against the Milvus fix in #3453:
default/Always/Never in install and upgrade mode, with four named containers
checked per case. The original combination fails four inject-bash assertions;
the corrected combination passes all 24 assertions. No cluster was contacted.

Design: [DOC-96@v1](https://infracreate.com/project-manager/designs/11519a1a-b650-4de8-adf2-0c23d738ec87).
Regression: `python3 addons/etcd/tests/test_inject_pull_policy.py`.
Based on release-1.1 `2a9e1758` after #3453, with only etcd changes in this PR.
`nopick` is appropriate because main already has the fix.

### DIFF
--- a/addons/etcd/templates/cmpd.yaml
+++ b/addons/etcd/templates/cmpd.yaml
@@ -13,7 +13,7 @@ spec:
   runtime:
     initContainers:
       - name: inject-bash
-        imagePullPolicy: {{default .Values.images.pullPolicy "IfNotPresent"}}
+        imagePullPolicy: {{ default "IfNotPresent" .Values.images.pullPolicy }}
         volumeMounts:
           - name: scripts
             mountPath: /scripts
--- a/addons/etcd/tests/README.md
+++ b/addons/etcd/tests/README.md
@@ -1,0 +1,20 @@
+# Inject Image Pull Policy Regression
+
+Requires Helm 3 or 4, Python 3.8+, and PyYAML (`python3 -m pip install PyYAML`).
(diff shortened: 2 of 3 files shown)


leon-ape (approved):
Reviewed exact source and full archives against base 2a9e1758. Independently reran 100 Helm renders: original single-chart and combined failures reproduced; fixed release and current main passed. All 28 whole-render comparisons preserve other fields, including unchanged Milvus output. Hosted executed checks pass. Source/render approval only; no live cluster or Job acceptance.

PR #3456, opened 2026-09-07
feat(mongodb): support TLS for replica set and sharding clusters

gnolong:
## Summary

Enable TLS for MongoDB replica set and sharding clusters through native KubeBlocks certificate mounts and Cluster CR settings. Combine the certificate and private key under `/etc/mongodb/tls/mongodb.pem` with restrictive permissions, and configure TLS through startup arguments. MongoDB 4.0 uses legacy SSL flags; supported newer versions use TLS flags.

Share MongoDB client selection, TLS options and JSON result parsing across runtime scripts, shard management, custom balancer operations and backup/restore ActionSets. Keep shell diagnostics separate from query results so they cannot affect state checks. Managed clients and internal MongoDB connections use encrypted connections without certificate or hostname verification. Update Syncer to `0.7.9`.

The branch includes the main-branch removal of the sharding provision order. No new unit test cases are added. Existing startup and balancer test fixtures are updated for the shared client helpers.

## Validation

- Local MongoDB 4.4 and 8.0 addon-suite tests passed.
- Helm lint and template rendering passed.
- MongoDB ShellSpec: 75 existing examples passed.
- Bash syntax checks, ShellCheck 0.11.0 with the CI options, and git diff checks passed.


gnolong:
how otld connects to mongodb when tls is enabled?

wangyelei (dismissed):

xuriwuyun (approved):

PR #3457, opened 2026-09-08
fix(milvus): declare MinIO password generation [release-1.1]

weicao:
The embedded MinIO ComponentDefinition requires the admin password credential,
but release-1.1 omits the explicit password-generation configuration. Add
`passwordConfig: {}` so the matching KubeBlocks CRD applies its standard defaults.
The admin account, credential references, and image pull policies stay unchanged.

This independently ports the declaration from #3180 onto release-1.1 `0cb6a190`.
It requires the ComponentDefinition CRD/API from apecloud/kubeblocks#10739
(merged as `f3a6d925`) or a compatible successor. Rendering does not prove the
deployed CRD and controller match. The original main PR remains separate.

Validation:
- Real Helm 3.19.0 and 4.2.0 install/upgrade regressions confirm the explicit
  configuration and required password reference; the unchanged pull-policy
  matrix also passes. The baseline fails both password assertions.
- Full install and upgrade output comparison changes only the account's new
  `passwordConfig` field.
- `git diff --check` passes.

Run: `python3 addons/milvus/tests/test_minio_pull_policy.py`.
Design: [DOC-97@v1](https://infracreate.com/project-manager/designs/6132c0a2-22a3-4027-b70d-84345c7c4ec8).
Live nine-suite validation is tracked separately and is not claimed by these
source/render results. No running cluster was changed for this PR.

### DIFF
--- a/addons/milvus/templates/cmpd-minio.yaml
+++ b/addons/milvus/templates/cmpd-minio.yaml
@@ -14,6 +14,7 @@ spec:
   systemAccounts:
     - name: admin
       initAccount: true
+      passwordConfig: {}
   vars:
     - name: MINIO_ACCESS_KEY
       valueFrom:
--- a/addons/milvus/tests/README.md
+++ b/addons/milvus/tests/README.md
@@ -18,3 +18,10 @@ Optional environment variables: `HELM_BIN` selects the Helm executable;
 `CHART_DIR` selects another Milvus source directory with an adjacent kblib;
 `RENDER_DIR` retains the complete rendered YAML for each case.
 This test does not contact a Kubernetes cluster.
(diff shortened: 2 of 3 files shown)


leon-ape (approved):
Independent fixed-source review passed. On directly fetched release-1.1 sources, the baseline reproduces both missing-passwordConfig assertions; the candidate passes password and pull-policy render checks. Full parsed Helm install/upgrade comparison changes only admin.passwordConfig. Matching KubeBlocks f3a6d925 API/CRD semantics and all hosted checks were verified. Approval is for source/CI; a compatible deployed CRD/controller and live suite results remain separate.

PR #3458, opened 2026-09-08
fix(etcd): propagate role probe failures [release-1.1]

weicao:
The roleprobe entrypoint can return success with empty output when endpoint
status fails, or misclassify malformed output as a leader. Propagate failure
through the entrypoint and require exactly one valid MemberID, Leader, and
IsLearner field before classifying a role.

This independently ports commit `276a9b4c` from #3267 to release-1.1 `0cb6a190`.
The two changed files are byte-identical to that source commit. The roleprobe
script, its test, and common library on this release baseline match the original
commit's parent; no memberJoin changes are included. The original stacked PR
and its branch remain separate.

Validation:
- Production-bound focused ShellSpec: 19 examples, 0 failures.
- All etcd ShellSpec tests on this release baseline: 48 examples, 0 failures.
- Six controlled production-entrypoint cases: original client-error/malformed/
  empty-output behavior fails three assertions; the corrected entrypoint passes
  all six, preserving leader/follower/learner output.
- Real Helm install/upgrade output differs only in the ConfigMap roleprobe
  script, whose rendered content matches the production file.
- Bash syntax, ShellCheck, and `git diff --check` pass.

Design: [DOC-97@v1](https://infracreate.com/project-manager/designs/6132c0a2-22a3-4027-b70d-84345c7c4ec8).
Actual Milvus/etcd runtime validation is tracked separately. These are source,
controlled subprocess, and rendering results; no running cluster was changed.

### DIFF
--- a/addons/etcd/scripts-ut-spec/roleprobe_spec.sh
+++ b/addons/etcd/scripts-ut-spec/roleprobe_spec.sh
@@ -1,132 +1,215 @@
 # shellcheck shell=bash
-# shellcheck disable=SC2034
+# shellcheck disable=SC2034,SC2329
 
-# validate_shell_type_and_version defined in shellspec/spec_helper.sh used to validate the expected shell type and version this script needs to run.
 if ! validate_shell_type_and_version "bash" 4 &>/dev/null; then
   echo "roleprobe_spec.sh skip cases because dependency bash version 4 or higher is not installed."
   exit 0
 fi
 
-source ./utils.sh
-
-# The unit test needs to rely on the common library functions defined in kblib.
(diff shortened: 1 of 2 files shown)


leon-ape (approved):
Independent fixed-source review passed. All 48 etcd ShellSpec cases and six production-entrypoint cases pass; the unchanged base reproduces three false-success cases. Full Helm install/upgrade comparison changes only the ConfigMap roleprobe script. Production/spec bytes match276a9b4c and base inputs match its parent, without memberJoin changes. All hosted checks passed. Source/CI approval does not claim live etcd or Milvus acceptance.

PR #3459, opened 2026-09-08
fix(milvus): backport standalone parameter renderer

weicao:
Backport the Milvus standalone ParametersDefinition for the frozen release-1.1 KB 51883817 API. The 1.1 CRD requires component matching, parameter references, and config file binding on ParamConfigRenderer; the ParametersDefinition retains only supported fileName/schema/static fields. The standalone config is marked externalManaged. Adds a Helm render regression asserting the native 1.1 shape. Excludes Pulsar and 1.2 binding changes.

### DIFF
--- a/addons/milvus/templates/_helpers.tpl
+++ b/addons/milvus/templates/_helpers.tpl
@@ -347,6 +347,7 @@ Milvus user config - standalone
   volumeName: milvus-config
   namespace: {{.Release.Namespace}}
   defaultMode: 420
+  externalManaged: true
   restartOnFileChange: true
 - name: scripts
   template: {{ include "milvus-scripts.configTemplateName" . }}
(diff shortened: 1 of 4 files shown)


leon-ape (approved):
Independent fixed-source review: GO. Verified both complete source archives, reconstructed the exact candidate tree from the supplied patch, validated PD/PCR against exact KB51883817 schemas including unknown fields, and checked full PCR-to-PD/component/config/user.yaml bindings. Install/upgrade render comparisons pass both on the original base and the clean composition with current release-1.1; the existing MinIO/etcd fixes remain intact. The corrected rc=0 strict server dry-run has both admitted specs exactly matching the independent candidate render. No blocking code finding. Scope is the standalone parameters migration; full engine acceptance and separate services/Pulsar changes remain outside this source approval. After normal merge, verify the actual merged source preserves this exact four-file delta atop the current release base.

leon-ape (approved):
LGTM

PR #3285, opened 2026-07-31
feat(sharding): use non-blocking shard removal

leon-ape:
## What changed

Enable `nonBlocking` for the `shardRemove` lifecycle action in:

- Redis
- FalkorDB
- MongoDB

Set `timeoutSeconds: 3600` for all three actions. For non-blocking actions, this is the total execution budget across all argument invocations, retries, and retry intervals.

## Why

These shard-removal actions perform data-dependent work that can exceed the synchronous Action limit:

- Redis rebalances hash slots away from the shard and removes its cluster nodes.
- FalkorDB rebalances slots to zero and removes the shard nodes.
- MongoDB waits for chunk migration, moves database primaries when necessary, and waits for `removeShard` to complete.

Without an explicit timeout, MongoDB would use the 30-second default. The existing 600-second values for Redis and FalkorDB would become hard total limits after switching to non-blocking execution. A finite one-hour window matches the intended long-running shard-operation configuration while still bounding stuck actions.

KubeBlocks can persist the selected targets and poll these actions until they reach a terminal result.

## User impact

Scale-in for these sharded databases no longer depends on shard removal finishing within one synchronous controller call and has up to one hour to finish migration and retries. This requires apecloud/kubeblocks#10725.

## Validation

For each of `redis`, `falkordb`, and `mongodb`:

- `helm dependency build --skip-refresh addons/<addon>`
- `helm lint addons/<addon>`
- `helm template <addon> addons/<addon>` and verified `spec.lifecycleActions.shardRemove.nonBlocking: true` and `timeoutSeconds: 3600`

PR #3470, opened 2026-09-09
fix(redis): support Redis 6.0 Sentinel DNS and Cluster TLS announcements

### LINKED ISSUE #3469: Redis 6.0.20 Sentinel DNS registration and Cluster TLS announce compatibility
Redis 6.0.20 support added in #3465 leaves two compatibility failures.

1. With the default DNS replication topology, postProvision passes the primary Pod FQDN to `SENTINEL MONITOR`. Redis 6.0 Sentinel only accepts numeric addresses. Registration fails after retries and the primary remains unmonitored.
2. With Redis Cluster, TLS and NodePort or hostNetwork, the startup script emits `cluster-announce-tls-port`. Redis 6.0.20 does not support that directive and exits with `FATAL CONFIG FILE ERROR`. In Redis 6.0, `cluster-announce-port` advertises the TLS port when `tls-cluster yes` is set.

Reviewer reported both failures against #3465 head `1cb5d5421e00303cfd2876fd7d2887db3946825b` using actual Redis 6.0.20 binaries: numeric-IP Sentinel registration and default-network TLS PING succeeded as controls. That runtime evidence is reviewer-owned; it is not a runtime validation of the proposed fix.

Reproduction scenarios:
- Provision Redis 6.0.20 replication with Sentinel and default Pod DNS addresses; inspect postProvision MONITOR arguments and the resulting monitored master list.
- Provision Redis 6.0.20 Cluster with TLS and NodePort, then hostNetwork; inspect the generated configuration and server startup error. Compare default-network TLS with the same binary and certificates.

Code-level regression tests reproduce the hostname registration and TLS configuration failures on current main `5e2f1e1e1a323602fa7f90f70eb0e6791cc4c2ea`. Follow-up patch validation must include real provisioning, monitoring/failover and TLS data connectivity on its own pinned package.

weicao:
Fixes #3469. Fix two Redis 6.0.20 compatibility failures introduced with #3465:

- Resolve primary hostnames before Sentinel MONITOR only for Redis 6.0, retaining numeric IPv4/IPv6 addresses and Redis 6.2+ DNS behavior. Failed resolution stops registration; existing monitored primaries are retained.
- Use `cluster-announce-port` for Redis 6.0 TLS on NodePort/hostNetwork. Keep the separate `cluster-announce-tls-port` behavior for 6.2+, and retain `tls-cluster yes` and plaintext `port 0`.

Validation: 11 new regression cases; focused 27/27 and full Redis ShellSpec 444/444 pass with Bash 5.3. Helm lint/template and diff checks pass. Baseline with the new tests fails four cases before the fix. ShellCheck reports two existing warnings (SC2155 and SC2188) outside the patch.

Draft: real Redis 6.0.20 provisioning, Sentinel failover and TLS data-path acceptance of this patch remain pending. Existing reviewer binary reproductions establish the original failures only. No runtime PASS is claimed.

### DIFF
--- a/addons/postgresql/scripts-ut-spec/pgdump_restore_spec.sh
+++ b/addons/postgresql/scripts-ut-spec/pgdump_restore_spec.sh
@@ -93,14 +93,25 @@ EOF
     The result of function call_log should include "pg_restore"
   End
 
-  It "treats ignored per-object errors as success under the default CONTINUE policy"
+  It "fails when the ignored-errors warning has no error details"
     export DATASAFED_LIST_OUT="backup-test.tar"
     export PG_RESTORE_EXIT=1
     export PG_RESTORE_STDERR="pg_restore: warning: errors ignored on restore: 3"
     When run bash "$(script_path)"
+    The status should be failure
+    The output should include "parameters:"
+    The error should include "errors ignored on restore"
+  End
+
+  It "treats existing-object errors as success under the default CONTINUE policy"
+    export DATASAFED_LIST_OUT="backup-test.tar"
+    export PG_RESTORE_EXIT=1
+    export PG_RESTORE_STDERR='pg_restore: error: could not execute query: ERROR: relation "items" already exists
+pg_restore: warning: errors ignored on restore: 1'
+    When run bash "$(script_path)"
     The status should eq 0
     The output should include "treating as success under conflict_policy=CONTINUE"
-    The error should include "errors ignored on restore"
+    The error should include "already exists"
   End
 
   It "propagates ignored-error failures when conflict_policy is FAIL"
--- a/addons/redis/redis-cluster-scripts/redis-cluster6-server-start.sh
+++ b/addons/redis/redis-cluster-scripts/redis-cluster6-server-start.sh
@@ -587,10 +587,11 @@ build_cluster_announce_info() {
     {
       echo "cluster-announce-ip $redis_announce_host_value"
       echo "cluster-announce-bus-port $redis_announce_bus_port_value"
-      if [ "$TLS_ENABLED" == "true" ]; then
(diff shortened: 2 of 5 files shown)


weicao:
CI status for head `73103b9bc00a5927116f60cfaa7bf3e5e4db56eb`:

- `shell-check`: PASS
- all chart checks and `generate-files`: PASS
- `shellspec-test`: failed after 2,500 examples because unrelated `addons/postgresql/scripts-ut-spec/pgdump_restore_spec.sh:96` expected a conflict-policy message but received only the parameter line. The log reports 1 failure and 9 pending examples, all other failures absent.

The Redis-specific full ShellSpec run on this exact source passed 444/444 with Bash 5.3, and the focused F1/F2 run passed 27/27. This CI failure is recorded as an unrelated repository-wide test failure; no Redis test was reported failing. Hosted checks remain non-green, so this draft stays blocked from readiness.

weicao:
Applied the combined test-only fix from #3474 (`578f19e2c689504744234865a0479a9bb2d75426` plus `d591b76a075be86e3c4dd6b96b70049d58de8c8a`). Only `addons/postgresql/scripts-ut-spec/pgdump_restore_spec.sh` changes; production files are unchanged.

New candidate heads:
- #3285: `69bb54f4cd3b07da5ab6eac7466f4b26b96a7071`
- #3470: `1daae7f35a9e6c06c2cad772ab79907373cda9d5`

The PostgreSQL regression passes 7/7 on each candidate when run separately. Concurrent local runs initially collided through the existing shared `/tmp/pg_restore.log`; both passed after serial reruns. Hosted checks have been triggered on the new heads and are pending. Earlier full engine test results belong to the preceding heads; those engine sources are unchanged by this test-only update.

Both PRs remain draft pending exact-head CI, independent review and runtime acceptance. This update does not provide runtime evidence or change execution authorization.

PR #3474, opened 2026-09-09
test(postgresql): fix pgdump restore ShellSpec fixture

weicao:
Update the pgdump restore ShellSpec fixture to model a benign existing-object error followed by the ignored-errors warning, and add a regression ensuring warning-only output remains a failure. Production restore logic is unchanged.


leon-ape (approved):
LGTM

ldming (approved):

PR #3475, opened 2026-09-10
fix(etcd): make rollout joins replay-safe and wait for registration

shanshanying:
Refs apecloud/kubeblocks#10873.

Make memberJoin replay-safe with identity checks before and after adding. For empty-WAL `existing` startup, confirm registration via the peer `/members` API without requiring quorum, and map registered unnamed peers by exact peer URL. Preserve `new`/WAL startup behavior and the original immutable memberJoin command; keep bounded failure diagnostics. Chart version unchanged.

Validation: 47 focused tests, Bash 3 skip checks, ShellCheck, Helm lint/render and API dry-run passed. On kind-kbv11 (KB 1.1.0-beta.11, etcd 3.5.6), fresh sequential 1→2→3 and concurrent 3→5 passed with zero restarts, healthy membership, equal KV hashes and all 21 keys on every replica. Captured the no-quorum 1→2 window and two simultaneous unnamed members. Updating scripts preserved ComponentDefinition spec/generation and Available status.

Limitation: batch 1→3 still exposed a second Pod entering `new` mode; it is not claimed as passing. The repository example supports one added member per operation. Prior canary/restart validation predates this revision; TLS integration was not rerun.

### DIFF
--- a/addons/etcd/Chart.yaml
+++ b/addons/etcd/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.1.0-alpha.1
+version: 1.1.0-alpha.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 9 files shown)


shanshanying:
Closing this PR because its head branch is owned by a personal fork. I will recreate the change from the official release-1.1 branch with the peer fallback review fix.

PR #3481, opened 2026-09-13
fix(valkey): honor configured runtime image pull policies

weicao:
The Valkey Chart exposes image.pullPolicy and metrics.image.pullPolicy, but its data, Sentinel and metrics containers hardcode IfNotPresent. When daemon image mappings generate Always, Helm ignores those policy overrides and runtime coverage rejects the rendered Sentinel ComponentDefinition before tests start.

Bind the three template fields to their existing values keys. Defaults remain IfNotPresent. No test suite assertions or runner validation rules change.

Validation: five offline Helm regression tests pass, covering both supported majors, defaults, Always/IfNotPresent/Never, independent engine and metrics overrides, and metrics disabled. The same tests fail on the original Chart. A replay using the frozen daemon mappings and production values/coverage functions reproduces the original Sentinel rejection and passes all six runtime containers with the patch. Parsed default renders are identical; helm lint passes. Never is tested as a Chart setting only, not as a policy supported by this daemon. No live Run was submitted for this patch.

### DIFF
--- a/addons/valkey/templates/cmpd-valkey-sentinel.yaml
+++ b/addons/valkey/templates/cmpd-valkey-sentinel.yaml
@@ -199,7 +199,7 @@ spec:
   runtime:
     containers:
       - name: valkey-sentinel
-        imagePullPolicy: IfNotPresent
+        imagePullPolicy: {{ $.Values.image.pullPolicy }}
         ports:
           - name: valkey-sentinel
             containerPort: 26379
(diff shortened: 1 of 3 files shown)


weicao:
Closing per the ct-platform ownership boundary: addon image pull policies and addon behavior remain owned by the addon team. ct-platform must not force a pull policy; configured per-image policies must pass through unchanged. See Infracreate DOC-78@v16.

PR #3485, opened 2026-09-14
fix(mongodb): generate root passwords with the v1 account schema

weicao:
MongoDB root credentials are empty on the current KubeBlocks v1 API because the Chart emits `passwordGenerationPolicy`, which the installed CRD does not accept. Use `passwordConfig` for replica-set and config-server root accounts, preserving length 16, 8 digits, 0 symbols and MixedCases. Required credential consumers and entry scripts are unchanged.

The rendered-Chart regression reproduces the old-field failure and checks both account producers, required root credentials and runtime aliases. All 76 MongoDB ShellSpec examples, Helm lint and diff checks pass; the candidate also passed remote ShellSpec, Shell Check and Chart workflows. The actual Guest v1 API rejects the old field and retains the new policy.

Focused runtime validation of candidate `30bfdc1fb8c9a0893878590c34806dac6f5faa23`, KubeBlocks `35c611fd58886a017a15da41efae40a6bb8f2b07`, tests `fe984839a538a8a730bd76789874f616a9426b7c` and vcluster `eacd8ce99b51793d0adb3a6e1a8f9416e715e62c` passed authenticated ping on all three Pods. Actual Helm inputs fixed syncer to `sha256:6c366e94fa0222dfe8f5c6cc2e24ffb10872769e7413c5b5cd71cb882c23b87f`; all three imageIDs and copied syncer/syncerctl binary hashes matched. Fifteen mounted-script checks matched the frozen Chart. Secret generation constraints, credential aliases and resource ownership passed. Ordinary UID-safe cleanup confirmed Cluster/namespace absence and zero Guest PVC/PV or associated Host residues.

This is focused validation recovered after a node DNS/image-pull interruption, not a complete original-suite PASS. The original failed Run remains unchanged. Different cached contents under `syncer:0.7.9` were traced to two releases overwriting that tag ([August 20](https://github.com/apecloud/syncer/actions/runs/32355926737), [September 10](https://github.com/apecloud/syncer/actions/runs/34453344356)); explicit digest binding closes the tested runtime identity, while node import provenance and requested host cache clearing remain separate operations.

Contract: kubeblocks-addon-docs `docs/addon-api/07-accounts-and-tls.md:61` and `docs/troubleshoot/addon-account-schema-version-pruning.md`. This change targets main's `>=1.2.0` Chart/API and is not a blanket backport to older CRDs.


leon-ape (approved):
LGTM

PR #3489, opened 2026-09-17
fix(rabbitmq): version served-v1 account schema upgrade

weicao:
Using `passwordGenerationPolicy` with the served-v1 CRD loses the RabbitMQ root password constraints during schema pruning. This change uses `passwordConfig`, preserving length 16, 8 digits, 0 symbols, mixed letter case, and the existing init-account behavior.

Because `systemAccounts` is immutable, the addon and matching Cluster Chart now use `1.2.0-alpha.2`. Helm creates `rabbitmq-1.2.0-alpha.2` instead of replacing the existing alpha.1 definition. The default resource retention preserves older definitions. No immutable-check bypass is added. Existing Clusters and account Secrets are not automatically migrated; a live Cluster upgrade remains a separate validation scope.

Validation for `880ef2466eb224745a7ded0159e3d2888d7845f6`:
- ShellSpec: 9 examples, 0 failures, including real Helm-rendered definition, parameter, ComponentVersion and Cluster references.
- Original alpha.1 identity fails the new regression; independently restoring only the Cluster Chart to alpha.1 also fails.
- Both charts pass Helm lint/render; real Kubernetes structural pruning with the pinned served-v1 CRD removes the legacy field and preserves all four new constraints. ShellCheck and diff check pass.

Historical live evidence belongs to prior head `85fa0fbf06272968837d70babb9985901d19137a`, platform `a924932659f46438108c45426d898a6e550d21a4`, tests `327f86194653ef4f299a213b7121dcee1a1bc414`, KB `ade0905e45b6ab0e28ea09673462596f134fc2f6`, Run `tr-20260917221602351-9e611ca9`: env 33/0/0, static 81/0/0, smoke 10/0/0, total 124/0/0; product and cleanup PASS, one execution. Reviewed source and original logs cover newly published/read messages before and after restart, not old-message persistence across restart; six HTTP bodies were not included. These results are not claimed for the revised alpha.2 candidate or for a live upgrade.

Contract anchors: `kubeblocks-addon-docs/docs/addon-api/02-component-definition.md`, `04-component-version.md`, and `07-accounts-and-tls.md`. The immutable-field failure is also established by the review finding and current controller implementation, not inferred as a new addon API requirement.


leon-ape (commented):
This PR changes the immutable `spec.systemAccounts` of the existing `rabbitmq-1.2.0-alpha.1` ComponentDefinition from `passwordGenerationPolicy` to `passwordConfig`, while the chart and ComponentDefinition identity remain unchanged and the rendered definition has no `apps.kubeblocks.io/skip-immutable-check`. KubeBlocks hashes this spec and rejects the same-name upgrade with `immutable fields can't be updated`, marking the ComponentDefinition Unavailable; existing RabbitMQ upgrades and new clusters resolving this definition can fail before any Pod starts.

weicao:
Withdrawn at the requester's direction on 2026-09-18: do not change the RabbitMQ addon. This supersedes the previous request to repair and merge this PR. The candidate remains unmerged; the recorded review and test evidence are retained as historical records. No new addon implementation or upgrade validation is planned under this PR.

PR #3491, opened 2026-09-18
fix(mssql): use passwordConfig for system accounts

weicao:
SQL Server's three system accounts still use `passwordGenerationPolicy`. With KubeBlocks v1.2.0-alpha.3 (`4177ce0e`), the v1 schema drops that field and the controller creates empty passwords, so SQL login/readiness fails before the smoke test can run DML.

Change sa, kbadmin and master to `passwordConfig`, retaining all existing generation parameters. This patch targets main `2865430c`; it does not change KubeBlocks, account names, password complexity, scripts or test assertions.

Validation: Helm lint passes. Rendering before the patch reproduces three unknown fields against the frozen KB v1 CRD; rendering after the patch validates all three policies and differs only in the three field names. The original daemon smoke (n=1, one replica) passed on candidate `94b2bb17`: **10 PASS / 0 FAIL / 1 SKIP**. SQL login/primary health, DML, and row-count query (expected 1) passed. HA switchover was skipped for one replica. Run `tr-20260918122753375-106fad17` completed SUCCEEDED with cleanup PASS and UID-bound VCluster deletion/zero residue; KB `4177ce0e`, tests `a5fb8d68`, VCluster 0.24.0, SQL service `2022.19.0` / image `2022-CU19`. This is one smoke round, not full release or HA acceptance.

Contract reference: `kubeblocks-addon-docs/docs/addon-api/07-accounts-and-tls.md` account generation and runtime credential checks. Its legacy field spelling is stale for KB 1.2; the pinned v1 API and CRD define `passwordConfig`.


leon-ape (approved):
LGTM

PR #3494, opened 2026-09-19
fix(seaweedfs): accept gRPC port suffix in master role probe

weicao:
SeaweedFS 4.47 serializes master `Leader` as `host:9333.19333`. The role probe required the address to end in `:9333`, so it returned `unknown` even for a valid leader or follower response. Accept an optional numeric gRPC port suffix while preserving plain HTTP addresses and rejecting missing hosts or malformed suffixes.

This child PR targets `dev/seaweedfs-release-1.0` for #3493 and includes the static filer-template fix in its base (`a403d79`). The regression covers DNS and IPv4 addresses, a custom gRPC port, explicit and omitted `IsLeader:false`, and invalid address forms. The new tests first failed on the original script. After the fix, script tests 12/12, chart/schema checks 10/10, ShellCheck, and the ShellSpec entry passed. This is offline validation; runtime role-label convergence remains to be verified.

Sources: [4.47 ServerAddress constructor](https://github.com/seaweedfs/seaweedfs/blob/c5073360007d28385a33426a42ac3e4ec504c5a3/weed/pb/server_address.go#L17), [status response](https://github.com/seaweedfs/seaweedfs/blob/c5073360007d28385a33426a42ac3e4ec504c5a3/weed/server/raft_server_handlers.go#L13), and [addon role-probe contract](https://github.com/apecloud/kubeblocks-addon-docs/blob/96f3ec13e76daac8c08a98f7adbd23f3f9535f61/docs/addon-api/05a-lifecycle-roles-and-probe.md#L51).


leon-ape (commented):
P2: The chart advertises KubeBlocks >=1.0.0, but all six new ComponentDefinitions emit `spec.podUpdatePolicy`, which is absent from the released v1.0.0 and v1.0.1 CRDs. The PR's rendered-resource test fails against v1.0.0 with `podUpdatePolicy was unexpected` for each definition, while it passes against the newer pinned release-1.0 commit. Installations using strict schema validation on these advertised versions therefore reject the addon definitions.

weicao:
Superseded by merged PR #3506, which contains the same SeaweedFS gRPC master role probe fix on release-1.0.

PR #3495, opened 2026-09-19
fix(rustfs): use passwordConfig for root account

weicao:
RustFS declares its root account password policy as `passwordGenerationPolicy`, while the current KubeBlocks v1 API uses `passwordConfig` and treats a missing configuration as passwordless. The RustFS r4 smoke run reached a healthy four-pod cluster but stopped at the T03 non-empty secret-key check.

Rename the root account field to `passwordConfig`, preserving length 16, 8 digits and mixed case. The rendered manifests have no other semantic changes.

Validation:
- Helm dependency build and lint passed for the RustFS chart.
- Compared before/after Helm output against the KubeBlocks `ec3387e` account schema: the old field is unknown; the new configuration is valid, with the same password policy.
- Existing RustFS ShellSpec suite: 84 examples, 0 failures.
- No new cluster run or existing-cluster upgrade validation has been performed for this change.


leon-ape (approved):
LGTM

PR #3501, opened 2026-09-20
fix(minio): use passwordConfig for root account generation

weicao:
MinIO's root account still uses `passwordGenerationPolicy`, which is not present in the current KubeBlocks v1 account schema. The recognized field is `passwordConfig`; an absent password configuration is treated as a passwordless account.

Rename this single field in `addons/minio/templates/cmpd.yaml`. Keep the root/init account, length 16, 8 digits, mixed-case letters, credential references and image configuration unchanged.

Validation against addon base `a412e7a0371b7d846c2326496e3c3e7eb17436ea` and KubeBlocks `ec3387e4f992b29b84da08fd3a679b66f5d5e318`:
- Complete Helm rendering differs only by the field name; Helm lint passes.
- Strict offline validation using the actual v1 CRD account schema rejects the original unknown field and accepts the updated account with unchanged generation rules.
- Existing MinIO ShellSpec suite: 13 examples, 0 failures.
- IDC Guest r2 on this exact addon head: both MinIO Pods became Ready, and the generated password had16 characters,8 digits, uppercase and lowercase letters. S3 write succeeded; read-only diagnosis subsequently verified all45 original payload bytes. The Run remains FAILED because the Tests script required `cmp`, absent from the selected image; Tests9578f66 fixes that dependency and r3 subsequently passed all6 smoke steps, including exact S3 read-back and object/bucket deletion. R3 password generation again met the original rules. Its whole Run is BLOCKED/NOT_COUNTED because the platform cleanup entry left the business Cluster present, preventing addon uninstall. Platformb321936 fixes that separate lifecycle issue. Fresh r4 `tr-20260920101324794-03c6b7b9` using this exact addon head completed SUCCEEDED/PRODUCT_PASS/cleanupPASS, with all6smoke steps, S3counts1/1/1 and bucket deletion. Business cleanup and addon uninstall passed; independent Host absence and owned-process shutdown were verified. This is one smoke round, not broader release qualification.

Contract reference: [accounts/API-version guidance, lines 61–66](https://github.com/apecloud/kubeblocks-addon-docs/blob/08e824f62a79209f346add07484faeca75e8fd33/docs/addon-api/07-accounts-and-tls.md#L61-L66) and the linked field-pruning diagnosis. This is the separately authorized MinIO field change; it does not include the Milvus-embedded MinIO change in PR3180.

Hosted CI: the chart checks (default, Docker Hub and Aliyun), cluster chart, generated files and Shell Check pass. The repository-wide ShellSpec job reports 2495 examples, 2 failures, 9 pending. Both failures already occur on the exact main base `a412e7a0371b7d846c2326496e3c3e7eb17436ea`, whose [ShellSpec run 35483732925](https://github.com/apecloud/kubeblocks-addons/actions/runs/35483732925) failed before this PR with the same errors and totals:
- `addons/mysql/scripts-ut-spec/mgr_version_support_spec.sh:71` still expects a version-suffixed MGR selector and rejects `mysql-mgr-8.0`.
- `addons/rocketmq/scripts-ut-spec/exporter_version_compatibility_spec.sh:27` expects version-specific selectors but receives `rocketmq-exporter`.

Base PR3497 switched these ComponentVersion selectors to name prefixes. This PR does not change their files, shared Helm helpers or the ShellSpec workflow. A focused local run on the unmodified exact base also reproduces both failures (2 examples, 2 failures). The [PR ShellSpec run 35489773289](https://github.com/apecloud/kubeblocks-addons/actions/runs/35489773289) is therefore recorded as failed with inherited assertions, not as a passing full suite.


leon-ape (commented):
P1: This change replaces `spec.systemAccounts[0].passwordGenerationPolicy` with `passwordConfig` while the chart remains `1.2.0-alpha.0` and renders the same `minio-1.2.0-alpha.0` ComponentDefinition. KubeBlocks treats `spec.systemAccounts` and its password configuration as immutable and hashes this spec; an existing definition can therefore become `Unavailable` with `immutable fields can't be updated` during addon upgrade, blocking new Components before any Pod starts.

PR #3505, opened 2026-09-20
fix: omit empty fields that fail CRD validation under Helm SSA

shanshanying:
## What this PR does

Cherry-pick of https://github.com/apecloud/kubeblocks-addons/pull/3502 onto `release-1.0`.

Helm 4 server-side apply submits empty YAML keys as JSON `null`. CRD OpenAPI then rejects array/string fields, for example:

```
spec.runtime.volumes: Invalid value: "null": spec.runtime.volumes in body must be of type array: "null"
```

Omit those keys when the value is empty instead of emitting `null`.

## Cherry-pick conflict

`addons/mysql/templates/cmpd-proxysql.yaml` conflicted:

- `release-1.0` had `runtime.volumes: []`
- #3502 dropped an empty `volumes:` key

Resolved by omitting `volumes`, matching the PR intent.

## Addons

- mysql: drop empty ProxySQL `runtime.volumes`
- mongodb / apecloud-mysql: emit log-collector volumes only when enabled
- apecloud-mysql: align `_helpers.tpl` env defaults with main (`MYSQL_*_CONFIG` quoted empty strings, `VTTABLET_GRPC_PORT=16100`)
- clickhouse: guard `logConfigs` with `not (empty ...)`
- pulsar: wrap debug-only initContainer `env`; merge duplicate `init-bookies` env lists so debug vars are not overwritten
- oceanbase-ce: emit static/immutable parameter lists only when non-empty
- etcd / milvus: omit empty `serviceName`

## How to test

```bash
helm template kb-addon-mysql ./addons/mysql | kubectl apply --server-side --dry-run=server --force-conflicts -f -
```

Repeat for the addons above. Server-side apply must not fail with `must be of type array/string: "null"`.


shanshanying:
Close EMPTRY fields fix

